PR #16, opened 2015-04-13
Dev/master

MJHarrison-GFDL:
Following discussion w/ @Zhi-Liang updated closer to main branch. Should be ready for merge testing.
The search algorithm needs some re-thinking . We start with a prior guess for the nearest then search
locally. This failed when I tried with a newer grid.  This fix works for now.


underwoo:
This patch has been applied.  As I don't want the chance of an additional merge commit, I will simply close this pull request.

Issue #22, opened 2016-07-13
Axes attribute initialisation


nicjhan:
Axes%num_attributes is not initialised. Since it can take any value this can cause problems such as array bounds overruns.

nicjhan:
Example failure output:

https://climate-cms.nci.org.au/jenkins/job/mom-ocean.org/job/MOM6_run/build=DEBUG,compiler=gnu,experiment=ocean_only-benchmark,memory_type=dynamic/28/console

underwoo:
Thank you for reporting this. We will get it fixed and update the dev/master branch.

underwoo:
#23 merged.

PR #23, opened 2016-07-13
#22 initialise num_attributes, value used in loops before being set.

nicjhan:


underwoo:
We have brought this into our internal development.  When we merge this into dev/master we will push dev/master up.  When we are ready to release the first verona patch this will be included in that release.  

Thank you for your contribution.

underwoo:
Testing with this branch is completed.  We are running more tests on other code updates.  We should have this merged into master soon.

PR #24, opened 2016-07-13
Add 'sum' time operation to the diagnostic manager

ashao:
Adds the option of outputting a diagnostic that is summed over the output interval. This is intended for quantities which represent accumulated quantities, for example 'ea' or 'eb' in MOM6. To specify this option in the diag_table, set the time_avg filed to 'sum' or 'cumsum'. Variables using this time operation have the '_sum' suffix appended.

This feature was test using the MOM6 Baltic case. Fields were saved both as snapshots at the end of each timestep and also summed using the diag_manager.

### DIFF
--- a/diag_manager/diag_util.F90
+++ b/diag_manager/diag_util.F90
@@ -1491,6 +1491,7 @@ CONTAINS
     output_fields(out_num)%time_rms = .FALSE.
     output_fields(out_num)%time_min = .FALSE.
     output_fields(out_num)%time_max = .FALSE.
+    output_fields(out_num)%time_sum = .FALSE.
     output_fields(out_num)%time_ops = .FALSE.
     output_fields(out_num)%written_once = .FALSE.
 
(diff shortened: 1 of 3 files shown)


diag_manager/diag_util.F90
```diff
@@ -1563,12 +1564,19 @@ SUBROUTINE init_output_field(module_name, field_name, output_name, output_file,&
...
+               & output_fields(out_num)%output_name = TRIM(output_name)//'_sum'
+          method_selected = method_selected+1
+          t_method = 'sum'
        END SELECT
     END IF
 
     ! reconcile logical flags
     output_fields(out_num)%time_ops = output_fields(out_num)%time_min.OR.output_fields(out_num)%time_max&
-         & .OR.output_fields(out_num)%time_average
+         & .OR.output_fields(out_num)%time_average .OR.output_fields(out_num)%time_average
```
underwoo: I believe you wanted:

```Fortran
& .OR.output_fields(out_num)%time_average .OR. output_fields(out_num)%time_sum
```

Please correct.

ashao:
Sorry about that, forgot to commit that last typo when I was reviewing the code. Thanks for catching it.

underwoo:
Thank you for the update.  We will bring this into our internal development for testing.  Once tested, we will update the dev/master branch in github with the update.   When ready for our next release, we will merge this into master.

PR #26, opened 2016-09-05
During writes only increase mpp_io_stack when it is needed. Can impro…

### LINKED ISSUE #25: write_record() sets IO stack size even when not needed
The mpp_io_stack is used by write_record() as scratch space when writing a diagnostic with non-standard data packing. However the mpp_io_stack is increased to be the same size as the data even when no packing is used. 

This results in large unnecessary memory usage. For example I recently had a case where the mpp_io_stack was being increased to 200 Mb even though it was not needed. In some cases this can cause a process to run out of memory and exit. 

I suggest that we only increase the IO stack when it is needed.

nicjhan:
…ve memory usage significantly. Closes #25.

### DIFF
--- a/mpp/include/mpp_io_write.inc
+++ b/mpp/include/mpp_io_write.inc
@@ -1195,8 +1195,6 @@
       pointer( ptr1, data_r4)
       pointer( ptr2, packed_data)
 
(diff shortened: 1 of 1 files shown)


underwoo:
We are looking at this PR and should have word if we will take it shortly.

My apologies for the delay in looking at this.  I missed the email and have been extremely busy the last two months.

PR #35, opened 2017-08-08
Give FindNetCDF a hint to look for nc-config and nf-config

aerorahul:
Give FindNetCDF a hint to look for nc-config and nf-config. If $NETCD…F is defined, these configs are under $NETCDF/bin

Without this, the alternatives are to define paths for the executables explicitly or define environment variables. All these can be rather cumbersome. 

Thanks,
Rahul Mahajan


underwoo:
Thank you for your suggestion.  I will pull your suggestion into our internal development and run some tests.

PR #44, opened 2018-05-07
User/gmao/mppadjoint

danholdaway:
Merge request for the adjoint versions of the mpp communication routines required by FV3, used by NASA's Global Modeling and Assimilation Office and JCSDA's JEDI project.

We would appreciate if these adjoint routines could be part of the general FMS release so the code can be more easily shared in the future. It would also be helpful if the adjoint is tested when MPP undergoes general modifications so that code does not diverge.


Zhi-Liang:
I reviewed the codes and all the changes will not non-adjoint interfaces. I think we can accept those changes.

underwoo:
We have accepted this change into our internal repository, and marked it for the next release.  This next release will be out late July/early August.

danholdaway:
Thank you.

PR #46, opened 2018-07-12
MPI_Type, MPI_Alltoallw, mpp_global_field update

marshallward:
This patch contains three new features for FMS: Support for MPI datatypes, an
MPI_Alltoallw interface, and modifications to mpp_global_field to use these
changes for select operations.

These changes were primarily made to improve stability of large (>4000
rank) MPI jobs under OpenMPI at NCI.

There are differences in the performance of mpp_global_field,
occasionally even very large differences, but there is no consistency
across various MPI libraries.  One method will be faster in one library,
and slower in another, even across MPI versions.  Generally, the
MPI_Alltoallw method showed improved performance on our system, but this
is not a universal result.  We therefore introduce a flag to control
this feature.

The inclusion of MPI_Type support may also be seen as an opportunity to
introduce other new MPI features for other operations, e.g. halo
exchange.

Detailed changes are summarised below.

- MPI data transfer type ("MPI_Type") support has been added to FMS.  This is
  done with the following features:

  -  A `mpp_type` derived type has been added, which manages the type details
    and hides the MPI internals from the model developer.  Types are managed
    inside of an internal linked list, `datatypes`.
    
    Note: The name `mpp_type` is very similar to the preprocessor variable
    `MPP_TYPE_` and should possibly be renamed to something else, e.g.
    `mpp_datatype`.

  - `mpp_type_create` and `mpp_type_free` are used to create and release these
    types within the MPI library.  These append and remove mpp_types from the
    internal linked list, and include reference counters to manage duplicates.

  - A `mpp_byte` type is created as a module-level variable for default
    operations.

    NOTE: As the first element of the list, it also inadvertently provides
    access to the rest of `datatypes`, which is private, but there is probably
    some ways to address this.*

- A MPI_Alltoallw wrapper, using MPI_Types, has been added to the mpp_alltoall
  interface.

- An implementation of mpp_global_field using MPI_Alltoallw and mpp_types has
  been added.  In addition to replacing the point-to-point operations with a
  collective, it also eliminates the need to use the internal MPP stack.

  Since MPI_Alltoallw requires that the input field by contiguous, it is only
  enabled for data domains (i.e. compute + halo).  This limitation can be
  overcome, either by copying or more careful attention to layout, but it can
  be addressed in a future patch.

  This method is enabled in the `mpp_domains_nml` namelist group, by setting
  the `use_alltoallw` flag to True.

Provisional interfaces to SHMEM and serial ("nocomm") builds have been added,
although they are as yet untested and primarily meant as placeholders for now.

This patch also includes the following changes to support this work.

- In `get_peset`, the method used to generate MPI subcommunicators has been
  changed; specifically `MPI_Comm_create` has been replaced with
  `MPI_Comm_create_group`.  The former is blocking over all ranks, while the
  latter is only blocking over ranks in the subgroup.

  This was done to accommodate IO domains of a single rank, usually due to
  masking, which would result in no communication and cause a model hang.

  It seems that more recent changes in FMS related to handling single-rank
  communicators were made to avoid this particular scenario from happening, but
  I still think that it's more correct to use `MPI_Comm_create_group` and have
  left the change.

  This is an MPI 3.0 feature, so this might be an issue for older MPI
  libraries.

- Logical interfaces added to mpp_alltoall and mpp_alltoallv

- Single-rank PE checks in mpp_alltoall were removed to prevent model hangs
  with the subcommunicators.

- NULL_PE checks have been added to the original point-to-point implementation
  of mpp_global_field, although these may not be required anymore due to
  changes in the subcommunicator implementation.

  This work was by Nic Hannah, and may actually be part of an existing pull
  request.  (TODO: Check this!)

- Timer events have been added to mpp_type_create and mpp_type_free, although
  they are not yet initialized anywhere.

- The diagnostic field count was increased from 150 to 250, to support the
  current needs of researchers.


Zhi-Liang:
Hi,

   What is the reason that you need to use MPI_ALLTOALL_w? Do you have cases to send different data types in a single call? If not, you may try to use MPI_ALLTOALL_v, which is already implemented in mpp.  Also you may add a namelist in mpp_domains to have the option to use mpp_allto_all or mpp_send/mpp_recv. I tested the mpi_alltoall_v on our system and saw performance slowdown.

Thanks,

Zhi

marshallward:
Hi Zhi,

We needed to use a collective for this operation (either MPI_Alltoallw or otherwise) because the point-to-point method was causing hangs or crashes on our system.  For reasons that I don't fully understand, our library would either lose messages (OpenMPI 1.x) or would simply segfault (OpenMPI 2.x/3.x) if the message count became too large.  Using MPI_Alltoallw resolved this problem and allowed us to upgrade to the newest MPI library version.

The reason we used MPI_Alltoallw rather than MPI_Alltoallv is to support nonuniform grid domains and to skip over the halos in the local values.  I guess we could have packed the values into buffers and then use MPI_Alltoallv, but this felt like the cleaner approach.

I think the performance in MPI_Alltoall[vw] vs point-to-point can vary a lot depending on the library.  In my MPI_Alltoallw tests, point-to-point was fastest in OpenMPI 1.x (when it works!) but MPI_Alltoallw is significantly faster in 2.x and 3.x.  The results were opposite with Intel MPI (MPICH), with MPI_Alltoallw being slightly faster.  Even then, these numbers moved a bit on our different Infiniband networks (EDR vs FDR).

Which library are you using?  Is it a Cray MPI (MPICH) library?

In any case, I think we'd like to preserve the MPI_Type features if possible, since we'd like to continue and experiment with other MPI features, like the MPI_Cart_* methods.

Zhi-Liang:
Hi Marshall,

    I agree that we should have an option to do global_field using MPI_all2all. But I still prefer to use MPI_all2allv instead of MPI_all2allw. The reason is following,
1) MPP package is already complicated and add more codes will make it more complicated. Also we are in the middle of rewriting MPP package. 
2) I tried MPI derived type to do halo update and found the performance slow down a little bit comparing to original packing and unpacking. I guess MPI_all2allw will not have better performance than MPI_all2allv (possible will change depending on the mPI library). Currently we do not use MPI derived type in our MPP package. Unless we saw performance benefit, I think it is better not to make the change now.

Thanks,

Zhi

marshallward:
I agree that MPP is already complicated!  But I also believe that these changes here are largely non-intrusive and add new functionality while remaining consistent with existing FMS code.  I haven't changed the existing API, and haven't added any extra computation beyond the initialization of the derived type list.

I also agree that there could be some performance benefit to using `MPI_Alltoallv`, but this may also depend on the library and other components, and this could also change over time.  Using `MPI_Type` and associated functions have the potential to simplify the MPP code and to shift more work over to the library.  This may not be to our benefit right now, but this could change in the future as MPI developers prioritize these functions.

But the main issue for us is that we cannot run without switching to a `MPI_Alltoall*` collective, and relatively quickly.  The current point-to-point method is far too unstable in OpenMPI, which is the supported library on our machine, and we are already dependent on proprietary interfaces like MXM to keep things running, but this is raising other stability problems.  So we need something, and relatively quickly.

This one may not have optimal performance, but it seems to be better than what we are currently using, it's working, and it can be turned on or off as needed, so I do not see the harm in merging it.  But if you feel very strongly that it should not be in the code, then I can try to find some time to work on an `MPI_Alltoallv` version, although in the meantime we will probably not be able to use the official FMS branch.

marshallward:
Zhi, I had a look at using `MPI_Alltoallv` here, but I won't be able to use it without pre-copying the data into local buffers, since the messages between data still need to be contiguous, which is not the case when halos are present.  AFAIK, only `MPI_Type` allows one to skip over the halo data.

Since using local buggers not much different from the current point-to-point method, I probably won't try to implement an `MPI_Alltoallv` version.

Also, if this is creating unnecessary work for your MPP rewrite, I am happy to take responsibility for incorporating it into your new version.

underwoo:
@marshallward I would like to thank you for your desire to use the FMS code in this repository instead of the copy you currently have in [MOM5](/mom-ocean/MOM5).  We had an internal discussion today.  and we will be happy to approve this PR under a few conditions.

1. This updated cannot cause a change (numerically or performance) to the default configuration.  While I believe this is the case now, we won't know for sure until we run this through our internal regression tests.
1. We will not test this feature.  We currently don't have a test case, and we do not have the resources to expand our current regression test suite.
1. We cannot support this new feature.  If an update comes in that causes your update to no longer work, we would need you to supply the fix.  Any support requests (via github or any of the other support mechanisms we employ) will likely be pushed back to you or to the [ocean-mom/MOM5](ocean-mom/MOM5) project.
1. We will not port this feature during our redesign of the MPP/FMS infrastructure.  As mentioned by @Zhi-Liang, we are working on a redesign of the MPP/FMS code.  We will not port this feature into the new code.  If you continue to desire this feature after the redesign, you would need to do the port.  

We do appreciate your desire to contribute to the FMS code, and we understand if these conditions keep you from pointing to this project for MOM5.  We do hope you continue to contribute to the project.

marshallward:
I'm happy to take responsibility for this work.  If you want, I can try to add the tests to your test suite (either `test_mpp_domains.F90` or whatever you use internally).  I'm also prepared to handle any changes required by your FMS redesign, though it would be good to be kept updated on the progress of your MPP redesign.

Having said that, if you're concerned that these changes will interfere with your work or design goals, then we are also happy to work off a separate fork, with a commitment to keep up to date with GFDL's FMS release.

Whatever you decide is fine, I know that it can be disruptive to manage community contributions.

underwoo:
@marshallward, this pull request failed our automated compile tests with PGI. I hope this is only due to our test using an older version of PGI (as it also failed with an older version of Intel). I'm going to update the tests to use the latest version of PGI that we have available. 

While failing with PGI is not something that will keep us from including this, I would like to find something that will pass our PGI tests.

marshallward:
Thanks Seth, we have PGI as well so I will give it a try.

marshallward:
I have to apologise, it looks like our PGI licence has expired.  Also, I don't think our MPI libraries are configured for PGI anyway.

Can you send me the error?  I might be able to sort it out.

underwoo:
The issue  appears to be the use of the `RANK()` intrinsic.  It appears this was introduced in the TS 29113, but I cannot find out if/when PGI adopted it.

I did some testing, and I think I have a solution.  The following test case works with PGI, gfortran and Intel (version 16 and greater), and Cray (cce):

```Fortran
#ifdef __PGI
#define rank(X) size(shape(X))
#endif

program test_rank
  implicit none

  integer, dimension(3) :: a
  integer, dimension(3,4) :: b
  integer, dimension(4,5,6) :: c
  integer :: d

  write (*,FMT='("rank(a) = ",I1," size(shape(a)) = ",I1)') rank(a), size(shape(a))
  write (*,FMT='("rank(b) = ",I1," size(shape(b)) = ",I1)') rank(b), size(shape(b))
  write (*,FMT='("rank(c) = ",I1," size(shape(c)) = ",I1)') rank(c), size(shape(c))
  write (*,FMT='("rank(d) = ",I1," size(shape(d)) = ",I1)') rank(d), size(shape(d))
end program test_rank
```
The above gives the output

```
rank(a) = 1 size(shape(a)) = 1
rank(b) = 2 size(shape(b)) = 2
rank(c) = 3 size(shape(c)) = 3
rank(d) = 0 size(shape(d)) = 0
```

underwoo:
I put the `#define rank(X) size(shape(X))` in [mpp/mpp.F90](mpp/mpp.F90), and added a PR to your branch (https://github.com/marshallward/FMS/pull/1).  Please review, and if you like, take.

marshallward:
That's fine, I'd probably have put it in a header but otherwise it's what I'd do.  This is actually what I originally had before discovering the `rank` function (which I hadn't realised was a newer feature).

Anyway, I've merged it.  That commit had all of the GFDL work so hopefully that merge was clean!

underwoo:
These updates will be in our next round of internal testing.  Thank you.

underwoo:
You can see the updates in the PR, and all other updates scheduled to come out in the next release in branch [dev/master](https://github.com/NOAA-GFDL/FMS/tree/dev/master).

PR #48, opened 2018-08-08
Bugfix for Intel 18 compiler in mpp/include/mpp_gather.h

climbfuji:
Using the Intel 18 compiler on Cheyenne, NEMSfv3gfs crashes at the end of routine MPP_GATHER_PELIST_3D_. This is related to the object temp defined as

`type(array3d), dimension(size(pelist)) :: temp`

in the original code. Processors/MPI tasks for which `is_root_pe==.false.` do not make use of this derived data type and as such do not allocate members `temp(i)%data`. The crash does not happen if the code is compiled without optimization (NEMSfv3gfs's DEBUG mode), and it also does not happen with Intel 17, i.e. it is clearly a compiler bug.

This PR solves the problem by making temp an allocatable data structure and allocate/deallocate it inside the `is_root_pe` if-blocks. The proposed solution does not change the results, does not increase the runtime and does not increase the memory footprint.

### DIFF
--- a/mpp/include/mpp_gather.h
+++ b/mpp/include/mpp_gather.h
@@ -143,7 +143,7 @@ subroutine MPP_GATHER_PELIST_3D_(is, ie, js, je, nk, pelist, array_seg, data, is
    type array3D
      MPP_TYPE_, dimension(:,:,:), allocatable :: data
    endtype array3D
-   type(array3d), dimension(size(pelist)) :: temp
+   type(array3d), dimension(:), allocatable :: temp
 
    if (.not.ANY(mpp_pe().eq.pelist(:))) return
 
@@ -178,6 +178,7 @@ subroutine MPP_GATHER_PELIST_3D_(is, ie, js, je, nk, pelist, array_seg, data, is
 
 ! gather indices into global index on root_pe
    if (is_root_pe) then
+     allocate(temp(1:size(pelist)))
      do i = 1, size(pelist)
 ! root_pe data copy - no send to self
        if (pelist(i).eq.root_pe) then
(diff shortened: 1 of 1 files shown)


underwoo:
We will begin reviewing this PR and will let you know when we will merge this in.

Zhi-Liang:
The change looks fine to me.

Zhi

underwoo:
@climbfuji, we have pulled this update into our internal repository for testing.  This will be included in the next release.  This next release will be soon.  We don't have a firm date yet as we are working through other issues before minting the final release.

marshallward:
We've also noticed a bug in Intel 18 when derived types containing allocatable arrays are copied by vectors.  Specifically this, does not work, and only copies the first value.
```
vec(:) = value
```
This also does not work:
```
vec = value
```
but this does:
```
do i = 1, n
    vec(i) = value
end do
```
While we are able to use the explicit loop fix above, a colleague in the Bureau of Meteorology (Austraila) has been told by Intel that it will be fixed in Intel 19, and we have instead advised people to not use Intel 18.  (This also affects my MPI_Alltoallw patch, #46, specifically L413 and L420 in `mpp_do_global_field.h`).

The UM weather model has also shown similar problems with derived types.  I can't say if it's the same bug, but it seems very similar.

climbfuji:
Thanks @marshallward for sharing this information. This can save us a lot of time.

Zhi-Liang:
Hi Marshall,

   I tested the following program on our Gaea system and a Skylake box
using Intel 18. I do not see such issue.

Greetings,

Zhi

program test

implicit none
integer, parameter :: nx = 6
real, allocatable :: vec(:)
real :: value = 0.5
allocate(vec(nx))
vec(:) = value
print*, "vec=", vec
vec = value
print*, "vec2=", vec

end program test

marshallward:
Hi Zhi,
As mentioned in my comment, this only happens for derived types which contain allocatables.  Sample source below:
```
program dtvec

   implicit none



   type :: mpp_type

      integer, allocatable :: vec(:)

      integer :: id

   end type mpp_type



   type(mpp_type) :: s

   type(mpp_type), allocatable :: t(:)

   integer :: i, nd



   nd = 5

   s%id = 100

   allocate(t(1:nd))

   t(:) = s



   do i = 1, nd

       print *, t(i)%id

   end do

end program dtvec
```

Zhi-Liang:
Hi Marshall,

  I was able to reproduce the issue with your test case.

  Thanks,

Zhi

PR #49, opened 2018-09-07
Remove logic to get and set CPU affinity for Darwin (MacOSX) platforms

climbfuji:
FMS contains C code that makes Linux-specific system calls that allow to place (pin) threads on processors (affinity.c). These system calls are deprecated on Darwin (MacOSX) platforms.

This PR uses predefined CCPP macros (ifndef `__APPLE__`) to suppress building the C code and the calling functions code in data_override.F90 and test_mpp_domains.F90.

This advanced feature of explicitly pinning threads to processors is for performance reasons only. Since Darwin systems are used for development and not for high performance computing, it is - in my opinion - acceptable to disabling this code.

With the changes in this PR, FMS can be built on Darwin platforms using either clang/gfortran or gcc/gfortran (tested as part of NEMSfv3gfs).

### DIFF
--- a/mpp/affinity.c
+++ b/mpp/affinity.c
@@ -27,10 +27,12 @@
 #include <sys/resource.h>
 #include <sys/syscall.h>
 
+#ifndef __APPLE__
 static pid_t gettid(void)
 {
   return syscall(__NR_gettid);
 }
+#endif
 
 /*
  * Returns this thread's CPU affinity, if bound to a single core,
@@ -38,6 +40,7 @@ static pid_t gettid(void)
  */
 int get_cpu_affinity(void)
 {
+#ifndef __APPLE__
   cpu_set_t coremask;		/* core affinity mask */
 
   CPU_ZERO(&coremask);
@@ -60,6 +63,7 @@ int get_cpu_affinity(void)
 
   if (last_cpu != -1) {return (first_cpu);}
   return (last_cpu == -1) ? first_cpu : -1;
+#endif
 }
 
 int get_cpu_affinity_(void) { return get_cpu_affinity(); }	/* Fortran interface */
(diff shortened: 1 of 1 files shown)


bensonr:
Dom - is it possible to just create no-op versions of the functions in affinity.c and constrain the CPP directives to that file only?

climbfuji:
Sure, that's possible, too. Will have a look at it.

climbfuji:
@bensonr I updated the PR so that all changes are in affinity.c only. One could also move the ifdef statement around `static pid_t gettid` inside the routine in case that routine ever gets called from outside. Please let me know what you prefer, and whether I should point this PR to dev/master instead of master. Thanks!

bensonr:
I wasn't following this PR and am just now seeing this update.  It really doesn't matter if you point this to master or dev/master as it will be merged into dev/master by default and then pushed to master.

Has this fix been tested in the NEMSfv3gfs system for the various platforms being supported?

climbfuji:
Yes it has been tested with NEMSfv3gfs on Theia, Cheyenne and on MacOSX.

raphaeldussin:
Hi @climbfuji @bensonr @menzel-gfdl 

I would love to see that fix merged. I have tested in the latest FMS version and it compiles just fine on my mac (GCC8, MPICH3). Let me know if you want me to open a new PR with the updated branch or if you want to resuscitate this PR.

Thanks!

wrongkindofdoctor:
@bensonr @underwoo I was able to compile FMS and MOM6 with this version of affinity.c with gnu9 (source build), and run the simple ocean-only MOM6_solo_global_ALE_z case on my MAC (OS version: mojave 10.14.5). Rusty, please confirm whether I can merge it into Master, or if you want to close this request and incorporate the changes into your modified affinity.c file before merging.

PR #50, opened 2018-09-07
fms_platform.h: correct preprocessor directives for quad precision computing

climbfuji:
This PR corrects/updates the preprocessor directives in include/fms_platform.h.

As of version 4.6.0, the GNU compilers do support real*16 computations via libquadmath. However, the PGI compilers (as of today) still don't.

With the changes in this PR, FMS can be built with the PGI compilers, and GNU compilers can make use of real*16 computations. These changes were tested with gfortran 6.2.0 on Theia, gfortran 6.3.0 on Cheyenne, gfortran 6.2.0 and gfortran 8.1.0 on MacOSX, pgf90 17.7 on Theia, and pgf90 17.9 on Cheyenne.

### DIFF
--- a/include/fms_platform.h
+++ b/include/fms_platform.h
@@ -92,8 +92,8 @@
 #define NF_GET_ATT_REAL nf_get_att_double
 #endif
 
-#if defined __CRAYXT_COMPUTE_LINUX_TARGET || defined __GFORTRAN__
-!Cray XT compilers do not support real*16 computation
+#if defined __CRAYXT_COMPUTE_LINUX_TARGET || defined __PGI
+!Cray XT and PGI compilers do not support real*16 computation
 !also known as 128-bit or quad precision
 #define NO_QUAD_PRECISION
 #endif


underwoo:
We have recently updated the `include/fms_platform.h` file (https://github.com/NOAA-GFDL/FMS/blob/dev/master/include/fms_platform.h).  This new version will be included in our next release (due out soon).  Please review the new version, and if you feel it is needed, apply your updates to the new version.

climbfuji:
@underwoo thanks for the hint, good to know that the master branch is behind the latest development version. should I use dev/master for all testing and PRs instead of master then?

climbfuji:
@underwoo I tested the dev/master version with both GNU and PGI. For the case of NEMSfv3gfs, the results are different with the GNU compiler than with the current master (because quad precision is now switched off by default). This is not an issue for us, since we do not perform any bit for bit testing with the GNU compilers for NEMSfv3gfs yet (but maybe other applications do?).

The FMS code in dev/master also compiles with the PGI compilers without the above bugfix, but this is simply because quad precision is switched off by default. It would be safer in my opinion to guard the switch to quad precision in lines 109-115 so that QUAD_PRECISION becomes undefined or has no effect for the PGI compiler, e.g.:

```
!If you want to use quad-precision.
#define NO_QUAD_PRECISION
#if defined(QUAD_PRECISION) && !defined(__PGI)
#undef NO_QUAD_PRECISION
#else
#define QUAD_KIND DOUBLE_KIND
#endif
```

I am not going to create a PR for this unless you want me to, since it is not needed for our applications at the moment.

underwoo:
Thank you.  We will address this in a future update.  No need for you to add a PR.

include/fms_platform.h
```diff
@@ -92,8 +92,8 @@
 #define NF_GET_ATT_REAL nf_get_att_double
 #endif
 
-#if defined __CRAYXT_COMPUTE_LINUX_TARGET || defined __GFORTRAN__
-!Cray XT compilers do not support real*16 computation
+#if defined __CRAYXT_COMPUTE_LINUX_TARGET || defined __PGI
+!Cray XT and PGI compilers do not support real*16 computation
```
edhartnett: The correct way to accomplish this is to set some macro like HAVE_QUAD_REAL in configure.ac, and use it everywhere in the code.

Then, we could take the next step, which is to write a configure check which determines whether the compiler supports real*8. Then this code will work on all compilers, even ones that we don't know about or test on.

climbfuji:
I see that this has been fixed in tag 2019.01. Thanks!

PR #53, opened 2018-10-14
enable/disable affinity.c on macOS (APPLE)

aerorahul:
Simple bugfix to allow building on FMS on macOS.

Change-Id: Id17e31c82101c89b3d047c8d04a1027eb0b3dc03

### DIFF
--- a/mpp/affinity.c
+++ b/mpp/affinity.c
@@ -29,9 +29,14 @@
 
 static pid_t gettid(void)
 {
+#ifdef __APPLE__
+  return syscall(SYS_gettid);
+#else
   return syscall(__NR_gettid);
+#endif
 }
 
+#ifndef __APPLE__
 /*
  * Returns this thread's CPU affinity, if bound to a single core,
  * or else -1.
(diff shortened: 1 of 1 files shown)


underwoo:
Part of this is already implemented in #49.  With #49, is the PR still needed?

aerorahul:
@underwoo You are right. I did not see the other pull request. I will test it and then close this unless I find any discernible differences that may need addressing. 
The only thing I can think of is the Apple implementation of gettid.

climbfuji:
I just saw this PR. My original version at the Apple implementation of gettid in there, but not the settid part, which would have been much harder to implement. What @bensonr and I discussed at that time was that when running on MacOSX platforms, performance and in particular fine tweaks such as thread pinning to achieve the best performance, are not the priority. Comments?

underwoo:
I agree with @bensonr: performance of the code on the Mac OS X platform is not a priority.  We will, as much as possible, take these types of updates as long as they don't impact too much the readability of the code, and have no impact to the performance or reliability on the main production platforms where the code is in use.

PR #63, opened 2019-02-26
affinity.c get_cpu_affinity only one return

### LINKED ISSUE #62: affinity.c has only one possible return value
Looking at the `mpp/affinity.c` code, the `get_cpu_affinity` function has two return lines.  

```c
  if (last_cpu != -1) {return (first_cpu);}
  return (last_cpu == -1) ? first_cpu : -1;
```
However, when looking at all the possible ways the two `if` statements can be evaluated, there is only one true return value: `first_cpu`.  This should be reduced to only one return statement.  

Further evaluation shows the function sets a default value for `first_cpu = -1`, and this should be safe.

underwoo:
The `get_cpu_affinity` function in `mpp/affinity.c` had two return lines
that essentially resulted in only one possible return value `first_cpu`.
This commite reduces the return statement to simply `return first_cpu`.

Resolves #62.

This needs to be merged at the same time as NOAA-GFDL/coupler#9.

### DIFF
--- a/mpp/affinity.c
+++ b/mpp/affinity.c
@@ -67,7 +67,7 @@ int get_cpu_affinity_(void) { return get_cpu_affinity(); }	/* Fortran interface 
 /*
  * Set CPU affinity to one core.
  */
-void set_cpu_affinity( int cpu )
+int set_cpu_affinity( int cpu )
 {
   cpu_set_t coremask;		/* core affinity mask */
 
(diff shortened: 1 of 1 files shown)


bensonr:
This code update is acceptable.

There needs to be an update to coupler_main.F90 to test the result of calls to this function to ensure it is a valid return code (/= -1).  Could you please create an issue and assign it to me.

underwoo:
Issue NOAA-GFDL/coupler#7 created.

menzel-gfdl:
@underwoo I see several problems with this file:

1.) The function set_cpu_affinity() is defined to return void, but instead tries to return 0 or -1.
2.) Similar issue for the function set_cpu_affinity_
3.) The variable last_cpu is no longer used in the function get_cpu_affinity()

underwoo:
@menzel-gfdl I updated the function prototypes to be `int`.

bensonr:
+1

Code updates mirror what I tested with internally.

We may need to rework pull request #49 in light of the interface and return code changes.

PR #74, opened 2019-03-07
Parallel netCDF I/O support

marshallward:
This patch enables parallel netCDF support for FMS, allowing files that would otherwise be saved as `*.nc.####` IO domains to be written and modified by multiple processes as a single file via MPI-IO.

**Note:** We submit this with the knowledge that FMS and the `mpp` modules in particular are undergoing a re-write, and that any of these changes may be phased out.  We would only ask that they be included in a final release before the MPP overhaul, as they were used in a manuscript to evaluate the performance of parallel netCDF.

Actual performance will depend on many factors, such as MPI-IO and Lustre configuration, so it would be best to consult us before doing any performance tests on local machines.

Detailed changes are available in the commit logs, a summary is provided below.

----

The following namelist control variables have been added to `mpp_io_nml`:
```
&mpp_io_nml
  parallel_netcdf = .true.     # enable parallel I/O (Default: .false.)
  parallel_read = .false.      # Enable parallel I/O for read operations (Default: .false.)
  pnetcdf = .false.            # enable PnetCDF backend (.true.) or HDF5 backend (Default: .false.)
  parallel_chunk = .true.      # Default chunking (.false.) or customized chunk for netCDF-4 format(.true.).
  chunk_layout = cnk_x, cnk_y   # The user defined chunk layout if parallel_chunk is set as .true.
/
```

We have added the following methods to the `domain2d` object:

* `mpp_get_num_io_domains`: returns the number of I/O domains
* `mpp_get_io_root_pes`: Returns the rank IDs of the root I/O PEs, as defined in the domain `pelist`
* `mpp_domain_is_missing_io_domain`: Returns true if an I/O domain has been removed from a domain from masking

A quicksort method in `sort/sort_quicksort.f90` has been added to reorder the I/O root PEs for generation of the `peset`.

The following wrappers have been added to support parallel netcdf file open/creation:

* `mpp_nc_create`
* `mpp_nc_open`

Usage of `NF_PUT_ATT_DOUBLE(..., some_double)` in a few places has been replaced with `nf_put_att_real(..., real(some_double), FLOAT_KIND))`, due to a potential issues in `nf_put_att_double` when using parallel netCDF.  Specifically, these functions appear only write the lower 32 bits as mask values over masked I/O domains where no MPI rank writes to the file.

----

Note that `io_layout` is required to be defined when using this, since we use it in part to define the MPI ranks which participate in I/O.


mpp/mpp_io.F90
```diff
@@ -1063,8 +1063,18 @@ module mpp_io_mod
   integer            :: deflate_level = -1
   logical            :: cf_compliance = .false.
 
-  namelist /mpp_io_nml/header_buffer_val, global_field_on_root_pe, io_clocks_on, &
-                       shuffle, deflate_level, cf_compliance
+
+  ! TODO: Move parallel_netcdf to fms_io.F90 (also rename?)
```
marshallward: Remove me

mpp/mpp.F90
```diff
@@ -1290,6 +1299,10 @@ module mpp_mod
 #include <mpp_util.inc>
 #include <mpp_comm.inc>
 
+! TODO: Where to put this...?
```
marshallward: Remove me

mpp/mpp_domains.F90
```diff
@@ -405,6 +407,10 @@ module mpp_domains_mod
      type(overlapSpec),  pointer :: update_C      => NULL() ! send and recv information for halo update of C-cell.
      type(overlapSpec),  pointer :: update_N      => NULL() ! send and recv information for halo update of N-cell.
      type(domain2d),     pointer :: io_domain     => NULL() ! domain for IO, will be set through calling mpp_set_io_domain ( this will be changed).
+
+     ! IO domain metadata used for parllel I/O
```
marshallward: Typo in comment

sort/sort_quicksort.f90
```diff
@@ -0,0 +1,86 @@
...
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+!***********************************************************************
+
+! Quicksort implementation
+!   Simple implementation of a quicksort of an integer array.
+!   Directly based on the pseudocode implementation at Wikipedia.
+!   (https://en.wikipedia.org/w/index.php?title=Quicksort&oldid=883205571)
+!   Notes:
+!       - In-place sort; the original order is not preserved.
+!       - Pivot is the median value
```
marshallward: Middle, not median

mpp/include/mpp_write_2Ddecomp.h
```diff
@@ -306,17 +327,36 @@
...
+                      position = position, &
+                      flags=XUPDATE+YUPDATE+GLOBAL_ROOT_ONLY, &
+                      default_data=default_data &
+                  )
+                  ! Domain check
+                  if(mpp_file(unit)%write_on_this_pe ) &
+                      call WRITE_RECORD_(unit, field, size(gdata(:,:,:)), &
+                                         gdata, tstamp, io_domain)
+
+                  io_domain => NULL()   ! TODO: move outside conditional?
```
marshallward: Resolve TODO

mpp/include/mpp_write_2Ddecomp.h
```diff
@@ -124,8 +130,17 @@
...
+             call mpp_get_global_domain(domain, is, ie, js, je)
+             start(1:2) = [is, js]
+             axsiz(1:2) = [ie - is + 1, je - js + 1]
+          end if
 
-          if( debug )print '(a,2i6,12i6)', 'WRITE_RECORD: PE, unit, start, axsiz=', pe, unit, start, axsiz
+!     print *, 'WRITE_RECORD: PE= ',pe, ' name= ',trim(field%name),' start= ', start,' axsiz= ', axsiz
+!          if( debug )print '(a,2i6,12i6)', 'WRITE_RECORD: PE, unit, start, axsiz=', pe, unit, start, axsiz
+!     print '(a,i4,a,a,a8,i,a8,i)', 'WRITE_RECORD: PE= ',pe, ' name= ',trim(field%name),' start= ', start,' axsiz= ', axsiz
+!          print '(a,2i6,12i6)', 'WRITE_RECORD: PE, unit, start, axsiz=', pe, unit, start, axsiz
```
marshallward: Remove comment block, restore original debug output statement

mpp/include/mpp_write_2Ddecomp.h
```diff
@@ -72,6 +70,13 @@
                  endif
               endif
               call netcdf_err( error, mpp_file(unit) )
+
+              if (parallel_netcdf) then
+                  error = nf_var_par_access(mpp_file(unit)%ncid, field%id, nf_collective)
+                  if(mpp_pe() == mpp_root_pe()) &
+                      print *, 'WRITE_RECORD: PE= ',pe,'set nf_collective'
```
marshallward: Remove this debug message

marshallward:
Looks like this have gotten a little more strict under the new build system:
```
../mpp/include/mpp_netcdf.inc:32.8:
    Included at mpp.F90:1410:
    if (IAND(mode, NF_NOWRITE)) then
        1
Error: IF clause at (1) requires a scalar LOGICAL expression
```
I will sort this out once I get into work this morning.

marshallward:
Seems to be an automake issue, guessing the files are no longer automatically detected by `mkmf`?
```
../mpp/include/mpp_domains_define.inc:470.8:
    Included at mpp_domains.F90:2793:
    use quicksort_mod, only : quicksort
        1
Fatal Error: Can't open module file 'quicksort_mod.mod' for reading at (1): No such file or directory
```

sort/sort_quicksort.f90
```diff
@@ -0,0 +1,86 @@
+!***********************************************************************
```
edhartnett: There is already a drifters/quicksort.F90. Are you sure you can't use that instead of adding a new one?

marshallward: Possibly, I had not known it was there.  I can try it.

edhartnett:
This would also be easier if you merged the current master into your branch before putting up the PR.

marshallward:
Agreed, trying to get acquainted with the automake build now.

sort/sort_quicksort.f90
```diff
@@ -0,0 +1,86 @@
...
+!*                   GNU Lesser General Public License
```
edhartnett: This file ends in .f90, but all the other FMS files end in .F90. Is this intentional?

My understanding is that some compilers will not run pre-processor on .f90, but will on .F90, but I may be incorrect.

marshallward: There wasn't any preprocessing, so I assumed it would be find to use `.f90`.   But as mentioned in your other comment, maybe it is not needed.

menzel-gfdl:
@marshallward I apologize for the delay in getting to this.  As I think @underwoo has told you, we prefer to pull these updates into a stream separate from master.  I have pushed the branch "with-parallel-netcdf"; can you please edit this pull request to make that the target instead of master?

marshallward:
I've updated the PR to go to the new branch.

This branch is still not compatible with the automake changes since AFAIK it needs the explicit paths in `Makefile.am` and other `configure.ac` changes, but that's probably not a concern if the changes are not going into master.

If you want, you can just close this branch and merge #75 over to NOAA-GFDL:with-parallel-netcdf, but I don't have control over that branch.

I would still like to see this implemented in the newer FMS IO, but I'm happy to wait until they are stable before trying to port it over.

PR #75, opened 2019-03-07
PR #74 merged with master + build system

edhartnett:
This may not be ready to merge, based on @marshallward .

This is PR #74 with the necessary changes to the build system. 

I have added comments in the PR code review explaining what I added to the build system, and why.


marshallward:
Thanks very much @edhartnett 

GFDL folks, I'm happy to remove the `sort` content if the drifters quicksort is sufficient.

edhartnett:
Not so fast!

drifters depends on mpp, and mpp needs this sort, so if you want to use the same sort in drifter and mpp, you actually have to separate it out like this anyway. So you can't just use the drifter one without moving it. ;-)

Makefile.am
```diff
@@ -5,8 +5,8 @@
 ACLOCAL_AMFLAGS = -I m4
 
 # Make targets will be run in each subdirectory. Order is significant.
-SUBDIRS = include platform constants tridiagonal mpp memutils fms	\
-mosaic time_manager diag_manager drifters axis_utils horiz_interp	\
+SUBDIRS = include sort platform constants tridiagonal mpp memutils	\
```
edhartnett: Note here that a new directory "sort" has been added. Order is significant, it must be before mpp.

configure.ac
```diff
@@ -64,6 +64,7 @@ AC_CONFIG_FILES([Makefile
         time_interp/Makefile
         time_manager/Makefile
         constants/Makefile
+        sort/Makefile
```
edhartnett: Every directory needs a Makefile.am, and every Makefile.am generates a Makefile, which must be listed in configure.ac. In this case, order is not important.

mpp/Makefile.am
```diff
@@ -9,6 +9,7 @@ SUBDIRS = include
 # Include .h and .mod files.
 AM_CPPFLAGS = -I${top_srcdir}/include
 AM_CPPFLAGS += -I${top_srcdir}/mpp/include
+AM_CPPFLAGS += -I${top_builddir}/sort
```
edhartnett: Since we are using the sort mod, we need to point to its directory in the include.

mpp/include/Makefile.am
```diff
@@ -33,4 +33,5 @@ mpp_domains_comm.inc mpp_domains_util.inc				\
 mpp_io_unstructured_read.inc mpp_read_distributed_ascii.inc		\
 mpp_unstruct_domain.inc mpp_util_sma.inc mpp_comm_mpi.inc		\
 mpp_data_nocomm.inc mpp_domains_define.inc mpp_io_connect.inc		\
-mpp_io_unstructured_write.inc mpp_sum_ad.inc mpp_util.inc
+mpp_io_unstructured_write.inc mpp_sum_ad.inc mpp_util.inc		\
+mpp_netcdf.inc
```
edhartnett: Since mpp_netcdf.inc has been added to the build, it must be listed in a Makefile.am somewhere, or else it will not be present in the tarball distributions. Just having it in the git repo is no longer enough, since end users will not be cloning the repo.

Since the travis CI on this github project does a make distcheck, this is checked by travis.

sort/Makefile.am
```diff
@@ -0,0 +1,20 @@
+# This is an automake file for the sort directory of the FMS
```
edhartnett: A simple Makefile.am based on the one in the platform directory.

PR #79, opened 2019-03-08
add default fortran flags set in configure

edhartnett:
Part of #73.

In this PR I add the setting of default fortran flags by configure. 

Setting of fortran flags can be disabled with --disable-fortran-flags. The FCFLAGS variable remains for the user to use, and any flags set in FCFLAGS will also be used (appended to the flags selected by configure).

Currently configure only sets flags for gfortran. If it seems good we can add other compilers as needed. If the user is using an unknown compiler, their best bet will be to use --disable-fortran-flags and figure out their own flags.

This is not a good thing to do - in fact it's a severe hack. But we will do it to help a smooth transition to the new build system. (The proper solution is for this code to be buildable without flags.)

Each currently-necessary flag that can be removed simplifies the build system and make build errors less likely, thus reducing maintenance and support for you. The ideal is a build-anywhere package which requires almost no support. netcdf-fortran has achieved this, so it is certainly possible for FMS.

I hope that this feature can be removed from the build system, after the code has been changed to reduce the dependence on fortran compiler flags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,54 @@
 *.o
 *.mod
 *.s
-Makefile*
 pathnames*
+
+Makefile
+# From https://github.com/github/gitignore/blob/master/Autotools.gitignore
+# http://www.gnu.org/software/automake
+Makefile.in
+/ar-lib
+/mdate-sh
+/py-compile
+/test-driver
+/ylwrap
+*.log
+*.trs
+
+# http://www.gnu.org/software/autoconf
+autom4te.cache
+/autoscan.log
+/autoscan-*.log
+/aclocal.m4
+/compile
+/config.guess
+/config.h.in
+/config.log
+/config.status
+/config.sub
+/configure
+/configure.scan
+/depcomp
+/install-sh
+/missing
+/stamp-h1
+.deps
+
+# https://www.gnu.org/software/libtool/
+libtool
+/ltmain.sh
+*.lo
+*.la
+.libs
+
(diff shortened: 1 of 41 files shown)


edhartnett:
@underwoo do you want to try this to allow the new build to start setting fortran flags?

underwoo:
I'll review this, and run some tests next week.  I had hoped to look into this earlier this week, but I was unable.  Thank you for the work you are doing on this.

edhartnett:
I have written up this work as a paper, perhaps for the AMS or AGU. If you send me your email I will share the document with you. It also serves to document the new build system.

underwoo:
I added my email to my profile.

edhartnett:
Tried to share, but got this error message:

The administrator for underwoo.io has disabled the ability to receive items from outside their domain. If those you're attempting to share with would like more information, they should contact their domain administrator directly.

edhartnett:
You can just send me mail if you want: edward.hartnett@noaa.gov.

underwoo:
Sent you an email.
--
Seth Underwood

S/W Development Process Lead
Modeling Systems Group
GFDL/NOAA/DOC
201 Forrestal Road
Princeton, NJ 08540-6649

(609) 452-5847 Office
(304) 376-9002 Cell
(609) 987-5063 Fax
Seth.Underwood@noaa.gov

edhartnett:
Is this still relevant or should I close it?

underwoo:
I see no reason to keep this PR open.

PR #81, opened 2019-03-13
Define default_data in save_default_restart to avoid NaN trap

wrongkindofdoctor:
-Initialized default_data to 0.0 in fms_io::setup_one_field as a precaution.
-Added default_data local variable to fms_io:: save_defaul_restart.
-Added a check to determine if cur_var%default_data attribute is a NaN.  If it is a NaN, default_data is set to MPP_FILL_DOUBLE (~9e36) as in setup_one_field; otherwise default_data is set to cur_var%default_data. 
-default_data is passed to mpp_write instead of cur_var%default_data.
-This modification was tested with fv3gfs atmosphere and fms code forked from the Github GFDL:FMS repo compiled with intel 16 in debug + avx on C3


underwoo:
Per our conversation yesterday, I am closing this PR.  Please reopen when we have fully decided how to deal with this problem.

PR #82, opened 2019-03-18
User/jml/default data fix overload r8

wrongkindofdoctor:
@menzel-gfdl 
@bensonr 
Added conditional statements in register_restart_field_r3d8 and register_restart_field_r2d8 that only pass the optional 'data_default_r4' argument to 'setup_one_field' if data_default is initially passed to the register_restart_field interface. These modifications allow 'setup_one_field' to set the variable attribute 'default_data' to MPP_FILL_DOUBLE. 

The current code passes data_default_r4, and uninitialized real variable, to setup_one_field regardless of whether it is defined or not. Since the fv3gfs_io register_restart_field calls do not define data_default, the uninitialized data_default_r4 is passed to setup_one_field. Setup_one_field sees that data_default is present, and sets the variable attribute 'default_data' to the uninitialized value, a NaN, which is trapped in the call to mpp_write by save_default_restart in the debug-compiled executable.

This fix was tested on Gaea C3 with Intel 16.

### DIFF
--- a/fms/fms_io.F90
+++ b/fms/fms_io.F90
@@ -1741,7 +1741,7 @@ function register_restart_field_r2d8(fileObj, filename, fieldname, data, domain,
   else
      call setup_one_field(fileObj, filename, fieldname, (/size(data,1), size(data,2), 1, 1/), &
                           index_field, domain, mandatory, no_domain, is_compressed, &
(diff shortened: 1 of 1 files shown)


wrongkindofdoctor:
All right. I'll make the change and resubmit

PR #94, opened 2019-07-01
Tests on Google Compute

underwoo:
This has an example of running test on a Google Compute VM with Slurm.


underwoo:
This was only a test, and is not needed ATT.

PR #95, opened 2019-07-08
compile time fix for OSX

raphaeldussin:
this is an updated PR of #49

it basically removes the CPU affinity logic for OSX builds. 
Just a workaround and not the perfect solution but right now FMS do not compile on OSX at all.

### DIFF
--- a/mpp/affinity.c
+++ b/mpp/affinity.c
@@ -27,10 +27,12 @@
 #include <sys/resource.h>
 #include <sys/syscall.h>
 
+#ifndef __APPLE__
 static pid_t gettid(void)
 {
   return syscall(__NR_gettid);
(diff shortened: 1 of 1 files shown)


raphaeldussin:
@underwoo this is the PR we talked about this morning. I think your travis is broken btw.

wrongkindofdoctor:
Since we already have a pull request #49 for a simple fix that encloses the process affinity binding in #ifndef __APPLE that I have tested with a simple ocean-only configuration on my own machine running mojave (10.14.5) with source builds of gcc 9.1, mpich 3.3.1, netcdf, and their dependencies, I will close this pull request merge in #49 upon approval from the FMS group heads.

PR #104, opened 2019-08-05
fms2_io and diag_manager updates

thomas-robinson:
@wrongkindofdoctor This branch should be made the head of the development stream for fms2_io

PR #108, opened 2019-08-08
Merges fms2_io and updates to FMS with master

thomas-robinson:
This branch resolves merge conflicts found in #104 and should be done in lieu of #104.


thomas-robinson:
After consulting with @menzel-gfdl we determined that merging this code is going to be hampered by the presence of the bats program. He included bats to use with autotools, and the reason the CI fails is because of running the tests with `make distcheck`.  We need to decide how we want to handle the tests before we can merge this in.

thomas-robinson:
@underwoo can you please review the updates the .travis.yml file to make sure you are alright with them.  I clone bats and updating the path before running the script.

PR #109, opened 2019-08-09
Removes bats from travis.yml

thomas-robinson:


thomas-robinson:
@underwoo this is the fix we talked about

underwoo:
@thomas-robinson I put in a pull request with a fix for the `addons:apt:sources:ubuntu-toolchain-r-test` error.  It also contains a few other cleanup items that we should include as well.  Please accept that PR.  I will then approve this PR --- that is assuming all the tests pass.

PR #110, opened 2019-08-09
Add 32-bit compile for libFMS

underwoo:
This adds an additional travis test to build libFMS in 32-bit mode, which is used by the fv3gfs models.


underwoo:
> I suggest we change FCFLAGS_REAL_KIND to FCFLAGS_KIND and hard type both integer and real for the 32-bit and 64-bit compiles. This way we don't assume the default real_kind=4 (or 32) - as we are in the 32-bit compile.

I don't have an issue with the suggestion, however I do not see gfortran options for default real(kind=4).  The build will still have an empty environment variable for the 32-bit compile.

PR #113, opened 2019-08-12
Adds the ensemble number to the file_name outut by diag_integral_mod

### LINKED ISSUE #111: Adding ensemble number to diag_integral.out
A user requested adding the ensemble number to the end of the diag_integral.out filename to distinguish between ensembles.

> As background, the reason I'm interested in this is that I've used this 
file to figure out the radiative imbalance of the model in the past with 
single member runs. Now that I have ensembles to play with I'd like to 
be able to do the same thing for each ensemble member and see the spread.
I still get a single diag_integral.out file from my runs, but I have no 
idea which member the data belongs to or whether the data gets written 
by the first, last, or "last to get to the routine" member. I don't even 
know if different timesteps are from the same member or an average of 
the entire ensemble.

thomas-robinson:
Resolves #111 
When an ensemble is run, the diag_integral.out file will be named diag_integral.ens_##.out where ## is a 2-digit number designated to the ensemble number.

### DIFF
--- a/diag_integral/diag_integral.F90
+++ b/diag_integral/diag_integral.F90
@@ -109,6 +109,7 @@ use fms_mod,          only:  open_file, file_exist, error_mesg, &
                              stdlog, close_file
 use constants_mod,    only:  radius, constants_init
 use mpp_mod,          only:  mpp_sum, mpp_init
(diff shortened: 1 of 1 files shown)


thomas-robinson:
This is branched off of xanadu, and a new dependency was added.  #114 Fixes the autotools problem.

PR #114, opened 2019-08-12
diag_integral output includes ensemble number

### LINKED ISSUE #111: Adding ensemble number to diag_integral.out

thomas-robinson:
Resolves #111
When an ensemble is run, the diag_integral.out file will be named diag_integral.ens_##.out where ## is a 2-digit number designated to the ensemble number.  This also resolves autotools compilation errors.

### DIFF
--- a/diag_integral/diag_integral.F90
+++ b/diag_integral/diag_integral.F90
@@ -1527,6 +1527,7 @@ real, dimension (size(data,1),size(data,2)) :: data2
 
 end function vert_diag_integral
 
+!> \brief Adds .ens_## to the diag_integral.out file name 
 function ensemble_file_name(fname) result(updated_file_name)
      character (len=mxch), intent(inout) :: fname
      character (len=mxch) :: updated_file_name
@@ -1534,7 +1535,7 @@ function ensemble_file_name(fname) result(updated_file_name)
      character(len=7) :: ensemble_suffix
      character(len=2) :: ensemble_id_char 
      integer :: i
-     !> Make sure the file name short enough ti handle adding the ensemble number
+     !> Make sure the file name short enough to handle adding the ensemble number
      if (len(trim(fname)) > mxch-7) call error_mesg ('diag_integral_mod :: ensemble_file_name',  &
           trim(fname)//" is too long and can not support adding ens_XX.  Please shorten the "//&
           "file_name in the diag_integral_nml", FATAL)
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,8 +7,8 @@ ACLOCAL_AMFLAGS = -I m4
 # Make targets will be run in each subdirectory. Order is significant.
 SUBDIRS = include platform constants tridiagonal mpp memutils fms	\
 mosaic time_manager diag_manager drifters axis_utils horiz_interp	\
-time_interp diag_integral column_diagnostics block_control		\
-data_override astronomy field_manager coupler monin_obukhov		\
+time_interp column_diagnostics block_control data_override		\
+astronomy field_manager coupler diag_integral monin_obukhov		\
 interpolator fft amip_interp oda_tools exchange topography		\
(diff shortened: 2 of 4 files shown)


thomas-robinson:
@wrongkindofdoctor you can squash the commits because there are a bunch of small ones where I try to get autotools to work.

diag_integral/diag_integral.F90
```diff
@@ -1522,7 +1527,40 @@ function vert_diag_integral (data, wt) result (data2)
...
 
-
+function ensemble_file_name(fname) result(updated_file_name)
+     character (len=mxch), intent(inout) :: fname
+     character (len=mxch) :: updated_file_name
+     integer :: ensemble_id_int
+     character(len=7) :: ensemble_suffix
+     character(len=2) :: ensemble_id_char 
+     integer :: i
+     !> Make sure the file name short enough ti handle adding the ensemble number
```
wrongkindofdoctor: minor typo: 'ti' should be 'to'

diag_integral/diag_integral.F90
```diff
@@ -1522,7 +1527,40 @@ function vert_diag_integral (data, wt) result (data2)
...
+          ensemble_id_int = get_ensemble_id()
+          write(ensemble_id_char,"(I0)") ensemble_id_int
+     !> Add a 0 if the ensemble is less than 10 (2 digit)
+          if (ensemble_id_int < 10) then
+               ensemble_suffix = ".ens_0"//trim(ensemble_id_char)
+          elseif (ensemble_id_int >= 10 .and. ensemble_id_int < 100) then
+               ensemble_suffix = ".ens_"//trim(ensemble_id_char)
+          else
+               call error_mesg ('diag_integral_mod',  &
+                ' Does not support ensemble sizes over 99.', FATAL)
```
wrongkindofdoctor: I know MAX_ENSEMBLE_SIZE is[ currently set to 100](https://github.com/NOAA-GFDL/FMS/blob/bb4e81fe3598af0069eb0ad576d6dd956275d209/coupler/ensemble_manager.F90#L35) in ensemble_manager.F90, but is this size sufficient to support the current (or anticipated) standard ensemble sizes defined in the atmos/ESM workflows of the users who want this feature implemented?

diag_integral/diag_integral.F90
thomas-robinson: I am the master of the typo.  It is fixed now.

diag_integral/diag_integral.F90
thomas-robinson: No, but this is a conversation that is beyond the scope of this merge request.  It's also a simple enough fix here, BUT running 100 model runs with 864 cores is also beyond the limits of what we are currently allowed to run on our production system.  This is definitely a larger conversation.

wrongkindofdoctor: All right. Thanks for confirming that this is a resource limitation rather than a standard workload size.

thomas-robinson:
`mxch` is the length of `file_name` which is the variable passed into the new routine.  `file_name` is a namelist variable, and I've only seen it set to "diag_integral.out".  In my first pass at this, I actually had an if statement that checked if `file_name` was set to diag_integral.out.  

Adding the ensemble name only adds 7 characters ".ens_00", so 64 should be sufficient.

PR #120, opened 2019-08-14
History files name fix

uramirez8707:
https://github.com/thomas-robinson/FMS/pull/7 fixed an issue where the names of the history files were coming out with an extra .nc suffix

I got a crash in a different test case because the length of the string fname_no_tile was less than len_file_name since in this test case it got into the L182-196 where "tile" was removed from the file name. 

I fixed it so that the loop goes from 1 to the length of fname_no_tile instead

### DIFF
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -196,7 +196,7 @@ CONTAINS
     endif
 !> If there is a .nc suffix on the file name, remove the .nc
     if (len(trim(fname_no_tile)) > 3 ) then
-       checkNC: do i = 3,len_file_name
+       checkNC: do i = 3,len(fname_no_tile)
          if (fname_no_tile(i-2:i) == ".nc") then
             fname_no_tile(i-2:i) = "   "
             exit checkNC
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -196,7 +196,7 @@ CONTAINS
     endif
 !> If there is a .nc suffix on the file name, remove the .nc
     if (len(trim(fname_no_tile)) > 3 ) then
-       checkNC: do i = 3,len(fname_no_tile)
+       checkNC: do i = 3,len(trim(fname_no_tile))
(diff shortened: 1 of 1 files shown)


uramirez8707:
@thomas-robinson @wrongkindofdoctor

uramirez8707:
The error is 
fms_lm4p_xanadu.x  0000000001946121  diag_output_mod_m         200  diag_output.F90
fms_lm4p_xanadu.x  0000000001026F11  diag_util_mod_mp_        1970  diag_util.F90
fms_lm4p_xanadu.x  000000000104B23B  diag_util_mod_mp_        2633  diag_util.F90
fms_lm4p_xanadu.x  0000000001042E9C  diag_util_mod_mp_        2529  diag_util.F90
fms_lm4p_xanadu.x  000000000104E850  diag_util_mod_mp_        2689  diag_util.F90
fms_lm4p_xanadu.x  00000000017C5E87  diag_manager_mod_        3766  diag_manager.F90
fms_lm4p_xanadu.x  00000000017BEB44  diag_manager_mod_        3674  diag_manager.F90
fms_lm4p_xanadu.x  000000000042270E  coupler_main_IP_c        2056  coupler_main.F90
fms_lm4p_xanadu.x  000000000041008E  MAIN__                   1148  coupler_main.F90
forrtl: severe (408): fort: (4): Variable FNAME_NO_TILE has substring ending point 19 which is greater than the variable length of 18

I see this with intel debug mode. 

The filename that cause the error has "tilex" in it. So it goes into L189, which removes the "tilex" part so the length of fname_no_tile is less than len_file_name. This wasn't an issue before because before I was only doing a run with 1 tile.

wrongkindofdoctor:
This seems all right for now. The next iteration(s) of diag_manager should not be parsing file names this way (e.g., checking that '.nc' is appended to the base name), though. It would be preferable to simply try to open the file with the new I/O routines, and issue a FATAL if the open fails that prints a message indicating that the file name obtained from the input namelist is invalid. This places the onus on the user to correctly define files and their locations rather than the software to play 'guess-the-file-name'.

PR #121, opened 2019-08-14
New_io: Fix in compressed_write where the incorrect compressed index was used

uramirez8707:
Ran into a crash due to a buffer size error in save_vegn_restart with compressed_write. This error was because the wrong compressed dimension index was used. The code was using the fileobj compressed dimension when assigning c and e instead of the variable's

The function get_variable_compressed_dimension_index was used the get the index of the compressed dimension for the variable. This index was then used to assign c and e:
```fortran 
c(compressed_dim_index) = fileobj%compressed_dims(compressed_dim_index)%npes_corner(i)
e(compressed_dim_index) = fileobj%compressed_dims(compressed_dim_index)%npes_nelems(i)
```
 In this case they were not the same because the file_obj had more than 1 compressed dimension, so the code crash. 

I modified the function get_variable_compressed_dimension_index to output a 2 element array where the 1st element is the compressed dimension of the variable and the 2nd element is the compressed dimension of the file obj. Then, I used the correct index when assigning c and e in compressed_write.inc.

### DIFF
--- a/fms2_io/include/compressed_write.inc
+++ b/fms2_io/include/compressed_write.inc
@@ -18,10 +18,10 @@ subroutine compressed_write_0d(fileobj, variable_name, cdata, unlim_dim_level, &
                                           !! where the data
                                           !! will be written to.
 
-  integer :: compressed_dim_index
+  integer, dimension(2) :: compressed_dim_index
 
   compressed_dim_index = get_variable_compressed_dimension_index(fileobj, variable_name)
-  if (compressed_dim_index .eq. dimension_not_found) then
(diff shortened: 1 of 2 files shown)


uramirez8707:
@wrongkindofdoctor @menzel-gfdl

wrongkindofdoctor:
@uramirez8707 I just requested Ray for review. In the mean time, could you pull the latest updates from fms-io-dev into your branch to update the diag_integral travis-ci.yml files, then commit the changes? This should resolved the build check failure for this PR.

PR #122, opened 2019-08-14
new_fms_io: add check for missing value in is_valid

uramirez8707:
Modified is_valid to check for the missing value attribute to support input files that use this in land.

### DIFF
--- a/fms2_io/netcdf_io.F90
+++ b/fms2_io/netcdf_io.F90
@@ -83,7 +83,6 @@ type, public :: Valid_t
   logical :: has_fill !< Flag that's true a user defined fill value.
   logical :: has_missing !< Flag that's true a user defined missing value.
   real(kind=real64) :: fill_val !< Unpacked fill value for a variable.
-  logical :: has_missing !< Flag that's true a user defined missing value.
   real(kind=real64) :: min_val !< Unpacked minimum value allowed for a variable.
   real(kind=real64) :: max_val !< Unpacked maximum value allowed for a variable.
 endtype Valid_t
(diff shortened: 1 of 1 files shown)


uramirez8707:
@wrongkindofdoctor

fms2_io/netcdf_io.F90
```diff
@@ -81,7 +81,9 @@ module netcdf_io_mod
 type, public :: Valid_t
   logical :: has_range !< Flag that's true if both min/max exist for a variable.
   logical :: has_fill !< Flag that's true a user defined fill value.
+  logical :: has_missing !< Flag that's true a user defined missing value.
```
wrongkindofdoctor: Remove one of the `has_missing` declarations

fms2_io/netcdf_io.F90
```diff
@@ -1493,6 +1495,25 @@ function get_valid(fileobj, variable_name) &
       add_offset = 0._real64
     endif
 
+    !Get default max/min from missing_value.  These could be overwritten by
```
wrongkindofdoctor: Change the description to be more explicit; state that max and and min data values are defined by the valid_range, valid_min, and valid_max attributes if they are present. If the fill_value attribute is present and valid_range is not, then fill_value determines valid_data values. Otherwise, the missing_value attribute determines valid data_values if it is present, and valid_range  fill_value attributes are not.

fms2_io/netcdf_io.F90
uramirez8707: Done

fms2_io/netcdf_io.F90
uramirez8707: Done

wrongkindofdoctor:
@uramirez8707 The gnu7 build fails with the following error:

```
netcdf_io.F90:1506:23:
 valid%missing_val = buffer(1)*scale_factor + add_offset
Error: ‘missing_val’ at (1) is not a member of the ‘valid_t’ structure
```

The full log is [here](https://travis-ci.org/NOAA-GFDL/FMS/jobs/571997334)

uramirez8707:
@wrongkindofdoctor Sorry about that. It passes now.

Issue #119, opened 2019-08-14
domain_read: answer changes in land with new io  


uramirez8707:
I am getting answer changes when running with the new io in land because the river restart is not read correctly. I think it is because code is setting every rank as the root pe and is using the same starting and ending index in the read, so they are all reading the same part of the file.

In L322-324 of fms_netcdf_domain_io.F90, when opening the file the code gets the pelist from the *io_domain* 
```F90
  pelist_size = mpp_get_domain_npes(io_domain)
  allocate(pelist(pelist_size))
  call mpp_get_pelist(io_domain, pelist)
````
The pelist ends up been just a **one** element array with the current pe

Then, in L433-442 of netcdf_io.F90, the code assigns the root pe as the first element in the pelist to the fileobj 

In domain_read.inc, in L133-140 (for a 2D read), it gets the starting index and length of each pe list
```F90
    call mpp_get_compute_domains(io_domain, xbegin=pe_isc, xsize=pe_icsize, position=xpos)
    call mpp_get_compute_domains(io_domain, ybegin=pe_jsc, ysize=pe_jcsize, position=ypos)
    do i = 1, size(fileobj%pelist)
      c(xdim_index) = pe_isc(i) - pe_isc(1) + 1
      c(ydim_index) = pe_jsc(i) - pe_jsc(1) + 1
      e(xdim_index) = pe_icsize(i)
      e(ydim_index) = pe_jcsize(i)
```
c and e end up being the same for all pe, so I get the same data. 

This is my work dir /lustre/f2/scratch/gfdl/Uriel.Ramirez/work/esm2mb-luh2/esm2mb-luh2_1x0m0d72h_64x1a.o201466261

In this case, there is only 1 tile.

uramirez8707:
@wrongkindofdoctor @menzel-gfdl

PR #126, opened 2019-08-21
Domain read/write bug fixes 

uramirez8707:
This fixes https://github.com/NOAA-GFDL/FMS/issues/119

There were two main problems:

1.  Domain_read was not reading a combined file correctly because the starting and ending index of the read were adjusted for all of the ranks 

This is fixed by adding a flag called "adjust_indices" in the fileobj. The flag is set to true by default unless you have combined file and the io_layout is not (1,1). 

2. Global checksums were not calculated correctly in cases were the position was set to North or East because the ending index of the buffer of each rank was counted twice. 

This is fixed by adding flags "extra_x_point" and "extra_x_point". These flags are true if the ending x/y index is not equal to the max index allowed and the position is set to east/north.  If true the x/y dimension of the buffer is decreased by one.

This also changes the unit_tests to test this case.


uramirez8707:
@wrongkindofdoctor  Please test this with your MOM test case

Issue #135, opened 2019-09-05
Compile failed with gcc/gfortran "debug" settings 


thomas-robinson:
I tried to compile FMS on gaea with PrgEnv-gnu/6.0.3 and gcc/7.3.0.  Here is the compile line and error:
```
ftn -DINTERNAL_FILE_NML -Duse_libMPI -Duse_netCDF -D_FILE_VERSION="`git-version-string /ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/affinity/fms_affinity.F90`" -Duse_LARGEFILE  -I/opt/cray/pe/netcdf/4.4.0/GNU/5.1/include -fcray-pointer -fdefault-real-8 -fdefault-double-8 -Waliasing -ffree-line-length-none -fno-range-check -O0 -g -W -fbounds-check -ffpe-trap=invalid,zero,overflow -fopenmp   -c -I/ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/include	/ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/affinity/fms_affinity.F90
/ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/affinity/fms_affinity.F90:149:6:

 !$OMP PARALLEL NUM_THREADS (nthreads) &
      1
Error: Unclassifiable OpenMP directive at (1)
/ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/affinity/fms_affinity.F90:150:15:

 !$             DEFAULT (none) &
               1
Error: Unclassifiable statement at (1)
/ncrc/home2/Thomas.Robinson/awg/fmsio/cm4p12_fmsio/src/FMS/affinity/fms_affinity.F90:173:18:

 !$OMP END PARALLEL
                  1
Error: Unexpected !$OMP END PARALLEL statement at (1)
Makefile:89: recipe for target 'fms_affinity.o' failed
```
I am working off of the fms-io-dev branch, but I also checked the master branch and encountered this problem.

thomas-robinson:
After some googling, I think I figured out how to solve this and put in pull request #136

PR #136, opened 2019-09-05
OMP line continuation

thomas-robinson:
This resolves a problem when compiling with gfortran using the -fopenmp flag in #135 .  It also adds a travis-CI build with the -fopenmp flag to spot issues with openMP in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,9 @@ addons:
     - pkg-config gfortran netcdf-bin libnetcdf-dev libnetcdff-dev openmpi-bin libopenmpi-dev 
 
 env:
-  - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ addons:
 env:
   - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8'
   - CPPFLAGS_ADD='-DOVERLOAD_R4 -DOVERLOAD_R8' FCFLAGS_REAL_KIND=''
+  - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8 -fopenmp'
```
underwoo: I would rather see a new OPENMP environment

```suggestion
  - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8' OPENMP='-fopenmp'
```

You'll then need to add `$OPENMP` to the `FCFLAGS` on line 33.  (also add to tother environments.

PR #139, opened 2019-09-05
Setting up autotools to run unit tests

uramirez8707:
This sets up autotools to run unit tests: 
1. I modified the yml script to download and install bats, instead of adding it to the FMS repo
2. I commented out tests that need input files: data_override, diag_manager, xgrid, interpolator, mosaic, horiz_interp, and time_interp for now
3. I commented out tests that I wasn't able to get to pass: mpp_domains

The tests run with a `make check` and `make distcheck`


uramirez8707:
@underwoo @wrongkindofdoctor

underwoo:
Thank you for taking `bats` out of the `.travis.yml`.  I ran some tests on my home linux system, and I have a few changes that were required to allow all the tests to run (and to use an external build directory).  I'll put in a PR to your branch shortly.  Unfortunately, the travis tests still failed.  I think it may be due to memory on the travis system.  I'm going to run a few more tests.

edwardhartnett:
The best way to proceed here would be to comment out the travis tests that are failing, and merge the rest. Then work on the failing tests incrementally.

ANY tests are better than no tests, which is the current state of affairs. By trying to achieve too much at once, the entire process is being delayed.

The use of BATS is not clearly needed. Simple tests that return a 0 for success and another value for failure are all that are needed. However, if you can get it working now, then let's get it merged, and we can address the BATS issues incrementally as well.

underwoo:
The issue with the tests is not a bats issue, but more than likely a problem with either amount of memory required for the test.  All the current tests run on my home Linux system, running the same OS version as the Travis CI VM.  As I had to reduce the size of some of the MPI tests as the amount of memory required was more than I have, I think this is the same problem with Travis.  However, I would like to know that for sure before we skip the test and perform the merge.

underwoo:
I have some updates that will help the tests pass.  Hopefully, these updates will correct the travis CI tests as well (they now pass on my Mac).  I also added the required bits to have `make check` interpret correctly the TAP output from bats.  This will also allow us to more easily see which tests have failed.  Unfortunately, until we get something that can upload the Travis CI artifacts, we won't be able to see exactly what failed  as it is hidden in the test log files.

I'm running through the make distcheck now.  When I get that to pass I'll push an update to my branch.

underwoo:
The tests are still failing on Travis in my fork (https://travis-ci.com/underwoo/FMS/jobs/236609548).  I'll add something to skip these tests for now, until I can get them working.  At this point I think it is more important if a few tests are available than none at all. 

@uramirez8707 keep an eye open for a pull request to this branch in your fork.  It should appear in the next day or so.

underwoo:
Think I got it working on Travis CI.  Many of the failures were due to the tests running in parallel on Travis CI.  And a lot of them were due to the fact that many tests need an `input.nml` file, and there is no way to change the name of that file.

Other tests are MPI failures on Travis.  These could be due to problems in the tests, or something specific to the Travis CI VM.  Most of these are in mpp.  Since we have an mpp rewrite in the works, I think time will be better spent ensuring those tests pass.

@uramirez8707, please accept my PR in your fork.  This should allow this PR to pass so it can be accepted.

uramirez8707:
@underwoo Thanks for this. I merged your PR.

underwoo:
Thank you @uramirez8707.  Unfortunately, one of the tests `test_drivers_input` fails when compiled in mixed mode.  I'm looking to see if there is a way to get around this -- looks like a double free.  If I can't find something by noon today, I'll skip the test until we can find a solution.  Keep your eyes open for another PR.

underwoo:
@wrongkindofdoctor I approve this PR.

edwardhartnett:
@wrongkindofdoctor can we get this merged soon? I am working on other test issues and don't want us all to drift apart...

Issue #96, opened 2019-07-08
preprocessor directive __APPLE__ is not obtained from system 


raphaeldussin:
contrary to C code, __APPLE__ is not defined in Fortran and used here:

https://github.com/NOAA-GFDL/FMS/blob/db20e2577cce4659ec4be7e9498895ef842fda3a/memutils/memutils.F90#L302

this requires the user to pass explicitly the cpp at compile time (e.g. ' -D__APPLE__')

underwoo:
It is defined by gfortran, at least version 9.1.0.

```
> gfortran --version
GNU Fortran (Homebrew GCC 9.1.0) 9.1.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

> gfortran -cpp -E -dM - < /dev/null | grep APPLE
#define __APPLE_CC__ 1
#define __APPLE__ 1
```

I don't have the ability to check if ifort or PGI Fortran ATT.  When I can I will check those as well.

marshallward:
Hi Seth, we also did that test and got the same result, but for some reason `__APPLE__` is not being set in the preprocessed+compiled Fortran source code.  It would be good if you could independently test that out.

Passing the file explicitly through `cpp` does set `__APPLE__`, however.

This issue does point to a possible issue, namely that GFortran switched from explicitly running `cpp` to invoking `libcpp`, but it's over 7 years old and may no longer be relevant:

https://stackoverflow.com/questions/23856859/preprocessor-macro-not-working-on-windows-in-fortran-code/23872966#23872966

raphaeldussin:
+1 with @marshallward 

```
MacBookPro:~/tmp/sandbox cat detect.F90
program test

#ifdef _WIN32
  print *,'Windows'
#endif
#ifdef __linux
  print *,'Linux'
#endif
#ifdef __APPLE__
  print *,'OSX'
#endif

end program

MacBookPro:~/tmp/sandbox gfortran -E detect.F90
# 1 "detect.F90"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "detect.F90"
program test

# 12 "detect.F90"

end program

MacBookPro:~/tmp/sandbox cpp -E detect.F90
# 1 "detect.F90"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 360 "<built-in>" 3
[...]

PR #141, opened 2019-09-10
Using AC_CANONICAL_TARGET to get the target_vendor for apple systems

### LINKED ISSUE #140: configure.ac test for Apple OS type fails for some Mac OS systems
The check in the configure.ac is not fully Mac OS X portable.  Some Mac OS X systems set OSTYPE to have `darwin[0-9]+`, instead of just `darwin`.  Something more portable should be used.

underwoo:
Using an autoconf set variable to determine if the target system is an apple, instead of relying on Apple to not change something later.

Fixes #140

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,7 @@ AC_INIT([FMS], [2.0-development], [])
 
 # Find out about the host we're building on.
 AC_CANONICAL_HOST
+AC_CANONICAL_BUILD
 
 # Find out about the target we're building for.
 AC_CANONICAL_TARGET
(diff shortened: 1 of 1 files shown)


thomas-robinson:
@underwoo how do I test this on my mac?

underwoo:
Run the commands:

```
autoreconf -if
./configure CPPFLAGS="<flags you need>" CFLAGS="<flags you need>" FCFLAGS="<flags you need>" CC=<MPI C> FC=<MPI Fortran>
```

Beside to include the flags you need on your machine.  Once finished, check the `config.h` file for `#define __APPLE__`.

edwardhartnett:
Please do not do this. It is a mistake. This is not the way these tools are intended to be used.

underwoo:
@edwardhartnett, the checks for `__APPLE__` is to allow compiles on Mac OS X to correctly get the process memory usage.  We need to set this cpp macro specifically due to `gfortran` not passing in this macro (see #96).  Other compilers do set this cpp macro, and pass it down to the Fortran preprocessor, including `gcc` and `g++`.

If/when gfortran begins to, correctly, pass `__APPLE__` we can remove this type of hack.  Note, we could use some other macro to correctly get the process memory,  but it would be the same hack.

edwardhartnett:
I withdraw my objection. Thanks!

PR #148, opened 2019-09-11
Check for real and double precision types being 8 bytes in the configure script

### LINKED ISSUE #147: Have configure ensure that Fortran is compiling real and double precision types as 8 bytes
As I recently was reminded on the AM4 project, this code needs to be compiled with Fortran flags such that real and double precision types are both 8 bytes.

We don't want to try and set flags but we can check that they have been set correct and suggest settings for common compilers.

edwardhartnett:
Fixes #147 

This PR comes after PR #145 in includes the changes from that PR.

In this PR I add two macros to the acinclude.m4 file, GFDL_CHECK_FORTRAN_REAL_8 and GFDL_CHECK_FORTRAN_DOUBLE_8. These are then called in the configure.ac.

I am checking these with autoconf macro AC_RUN_IFELSE. This builds a little test program, runs it, and looks at the result. The program that is built is:

```
        program foo
        real var
        if (sizeof(var) .ne. 8) then
        stop 2
        end if
        end program foo

```

What would be extra clever would be to have a program that fails to compile if sizeof(real) .ne. 8. In other words, if the program did not have to be run. However, this only comes into play in cross-compiling situations, which are pretty rare in the HPC world these days.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -8,17 +8,23 @@ AC_DEFUN([GFDL_CHECK_FORTRAN_LONG_LINES],
 # Tell autoconf that language tests are in fortran.
 AC_LANG_PUSH(Fortran)
 
+# Check a fortran compile with a test program with this file
+# extension.
 AC_FC_SRCEXT(F90)
+
+# Attempt to compile test program with 160 char line of code.
 AC_MSG_CHECKING([whether Fortran compiler can handle long lines of code])
 AC_COMPILE_IFELSE([AC_LANG_SOURCE([
         subroutine foo(bar)
         integer, intent(in) :: bar
-        print *,'7890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890'
+        print *,'789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890'
         end subroutine foo
         ])],
         [test_result=yes],
         [test_result=no])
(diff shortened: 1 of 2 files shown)


edwardhartnett:
One of the travis builds is failing. It is setting some flags that are new to me:

`CPPFLAGS_ADD='-DOVERLOAD_R4 -DOVERLOAD_R8' FCFLAGS_REAL_KIND=''`

Can someone tell me what these flags do?

When I look in the code I see:

```
#ifdef OVERLOAD_R4
#undef MPP_REDUCE_0D_
#define MPP_REDUCE_0D_ mpp_max_real4_0d
#undef MPP_REDUCE_1D_
#define MPP_REDUCE_1D_ mpp_max_real4_1d
#undef MPP_TYPE_
#define MPP_TYPE_ real(FLOAT_KIND)
#undef MPP_TYPE_BYTELEN_
#define MPP_TYPE_BYTELEN_ 4
#undef MPI_TYPE_
#define MPI_TYPE_ MPI_REAL4
#undef MPI_REDUCE_
#define MPI_REDUCE_ MPI_MAX
#include <mpp_reduce_mpi.h>
#endif
```

So, how shall this all be handled in configure? Do we want the ability to set the overload macros from configure? For example, we could have an --enable-overload-r4 option.

Apparently if these overload options are specified, I don't need to ensure that real is 8 bytes?

edwardhartnett:
OK, I will hit this problem again shortly. First let's sort out the fortran line length stuff.

PR #150, opened 2019-09-11
Change configure to ensure cray pointers are handled by fortran compiler

### LINKED ISSUE #149: Have configure check for Fortran handling of cray pointers
The code uses cray pointers (see https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gfortran/Cray-pointers.html)

This requires a Fortran flag (-fcray-pointer for gfortran).

This should be checked by configure, so the user gets a good error message from configure instead of a compile time error.

edwardhartnett:
Fixes #149 

This PR includes PR #148 .

In this PR I add a macro GFDL_CHECK_FORTRAN_CRAY_POINTERS to acinclude.m4, and run it in configure. If the user fails to set the cray pointer flag, then the configure will fail.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -8,17 +8,23 @@ AC_DEFUN([GFDL_CHECK_FORTRAN_LONG_LINES],
 # Tell autoconf that language tests are in fortran.
 AC_LANG_PUSH(Fortran)
 
+# Check a fortran compile with a test program with this file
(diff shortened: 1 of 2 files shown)


edwardhartnett:
I will circle around to this again shortly. Let me get the fortran line length stuff merged first...

PR #151, opened 2019-09-11
no range check flag not needed

edwardhartnett:
We have been using the -fno-range-check flag (in gfortran).

However, there seem to be no range check violations in the code. When I remove this flag, the code builds just fine.

That being the case, I suggest you remove it from the travis file as well. That way, if a program introduces some range violation into the code, it will be automatically caught.


edwardhartnett:
OK, guess I was wrong!

Seemingly this flag is needed when the overide macros are used. I will wait until I have more clarification there before dealing with the flag.

PR #153, opened 2019-09-13
cause configure to set fortran flags to allow extra-long lines of code, also give user ability to turn off fortran flag setting

### LINKED ISSUE #144: Have configure script ensure that Fortran compiler can handle extra long lines of code
Some of the lines of code in the code base are very long, and this makes fortran compilers unhappy.

For example, ../mpp/include/mpp_chksum_int.h:92:132:

`         errStr = errStr // trim(tmpstr4) // "has been called with these strange values. Check your KINDS, _FillValue, pack and mask_val. // &
`

This line is 141 chars long.

In gfortran, this requires Fortran flag -ffree-line-length-none. Other compilers will have their own flags.

Instead of failing in compile, the configure script needs to detect whether the fortran compiler can handle very long lines of code. If not, configure will fail with a message telling the user what went wrong. It is up to the user to then figure out the appropriate flag and set it in FCFLAGS.

The error message I have prints this:
`configure: error: Fortran compiler must be able to handle long lines of code.    Set FCFLAGS to allow this (-ffree-line-length-none for gfortran)`

What is the flag for ifort? I will mention it in the error message, as well as the gfortran flag.

### LINKED ISSUE #152: Allow user to turn off fortran flag setting with --disable-fortran-flag-setting option in configure
We have decided to have configure set fortran flags for the user, to make the code more readily build with autotools.

This is not ideal, because sometimes our flag choices will be incorrect for specific installs. The best way to cope with this is to allow the users to turn off fortran flag setting if they want.

So, by default, fortran flags will be set. But, if that causes everything to break for the user, then can use --disable-fortran-flag-setting and then they have complete control over fortran flags. The build system will not attempt to add any.

This is very easy to do and I will include it in a PR shortly.

edwardhartnett:
This is a new attempt at PR #145 

Fixes #144
Fixes #152 

In this PR I add the --disable-fortran-flag-setting option, in case a user wants to turn off fortran flag setting. By default, the configure will attempt to set fortran flags for the user.

Also in this PR, now we call the AC_FC_FREEFORM and AC_FC_LINE_LENGTH macros, to set the flags that allow fortran to build the extra long lines of code in FMS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,6 +53,10 @@ AC_LANG_POP(Fortran)
 # Require netCDF.
 #AC_CHECK_FUNC([nf_open], [], [AC_MSG_ERROR([NetCDF Fortran library required to build FMS])])
 
+# Check that long lines of Fortran code can be handled. This will add
+# necessary fortran flags.
+#AC_FC_LINE_LENGTH([unlimited])
+
 # These defines are required for the build.
 AC_DEFINE([use_netCDF], [1])
 AC_DEFINE([use_libMPI], [1])
--- a/configure.ac
+++ b/configure.ac
@@ -21,6 +21,19 @@ AC_CONFIG_MACRO_DIR([m4])
 LT_PREREQ([2.4])
 LT_INIT()
 
+# Process user optons.
+
+# Does the user want to turn off all attempts to set fortran flags?
+# This is necessary to handle situations where the configure-provided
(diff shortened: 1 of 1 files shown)


edwardhartnett:
@underwoo any thoughts on this PR? If we can get it merged I'll try and get the other flags taken care of this week before I go on a long vacation...

edwardhartnett:
Thanks!

Issue #160, opened 2019-09-19
namelist files should not be CLEANed in Makefile.am of tests


edwardhartnett:
When I tried to push my changes to the configure.ac for #159 , I see this:

```
ed@mikado:~/noaa/FMS$ git commit -m "now check for bats utility" -a && git push
[ejh_test f17fc7f] now check for bats utility
 15 files changed, 6 insertions(+), 271 deletions(-)
 delete mode 100644 test_fms/axis_utils/input.nml
 delete mode 100644 test_fms/drifters/input.nml
 delete mode 100644 test_fms/fft/input.nml
 delete mode 100644 test_fms/field_manager/field_table
 delete mode 100644 test_fms/field_manager/input.nml
 delete mode 100644 test_fms/horiz_interp/input.nml
 delete mode 100644 test_fms/interpolator/diag_table
 delete mode 100644 test_fms/interpolator/input.nml
 delete mode 100644 test_fms/monin_obukhov/input.nml
 delete mode 100644 test_fms/mpp/input.nml
 delete mode 100644 test_fms/mpp/input.nml_base
 delete mode 100644 test_fms/mpp_io/input.nml
 delete mode 100644 test_fms/time_interp/input.nml
 delete mode 100644 test_fms/time_manager/input.nml
```

Well that's not good! What is happening is that the namelist files are listed as EXTRA_DIST (which is correct) and also as CLEANFILES (which is incorrect).

CLEANFILES should never list anything in EXTRA_DIST, that's going to break the build. What will happen is that developers will do a "make clean" in the course of their work, and then they will commit the results, and if they do not notice, they will commit a delete of all the namelist files, which will break their build.

Why were we trying to clean the namelist files? Are they changed during the tests?

underwoo:
Thank you for pointing this out.  What should be in each is a base namelist file (e.g. input.nml_base as in diag_manager), and the `input.nml` will be deleted, as it does need to change for each test --- a bad design, but we simply yanked these tests out of the main Fortran modulefiles and put them in their own test files; we do have plans to update the tests during our rewrite of the FMS library components.

We will work on correcting this.

edwardhartnett:
I have a PR that I'll be putting up.

If they change during testing, does that mean "make check" will fail if I run it twice?

edwardhartnett:
OK, just rereading your comment. Do you want the repo namelist to be called something like input_base.nml, and then have that copied to an input.nml which can be changed during testing and then cleaned?

[...]

PR #164, opened 2019-09-19
fixing namelist cleaning problem for axis_util

edwardhartnett:
OK, second attempt at this. My last PR got messed up because I ran the tests and that changed the input.nml file, which then attempted to commit. Whoops!

Here's a PR which fixes the namelist cleaning problem for axis_utils. @underwoo please take a look. If this is what you had in mind I will apply this to the other affected directories.

Part of #160


edwardhartnett:
OK, great. I am going to do one or more PRs for the other directories like this.

When I hit the sed problem you mentioned, I'll leave those directories until last. Let's get these simple ones fixed with changes like those in this PR.

PR #167, opened 2019-09-19
further application of the cleanfile fix for namelists...

edwardhartnett:
Part of #160 
Spreading proposed fix to other affected directories. Running through travis...


edwardhartnett:
OK, I will take this down and try again.

The problem is, once tests are run, then the input.nml changes, and it's hard to get a PR without those changes in it! ;-)

Issue #169, opened 2019-09-20
Test scripts need to include set -e


edwardhartnett:
You must have "set -e" at the top of all your test scripts. 

Default shell behavior when a command fails is to continue. Even if commands fail, the script can report success.

If you type `help set` in a bash shell you can see all the options:

`     -e  Exit immediately if a command exits with a non-zero status.
`
Here is a test script from netcdf-c which illustrates this. Note that we also sometimes use:

`-x  Print commands and their arguments as they are executed.
`

This is very helpful for finding out what went wrong. In fact, it's the first step in debugging a failure, so might as well just turn this on for the test.

```
#!/bin/sh

# This script runs some PnetCDF I/O tests

set -e
echo
echo "Testing file created with PnetCDF is modifiable with netCDF..."
./tst_pnetcdf

echo "Testing file created with PnetCDF works when adding variables..."
./tst_addvar tst_pnetcdf.nc

# We assume a min of at least 2 processors is available
mpiexec -n 2 ./tst_parallel2
```

edwardhartnett:
I did a branch with this, and found some more namelist problems, which I have addressed in a separate PR.

However, I also found that set -e seemed to catch some test failures. Also the use of the "skip" in bats seems to cause a non-zero exit code to be returned - the sign of an error. I would have thought that skip would return success, since it a test is intentionally being skipped, but that is not what happens.

After I get my PR with namelist fixes working, I will start adding set -e to the test scripts and see what happens.

edwardhartnett:
OK, looks like bats has been deprecated: https://github.com/sstephenson/bats/pull/269

There is something called bats-core that replaces it. 

What I would suggest instead is that bats be removed from the build and not used at all.

underwoo:
We have a goal to remove bats as a requirement for testing.  However, due
to other commitments, we have not been able to take on that task.  I have
opened an issue (#221) requesting this to be done.
--
Seth Underwood

S/W Development Process Lead
Modeling Systems Group
GFDL/NOAA/DOC
201 Forrestal Road
Princeton, NJ 08540-6649

(609) 452-5847 Office
(304) 376-9002 Cell
(609) 987-5063 Fax
Seth.Underwood@noaa.gov

edwardhartnett:
I can take it out of the existing code base, starting today. Will this be acceptable?

[...]

PR #170, opened 2019-09-20
more namelist cleanfile - (drifters, fft, field_manager, horiz_interp)

edwardhartnett:
Part of #160 

This PR includes PR #164 which has been reviewed and approved already (the changes to the axis_util directory to fix the namelist cleanfile problems).

In this PR I have applied these changes to the drifter and fft directories. At this point I stopped because I am not sure what is going on. The fft directory test uses an input.nml, but the file is already called input_base.nml, so what is the program using? (This could be failing in your tests but not be noticed by travis. See #169.


test_fms/fft/test_fft.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-cp $srcdir/test_fms/fft/input.nml input.nml
```
edwardhartnett: Note, there was no file called input.nml. It was already names input_base.nml, so this copy command must be failing in your test, when lead me to realize you need to do a set -e in every test script.

edwardhartnett: Seems like input.nml was created by the test, otherwise your make distcheck could not have worked...

edwardhartnett: OK, never mind. I think I was looking at artifacts of a previous attempt in this directory. Travis straightened me out.

edwardhartnett:
BTW we have seen some intermittant travis failures here which are pretty normal. One build failed because the install of the software timed out. Another time, Travis just got stuck, reporting the build as still running (yellow) but when you click through, the build had succeeded and exited, however travis does not notice the success.

These are normal with Travis. It's not perfect! But it's still great.

Developers can restart the tests by checking in more changes on the branch (as I just did). Admins can also restart tests on their travis dashboard without changes being needed.

edwardhartnett:
OK, great! Now Travis is happy and I'm happy with these 4 directories. ;-) @underwoo take a look at your convenience.

I will do more directories on a separate PR...

PR #171, opened 2019-09-20
Namelist fixes for interpolator, monin_obukhov, and mpp directories

edwardhartnett:
Part of #160 

Namelist fixes for interpolator, monin_obukhov, and mpp directories.


edwardhartnett:
OK, these look good. I'm going to do the rest in a new PR. These are a bit finicky!

PR #181, opened 2019-10-04
Improve autotools build process

### LINKED ISSUE #149: Have configure check for Fortran handling of cray pointers

### LINKED ISSUE #202: Travis CI `make distcheck` not passing configure options
**Describe the bug**
When testing the `--without-openmp` build in Travis CI, it was discovered the `make distcheck` does not pass the configure options to the distribution check call to configure.

**To Reproduce**
autoreconf -if
./configure --without-openmp
make distcheck

**Expected behavior**
Both builds should build without openmp support.

**System Environment**
All systems running `make distcheck` if passed in configure options

underwoo:
Updates to check for Fortran support of netcdf, mpi, openmp, cray pointers, and others.

This update will add the minimum required compiler flags to build the FMS library.

Also include many Fortran macros to expand the autoconf Fortran macros.  

Fixes #202 
Fixes #149


underwoo:
The only configure option not tested on Travis is `--without-mpi`.  I found the tests do not pass if libFMS is built without MPI support.  When someone is ready to fix the tests to pass without MPI, we can enable those tests.

thomas-robinson:
@underwoo a lot of tests are skipped anyway, and they have a comment about why they are skipped.  Is the *without MPI* test skipped?

underwoo:
The *without MPI* test isn't skipped, I didn't include it.

configure.ac
```diff
@@ -44,59 +58,122 @@ AC_PROG_CPP
...
 AC_PROG_INSTALL
 
 # Is bats installed? It is required for testing.
-AC_CHECK_PROGS([FMS_BATS], [bats])
-if test -z "$FMS_BATS"; then
-   AC_MSG_ERROR([Cannot find bats utility. Install bats and try again.])
+AC_CHECK_PROGS([BATS], [bats])
+AM_CONDITIONAL([MAKE_CHECK_ENABLED], [test -n "$BATS"])
+if test -z "$BATS"; then
+   AC_MSG_WARN([Cannot find the bats utility. All checks will be disabled.])
```
edhartnett: This is a poor idea. You are encouraging users to build your package without running tests.

Is the minor benefit of bats worth that?

configure.ac
```diff
@@ -44,59 +58,122 @@ AC_PROG_CPP
...
 # Define __APPLE__ macro if on a Apple/Mac OS X
 AS_IF([test "x$target_vendor" = "xapple"], [AC_DEFINE([__APPLE__], [ ])])
-
-# Define an AM_CONDITIONAL to determine if you are on a CRAY
-AM_CONDITIONAL([CRAY], [test `env | grep CRAY | wc -l` -gt 0])
-
 # Define an AM_CONDITIONAL to determine if you are on an Apple/Mac OS X
 AM_CONDITIONAL([APPLE], [test "x$target_vendor" = "xapple"])
+# Define an AM_CONDITIONAL to determine if you are on a CRAY
+AM_CONDITIONAL([CRAY], [test `env | grep CRAY | wc -l` -gt 0])
```
edhartnett: Why do you need to know if you are on a cray?

bensonr: The static libraries are still the default on the Cray machines at NCRC.  The macro is needed by the various unit test Makefiles to create the correct library link specification.  The macro name could be changed to be more explicit, such as LINK_STATIC_LIB.  But to date, this is a machine-specific test.

configure.ac
underwoo: We still have some of our systems that do not have bats, and I don't want our users (who adopt the autotools build) to not be able to use this build method, which is why I decided to allow this.

I am investigating how to deal with the tests w/o bats, and if possible keep the TAP output. When we decide on how to do the testing w/o bats, we will remove this dependency.

configure.ac
edhartnett: OK, this is incorrect for several reasons and should not be merged.

In this case, tell the builders on the Cray machines at NCRC to use --disable-shared. (Also tell them to fix their installs; it's not hard.)

So no need to change the build system, this is already fully supported. You're done! ;-)

Secondly, even if this was necessary, there is no need to make this machine specific. Instead, always define a more explicit macro. Otherwise you are encouraging lazy programming and brittle build systems. (Because other programmers will see that you have CRAY defined, and will start using it. Now your changes, and theirs, will not work for non-cray systems that share the build characteristic you are dealing with. Everything will become machine-specific and non-portable.)

After defining a specific macro to cover the build feature you need to determine, go one step further, and write an autoconf test which works on all machines and resolve the question of how to set your explicit macro. (I can help here but Seth is also well-positioned to tell how to do this.)

Then you will have a fully-portable solution that will work on new environments, not just one cray at NCRC. (If these changes are merged, it will break shared library installs for all other crays, which have installed shared libraries correctly. That's certainly not a good idea.)

Please never check in any part of this build system that relies on machine-specific rules. You will be breaking portability for NOAA and other users, including at NCRC, when they get a new Cray or fix installs on their existing one.

configure.ac
edhartnett: OK, how about we change autoconf right now to warn you that make check will not work without bats. (But make all still will work of course).

Then, machines that don't have bats will not be able to run tests, but will still be able to build the code.

We previously discussed making bats optional. The more you have to distort your build to support bats, the less cost-effective it is. How about we change to use bats optionally, that is, use it when present, and don't when not, and determine whether it is present in configure?

So instead of checking in and maintaining these changes, how about we make bats optional? This can be done when I get back from vacation, and meanwhile make check will not work, but make all will.

m4/gx_fortran_options.m4
```diff
@@ -255,23 +255,28 @@ AC_LANG_POP([Fortran])
...
+_AC_FORTRAN_ASSERT()dnl
 m4_ifval([$2],[gx_fc_check_mod_only=",only:$2"],[gx_fc_check_mod_only=""])
 AS_LITERAL_WORD_IF([$1],
   [AS_VAR_PUSHDEF([gx_mod], [gx_cv_fc_check_mod_$1])],
   [AS_VAR_PUSHDEF([gx_mod], AS_TR_SH([gx_cv_check_mod_$1]))])
+dnl Autoconf does not pass CPPFLAGS to the Fortran tests.  As the user may
+dnl define the include options in CPPFLAGS instead of FFLAGS or FCFLAGS, we
+dnl force CPPFLAGS te be part of FFLAGS/FCFLAGS
+gx_save_[]_AC_LANG_PREFIX[]FLAGS=${[]_AC_LANG_PREFIX[]FLAGS}
+_AC_LANG_PREFIX[]FLAGS="${[]_AC_LANG_PREFIX[]FLAGS} $CPPFLAGS"
```
edhartnett: Disagree with this plan, it just adds confusion.

Instead, properly educate your users on FCFLAGS vs. CPPFLAGS.

underwoo: I agree that in this case (when checking for Fortran `.mod` files) adding `CPPFLAGS` to `FCFLAGS` or `FFLAGS` is not a solution for all Fortran packages.  However, in our case, all Fortran files are run through the preprocessor.  Automake will pass `CPPFLAGS` when compiling preprocessed Fortran.  Thus in this case, we are simply forcing the configure tests to mimic the behavior of automake.

underwoo: Yes, but in many cases users will have already added the `-I...` compile directive for libraries (like netCDF) in the `CPPFLAGS` for C.  On many of our systems, the C and Fortran include files are in the same directories.  Users can use one or both. My goal was to have the configure tests mimic what automake does.

edwardhartnett:
I applaud the idea of having standardized spacing and indentation. 

I suggest that you make a PR with only whitespace changes, to change all files at once. Since such a PR does not change code, it need not be thoroughly reviewed. 

Then later code changes can be easily reviewed because they will not include whitespace cleanup changes.

If we are all using emacs, then you will find that your changes to Makefile.am spaces before the continue backslash are not need. Use M-x q to properly format and indent any line or set of continued lines in the Makefile.am. You will easily achieve a nice-looking file with no effort.

On other (Java) projects, we have included a re-formatter that is automatically applied to all code, to give a consistent look.

.travis.yml
```diff
@@ -18,9 +18,10 @@ addons:
     - pkg-config gfortran netcdf-bin libnetcdf-dev libnetcdff-dev openmpi-bin libopenmpi-dev
 
 env:
-  - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8' OPENMP=''
-  - CPPFLAGS_ADD='-DOVERLOAD_R4 -DOVERLOAD_R8' FCFLAGS_REAL_KIND='' OPENMP=''
-  - CPPFLAGS_ADD='' FCFLAGS_REAL_KIND='-fdefault-real-8 -fdefault-double-8' OPENMP='-fopenmp'
+  - FCFLAGS_ADD='' CONFIGURE_OPTIONS=''
+  - FCFLAGS_ADD='' CONFIGURE_OPTIONS='--without-openmp'
```
thomas-robinson: The build for this compiles with -fopenmp.  It might be better to add -fno-openmp to the `FCFLAGS`

underwoo: The latest push in this PR resolves this.

edwardhartnett:
Can this be merged soon?

PR #189, opened 2019-10-15
Fix parameter out of range setting for mixed-mode compiles

### LINKED ISSUE #188: time_interp_missing is out of range in mixed mode
**Describe the bug**
A parameter in [time_interp/time_interp_external.F90](time_interp_time_interp_external.F90) is initialized out of range when in mixed-mode because all reals (including literal constants) are `real(kind=4)`.  

**To Reproduce**
Compile FMS in mixed mode (`configure` option `--enable-mixed-mode` in #181).  The compile will fail unless the GFortran option `-fno-range-check` is given in `FCFLAGS`. 

**Expected behavior**
The compile to build successfully with `configure --enable-mixed-mode` without giving any additional compile flags in `FCFLAGS`

**System Environment**
- OS: All tested (ubuntu 18, RHEL 7)
- Compiler(s): GFortran
- Configure options: `configure --enable-mixed-mode`

**Additional context**
This bug does not, by default, cause the builds with the Intel compilers to fail.  However, I suspect if given the proper `ifort` flags, warnings will be printed.

underwoo:
Fixed time_interp_missing parameter in time_interp_external.F90 to be within range when compiled in mixed-mode.

Fixes #188

### DIFF
--- a/time_interp/time_interp_external.F90
+++ b/time_interp/time_interp_external.F90
@@ -137,7 +137,8 @@ module time_interp_external_mod
   type(ext_fieldtype), save, private, pointer :: field(:) => NULL()
   type(filetype),      save, private, pointer :: opened_files(:) => NULL()
 !Balaji: really should use field%missing
-  real(DOUBLE_KIND), private, parameter :: time_interp_missing=-1e99
+  integer, private, parameter :: dk = DOUBLE_KIND
+  real(DOUBLE_KIND), private, parameter :: time_interp_missing=-1e99_dk
   contains
 
 ! <SUBROUTINE NAME="time_interp_external_init">


time_interp/time_interp_external.F90
```diff
@@ -137,7 +137,8 @@ module time_interp_external_mod
   type(ext_fieldtype), save, private, pointer :: field(:) => NULL()
   type(filetype),      save, private, pointer :: opened_files(:) => NULL()
 !Balaji: really should use field%missing
-  real(DOUBLE_KIND), private, parameter :: time_interp_missing=-1e99
+  integer, private, parameter :: dk = DOUBLE_KIND
+  real(DOUBLE_KIND), private, parameter :: time_interp_missing=-1e99_dk
```
thomas-robinson: if `DOUBLE_KIND` = 4, then won't this still cause an overflow?  Am I missing something here?

underwoo: You are correct.  However, `DOUBLE_KIND` should not equal 4 in any of our cases.

Issue #192, opened 2019-10-15
Ensure variables initialized before use


underwoo:
The `-Wmaybe-uninitialized` GFortran option will list many variables that may be used before initialized.  While several of these warnings are likely false positives, that is most likely due to the branching done in the code causing certain, unlikely paths (and sometime error paths) that do not initialize the variable.  These cases need to be evaluated, and if possible corrected.

PR #193, opened 2019-10-15
Initialze variables before use

underwoo:
Ensure several Fortran variables in mpp are initialized before used.

This is to clear up several warnings issued by Gfortran when given `-Wall`.

Start of #192

### DIFF
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -100,14 +100,14 @@
     integer :: stdlog,istat
     logical :: opened
     character(len=11) :: this_pe
-!$  logical           :: omp_in_parallel    
+!$  logical           :: omp_in_parallel
 !$  integer           :: omp_get_num_threads
 !$  integer           :: errunit
 
 
 !NOTES: We can not use mpp_error to handle the error because mpp_error
 !       will call stdout and stdout will call stdlog for non-root-pe.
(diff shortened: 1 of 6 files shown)


mpp/include/mpp_util.inc
```diff
@@ -124,6 +124,8 @@
 !$OMP end single
 !$  endif
 
+    ! Set to negative value for default/bad unit
+    stdlog = -1
```
bensonr: To me, this is a bug in gfortran.  In every case, stdlog will be defined on exit from the routine and there is no attempt to use stdlog prior to it being set.

mpp/include/mpp_util.inc
```diff
@@ -1091,7 +1093,9 @@ end function rarray_to_char
     integer,save :: i
     logical      :: l_open
 
-!  9 is reserved for etc_unit
+    ! Negative value for default/invalid unit number
+    get_unit = -1
```
bensonr: Again, I see this as a gfortran bug.  As above with stdlog get_unit will either return a number between 10 and 99 or throw a fatal error halting execution.  So I don't see how this is a case of trying to use something before it is defined.

underwoo:
Closing this PR for now.  Hopefully the IO rewrite will take care of some of these warnings.  If not, the work to clear out the warnings should be done in the updated code.

PR #199, opened 2019-11-01
Remove unused components - fft, station_data, diag_manager/diag_manifest.F90

underwoo:
Removed some components not used in current GFDL models.  These include:

* fft
* station_data
* diag_manager/diag_manifest.F90

This PR should only be taken after #181.


thomas-robinson:
@underwoo this pull request does much more than what the description implied.  For example, I looked at the [travis build](https://travis-ci.com/NOAA-GFDL/FMS/jobs/251968057) with the option `--without-openmp` and everything is compiled with `-fopenmp`.  Would it make more sense to break this up?

underwoo:
@thomas-robinson it does.  This PR needs to be taken after #181.  I should have said that in the description.

underwoo:
Thank you for pointing this out.  It appears the `make distcheck` option does not pass the original configure options options to the distcheck call of `configure`.  From reading the documentation, it looks like setting the variable `DISTCHECK_CONFIGURE_FLAGS` to the same options that will be passed to configure will resolve the problem.  I'll get a PR in to address this.

Note, the first build of FMS does honor the `--without-openmp` configure option.

configure.ac
```diff
@@ -24,17 +25,30 @@ LT_PREREQ([2.4])
...
-    [Disable any attempts by configure to set fortran flags. With this option, only user-provided flags (set in FCFLAGS) will be used compiling Fortran code.])])
-test "x$enable_fortran_flag_setting" = xno || enable_fortran_flag_setting=yes
-AC_MSG_RESULT([$enable_fortran_flag_setting])
+AC_ARG_ENABLE([mixed-mode],
+  [AS_HELP_STRING([--enable-mixed-mode],
+    [Build using mixed mode.  Enables both 64-bit and 32-bit reals in Fortran. This option will be ignored if --disable-fortran-flag-setting is also given.])])
+AS_IF([test ${enable_mixed_mode:-no} = no],
+  [enable_mixed_mode=no],
+  [enable_mixed_mode=yes])
+AC_ARG_WITH([mpi],
```
edwardhartnett: This is not the correct usage of the --with- feature for autoconf.

Use --with-XXX when XXX is a library that can optionally be used, and you may want to specify an alternate location.

This is not the case with MPI. Firstly, isn't MPI always required? Secondly, using MPI involves more than just the lib and include directories, additionally the compiler is different. This was a dumb decision by MPI but we are stuck with it.

So when the user says CC=mpicc ./configure --with-mpi=/some/other/location you will not be able to meet that request, since you don't control mpicc and where it looks for MPI.

What are you trying to achieve with --with-mpi here? Is it possible to build this code without MPI? If so, we can check for an MPI function in configure.ac, and set a flag accordingly. The user does not need to do anything except set FC/CC to MPI compilers.

configure.ac
```diff
@@ -24,17 +25,30 @@ LT_PREREQ([2.4])
...
+  [AS_HELP_STRING([--with-mpi],
+    [Build with MPI support.  This option will be ignored if --disable-fortran-flag-setting is also given. (Default yes)])])
+AS_IF([test ${with_mpi:-yes} = yes],
+  [with_mpi=yes],
+  [with_mpi=no])
+AC_ARG_WITH([openmp],
```
edwardhartnett: I think you want --enable-openmp here, not --with-openmp here.

configure.ac
```diff
@@ -44,59 +58,122 @@ AC_PROG_CPP
...
 
 # Find the install program.
 AC_PROG_INSTALL
 
 # Is bats installed? It is required for testing.
-AC_CHECK_PROGS([FMS_BATS], [bats])
-if test -z "$FMS_BATS"; then
-   AC_MSG_ERROR([Cannot find bats utility. Install bats and try again.])
+AC_CHECK_PROGS([BATS], [bats])
+AM_CONDITIONAL([MAKE_CHECK_ENABLED], [test -n "$BATS"])
```
edwardhartnett: Not clear what MAKE_CHECK_ENABLED would be for. Also this would not be the correct way to set it. Also also, we are eliminating the BATS check.

configure.ac
```diff
@@ -44,59 +58,122 @@ AC_PROG_CPP
...
+# Require netCDF
+GX_FORTRAN_CHECK_HEADERS([netcdf.inc], [], [AC_MSG_ERROR([Can't find the netCDF Fortran header file.  Set CPPFLAGS/FCFLAGS])])
+GX_FC_CHECK_MOD([netcdf], [], [], [AC_MSG_ERROR([Can't find the netCDF Fortran module.  Set CPPFLAGS/FCFLAGS])])
+GX_FORTRAN_SEARCH_LIBS([nf_create], [netcdff], [@%:@include <netcdf.inc>], [iret = nf_create('foo.nc', 1, ncid)], [],
+  [AC_MSG_ERROR([Can't find the netCDF Fortran library.  Set LDFLAGS/LIBS])])
+GX_FORTRAN_SEARCH_LIBS([nf90_create], [netcdff], [use netcdf], [iret = nf90_create('foo.nc', 1, ncid)], [],
+  [AC_MSG_ERROR([Can't find the netCDF Fortran library.  Set LDFLAGS/LIBS])])
+
+# Check if Fortran compiler has Cray pointer support
+GX_FC_CRAY_POINTER_FLAG()
```
edwardhartnett: If this and following functions set flags, you want to protect them by checking enable_setting_flags.

configure.ac
```diff
@@ -44,59 +58,122 @@ AC_PROG_CPP
...
 # Define __APPLE__ macro if on a Apple/Mac OS X
 AS_IF([test "x$target_vendor" = "xapple"], [AC_DEFINE([__APPLE__], [ ])])
-
-# Define an AM_CONDITIONAL to determine if you are on a CRAY
-AM_CONDITIONAL([CRAY], [test `env | grep CRAY | wc -l` -gt 0])
-
 # Define an AM_CONDITIONAL to determine if you are on an Apple/Mac OS X
 AM_CONDITIONAL([APPLE], [test "x$target_vendor" = "xapple"])
+# Define an AM_CONDITIONAL to determine if you are on a CRAY
+AM_CONDITIONAL([CRAY], [test `env | grep CRAY | wc -l` -gt 0])
```
edwardhartnett: I cannot overstate what a terrible idea this is. Do NOT check what kind of system you are on, and write code based on that. Instead check for the features you need, and program based on that.

What is it about a Cray that makes you think you need to know when you are compiling on the cray? What specifically?

All Crays are not the same. And other systems may also exhibit the same problem you are coding around. So figure out how to test whether the system supports what you need, and ignore the type of system. Ask for help if you need it to create the test.

configure.ac
underwoo: You are likely correct that `--enable-` is a better choice than `--with-`, especially since openMP is not something managed in a library separate from the compiler.  I can make this change in a new commit.

configure.ac
underwoo: The library can be built without MPI, and is done so in some of the GFDL idealized test cases (used internally by some GFDL scientists).  It is also possible to build with MPI without using the MPI compiler wrappers (`mpicc`, `mpif90`, etc.).  In fact, that was the standard way we built the library using our older build process (using `mkmf`).  At this time, the only valid responses for `--with-mpi` is `yes` or `no`.  I probably should have included a check for anything other than those two responses.  It may be possible in a future commit to make use of a path, but we will not support this usage ATT.

configure.ac
underwoo: This was a way to _not_ allow `make check`, or more specifically have `make check` do a no-op.  As we are removing `bats` from the check scripts, this section will no longer be needed.  (This commit was done before the `bats` removal changes started.

configure.ac
edwardhartnett: OK, then you should use --enable and not --with for this.

configure.ac
underwoo: Here that restriction is not needed.  The flag required to support cray pointer is only added to `FCFLAGS` if we are setting the flags for the user (see lines 135 and 156).

However, something should be done to catch the case where the user has not supplied the required flag(s).  That can be added in a future commit.

configure.ac
underwoo: We are looking for was to better implement these checks.  Some of it requires understanding the requirement(s) that cause things to not build or work properly on these specific platforms/systems.  For the Apple case, we know it is a missing C call that is different on the Darwin/Mac OS systems.  It may affect other BSD-style OS's, but we do not yet know, but at the time the problem was discovered we only found it on Mac OS systems.

The Cray case is more difficult to discover via other tests.  It has to do with the use of shared libraries on systems that use the Cray Programming Environment (PrgEnv).  We are still investigating ways to reliably capture the problem via configure tests.  Not all Cray systems use the Cray PrgEnv.  Those systems should not cause this conditional to be set.  Once we have a reliable way to capture this, this will be updated.

edwardhartnett: OK, is this only a problem for shared library builds?

There is already a --disable-shared option automatically built into configure.

Suggest you change the name of this from CRAY to something like CRAY_PRGENV_SHARED (or anything more specific). Otherwise, another programmer is going to come along, and see "CRAY" and start writing code based on that. Which is what you don't want. The more specific you make your macro names, the less temptation to misuse them later.

Also, from what you say, there will be some Cray systems that will not set CRAY because they don't have this issue. So that's a confusing macro name.

wrongkindofdoctor:
@underwoo I can merge this in as soon as you resolve the merge conflicts.

thomas-robinson:
@underwoo I looked at the conflict in diag_manager and it's on line  [232/233](https://github.com/NOAA-GFDL/FMS/blob/master/diag_manager/diag_manager.F90#L232).  You want manually fix this one.  
```fortran
 232        & max_file_attributes, max_axis_attributes, prepend_date, DIAG_FIELD_NOT_FOUND, diag_init_time, diag_data_init,&
 233        & write_manifest_file, fileobj, fileobjU, fnum_for_domain, fileobjND
```
You removed the `write_manifest_file` but I added `fileobj, fileobjU, fnum_for_domain, fileobjND`.  Please keep what I added.

PR #207, opened 2019-11-06
now check for sed and ncgen in configure.ac

### LINKED ISSUE #205: presence of sed and ncgen need to be tested in configure.ac
In our test scripts we use sed and (in test_drifters_input.sh) ncgen.

We need to confirm these are available in configure.ac. By confirming it there (as we do with bats) we prevent confusion and future (unnecessary) bug reports.

edwardhartnett:
Fixes #205 

Check for sed and ncgen in configure. If not found, give the user and error message and stop.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -61,7 +61,7 @@ if test -z "$FMS_SED"; then
    AC_MSG_ERROR([Cannot find sed. Install sed and try again.])
 fi
 
-# Is ncget installed? It is required for testing.
+# Is ncgen installed? It is required for testing.
 AC_CHECK_PROGS([FMS_NCGEN], [ncgen])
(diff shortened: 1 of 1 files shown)


underwoo:
@wrongkindofdoctor this PR is ready to go.

configure.ac
```diff
@@ -55,6 +55,18 @@ if test -z "$FMS_BATS"; then
    AC_MSG_ERROR([Cannot find bats utility. Install bats and try again.])
 fi
 
+# Is ncgen installed? It is required for testing.
+AC_CHECK_PROGS([FMS_SED], [sed])
```
edwardhartnett: Well my comments were incorrect. I have fixed and pushed changes.

PR #211, opened 2019-11-14
Ouputs the chksum as a 16 length character

uramirez8707:
Changes the compute_global_checksum routine to output the checksum as a 16 length character instead of trimming the empty spaces.


uramirez8707:
@wrongkindofdoctor

PR #222, opened 2019-11-26
demonstration of first stage of bats removal...

edwardhartnett:
In this PR I demonstrate how shell scripts may be used instead of bats, selectable by configure.

The configure option --enable-shell-tests will cause the tests in test_fms/axis_utils to run from a shell script, not a bats script. Running without --enable-shell-tests will cause bats to be used as usual.

The intention is to move all testing to shell scripts. Once all are satisfied that all tests have been moved, we can then remove bats, and remove the --enable-shell-tests option.

This PR includes my other outstanding PRs, to get through my own FMS CI system.


edwardhartnett:
OK, this shows the first stage of bats removal for axis_utils and diag_manager test directories.

Existing bats-running shell scripts must be changed to *.test instead of *.sh because of automake LOG_DRIVER restrictions, they must have different file extensions than the new shell tests.

Take a look and if this looks good I can quickly apply this to the other directories and then bats can be removed.

edwardhartnett:
OK, now this PR is complete. All tests run under bats can now be run by shell scripts instead with --enable-shell-tests at configure.

Next step is to remove the bats stuff. I will do that in a separate PR.

edwardhartnett:
This should not be merged, instead merge #228 
This remains open for ease of inspection.

PR #232, opened 2019-12-02
Fix CPP warning

DusanJovic-NOAA:
CPP macro should not be defined with different value if it's already defined. It should be undefined first.


underwoo:
Thank you @DusanJovic-NOAA for this PR.  These fixes are something that I have already done (which will be part of another pull request in the future).  However, we will get these updates in soon.

PR #247, opened 2019-12-20
Remove use_netCDF macro

### LINKED ISSUE #246: Remove preprocessor macro use_netCDF
Using netCDF is now required to build libFMS.  This macro should be removed so it is not needed when building libFMS.

No `use_netCDF` macro in any FMS sources.

underwoo:
**Description**

libFMS can now _only_ be built with netCDF, thus the use_netCDF preprocessor macro is not needed.  This commit removes the use_netCDF macro from all FMS sources.

Fixes #246 

**How Has This Been Tested?**
Standard build/make distcheck on Apple Mac OS X system.

**Checklist:**
- [X] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_io_misc.inc
+++ b/mpp/include/mpp_io_misc.inc
@@ -177,10 +177,8 @@
       if( pe.EQ.mpp_root_pe() )then
           iunit = stdlog()  ! PGI compiler does not like stdlog() doing I/O within write call
           write( iunit,'(/a)' )'MPP_IO module '//trim(version)
-#ifdef use_netCDF
           text = NF_INQ_LIBVERS()
           write( iunit,'(/a)' )'Using netCDF library version '//trim(text)
-#endif
       endif
 
 #ifdef CRAYPVP
@@ -234,12 +232,10 @@
       do unit = unit_begin,unit_end
          if( mpp_file(unit)%opened )close(unit)
       end do
-#ifdef use_netCDF
 !close all open netCDF units
       do unit = maxunits+1,2*maxunits
          if( mpp_file(unit)%opened )error = NF_CLOSE(mpp_file(unit)%ncid)
       end do
-#endif
 
 !      call mpp_max(mpp_io_stack_hwm)
 
@@ -262,7 +258,6 @@
       character(len=*), optional :: string
       character(len=256) :: errmsg
 
-#ifdef use_netCDF
       if( err.EQ.NF_NOERR )return
       errmsg = NF_STRERROR(err)
       if( PRESENT(file) )errmsg = trim(errmsg)//' File='//file%name
@@ -272,7 +267,6 @@
       if( PRESENT(string) )errmsg = trim(errmsg)//string
       call mpp_io_exit('NOSYNC')        !make sure you close all open files
       call mpp_error( FATAL, 'NETCDF ERROR: '//trim(errmsg) )
-#endif
       return
     end subroutine netcdf_err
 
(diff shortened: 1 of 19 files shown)


mpp/include/mpp_io_misc.inc
```diff
@@ -177,10 +177,8 @@
       if( pe.EQ.mpp_root_pe() )then
           iunit = stdlog()  ! PGI compiler does not like stdlog() doing I/O within write call
           write( iunit,'(/a)' )'MPP_IO module '//trim(version)
-#ifdef use_netCDF
           text = NF_INQ_LIBVERS()
           write( iunit,'(/a)' )'Using netCDF library version '//trim(text)
-#endif
       endif
 
 #ifdef CRAYPVP
```
bensonr: IIRC, this requires the use of the FFIO system and I don't think that is supported anymore.

test_fms/fms/test_unstructured_fms_io.F90
```diff
@@ -48,9 +48,7 @@ program test_unstructured_fms_io
                                             fms_io_exit
     implicit none
 
-#ifdef use_netCDF
 #include <netcdf.inc>
-#endif
```
bensonr: If this is the only place the macro exists in this file, I don't know that we even need the #include

test_fms/mpp_io/test_mpp_io.F90
```diff
@@ -39,9 +39,7 @@ program test
 
   implicit none
 
-#ifdef use_netCDF
 #include <netcdf.inc>
-#endif
```
bensonr: If this is the only place the macro exists in this file, I don't know that we even need the #include

mpp/include/mpp_io_misc.inc
underwoo: I still see the `assign` command and library call in the Cray PE (Fortran Reference Manual)[https://pubs.cray.com/content/S-3901/9.1/cray-fortran-reference-manual/enhanced-io-using-the-assign-environment].  However, I was unable to test if it works due to CCE fortran not linking, and the latest CCE version on Gaea does not have the Fortran compiler.  This is something we should look into.  I'll create a new issue.

test_fms/fms/test_unstructured_fms_io.F90
underwoo: I'll make sure this line is/is not needed, and delete if not.  Thank you.

test_fms/mpp_io/test_mpp_io.F90
underwoo: I'll make sure this line is/is not needed, and delete if not. Thank you.

test_fms/fms/test_unstructured_fms_io.F90
underwoo: Not needed.  New version pushed.

test_fms/mpp_io/test_mpp_io.F90
underwoo: Not needed.  New version pushed.

mpp/include/mpp_io_misc.inc
underwoo: Issue #248 created.

bensonr:
Although assign is still provided via the Cray Compiler Environment, it
looks like the specific use here is no longer supported.  The assign pages
you point to for CCE 9.1 no longer shows a -P option for the assign command.

underwoo:
> Although assign is still provided via the Cray Compiler Environment, it looks like the specific use here is no longer supported. The assign pages you point to for CCE 9.1 no longer shows a -P option for the assign command.

Thank you @bensonr for looking closer.  I'll be sure to add this information to #248.

PR #249, opened 2019-12-20
Improve Mac OS support and remove __APPLE__ from Fortran sources

### LINKED ISSUE #241: Instead of testing for APPLE, test for absence of function you care about...
**Describe the bug**
Checking for specific systems (like APPLE or CRAY) breaks the portability of the build system. The entire point of autotools is that you never, ever, should do this. If you find yourself identifying by system type, you are making a mistake, and need to give your changes more thought.

**To Reproduce**
Build on a non-apple system that also does not have the memuse() function. Your build will break.

**Expected behavior**
The build system can check for memuse(). If it is not present, the build can provide it. This will work on all apples, and all other systems as well.

**Additional context**
Once you define APPLE in your build, lazy programmers will start using it for other issues that also occur on apples. Soon you will not be able to easily fix this, as there will be more then one incompatibility wrapped up in the handling of APPLE. If we use the tools correctly, and test for the memuse() function, then no other programmer can use that test unless they also are interested in the memuse function. If another programmer finds a build problem on an apple that programmer will be forced to confront the problem correctly, instead of piggy-backing on another solution.

The problem is here, in configure.ac:
```
...

```

Then, in the makefiles of libFMS and memutils, we have something like this:

```
if APPLE
libFMS_la_LIBADD += ${top_builddir}/memutils/libmemuse_c.la
endif

```

Instead, use AC_CHECK_FUNCS (https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Generic-Functions.html). Like this:

```
# See if memuse() function is present on this system. This will set HAVE_MEMUSE if 
# memuse() is found.
AC_CHECK_FUNCS(memuse)

```

Then, in your memuse.c file, add an ifndef to check HAVE_MEMUSE. Now the Makefile.am files do not need to care about this issue, so AM_CONDITIONAL and references to it in the Makefile.am files can be removed.

Shall I submit a PR for this?

underwoo:
This commit removes all __APPLE__ macros from Fortran sources.  This fix also removes a call to the deprecated `syscall` on Mac OS X.

`__APPLE__` for the most part was used for the memory usage routines.  This fix confines all get memory calls to C, and changes mpp_print_memuse_stats to be the only location where the C function is called.  (Note, AFAIK the function to get the system memory is portable to most *nix systems.)

The `syscall` function is deprecated in Mac OS.  The affinity.c source was updated to not call `syscall` on Mac OS systems.

Fixes #241 

**How Has This Been Tested?**
Tested on Mac OS X, using `make distcheck`.

Tested using SPEAR model.

**Checklist:**
- [x] `make distcheck` passes


underwoo:
Unfortunately, it appears something in this PR may be wrong, as a test with a full coupled model is failing.  I'll do more investigating.

mpp/mpp_memuse.c
```diff
@@ -0,0 +1,94 @@
...
+
+#else
+#error "Cannot define getPeakRSS( ) or getCurrentRSS( ) for an unknown OS."
+#endif
+
+/* Ensure getpeakrss function has correct name mangling for being called from
+   Fortran.
+   The outer #ifdef FC_FUNC it for use with mkmf and other non-Autotools builds. */
+#ifdef FCFUNC
+#define GETPEAKRSS_FC FC_FUNC (getpeakrss, GETPEAKRSS)
```
bensonr: FCFUNC and FC_FUNC do not agree

test_fms/mpp/test_mpp_memutils_begin_2x.F90
```diff
@@ -0,0 +1,33 @@
...
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+!***********************************************************************
+program test_mpp_memutils_init_end
+#include "../../include/fms_platform.h"
+
+  use mpp_mod, only : mpp_init, mpp_exit
+  use mpp_memutils_mod, only: mpp_memuse_begin, mpp_memuse_end
+  implicit none
+
+  real, dimension(:), allocatable :: ralloc_mem
```
GFDL-Eric: Is this supposed to be allocated and set to 0.0 as in the begin_end test?

underwoo: Looking again, no.  This test case is a _failure_ test.  The program should exit non-zero when `mpp_memuse_begin()` is called multiple times.  I have updated the documentation in the source file to reflect this.

underwoo:
Further testing indicated the failure in the coupled model was not due to the code in this PR, but an error in one of the components -- which will be corrected in that component's repository.

mcgibbon:
The current 2020.01 also won't build inside docker ubuntu. Your PR fixes that, so thanks!

underwoo:
Thank you @mcgibbon for your feedback.  To clarify, when you say "2020.01" are you referring to the "2020.01-alpha1" tag? 

This PR will be merged soon (within the next week, we hope).  The hold up is due to several GFDL scientists still using the old (`mkmf`) build system, and we need to ensure they can continue to build using that until we can get them to use autotools.

mcgibbon:
I mis-typed, was using the head of the 2019.01 branch.

underwoo:
@mcgibbon that is likely due to missing compiler flag or other settings that were required in the autotools configuration in 2019.01, as we build all PR code on Travis-CI using an ubuntu system.  

Another PR (#181), which will also be included in the next release (2020.01) will also improve the autotools build experience.  That should also be merged in the next few weeks.

mcgibbon:
Possibly, I'm not passing any particular flags or settings into the build system. The error was triggering due to one of these `__APPLE__` flags re-defining a routine that was already defined in an included header.

wrongkindofdoctor:
@underwoo Please fix the merge conflicts that resulted from merging previous PRs.

PR #251, opened 2019-12-23
Remove string overflow warning in field_manager

### LINKED ISSUE #250: field_manager: Warning: reading 128 bytes from a region of 127
**Describe the bug**
When building field_manager with gfortran a warning is printed:

```
../../field_manager/field_manager.F90:1196:0:

 1196 |       val_name = val_name(2:length-1)
      |
Warning: '__builtin_memmove' reading 128 bytes from a region of size 127 [-Wstringop-overflow=]
```

It would be good to remove this warning.

**To Reproduce**
Compile libFMS with a gfortran compiler.

**Expected behavior**
No warning like this when compiling field_manager should be printed.

**System Environment**
All systems using gfortran to build libFMS

underwoo:
**Description** 

Gfortran would issue a warning of reading 128 bytes from a region of 127 in field_manager.  This update removes that warning

Fixes #250 

**How Has This Been Tested?**
Tested on Mac OS X, `make distcheck`

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/field_manager/field_manager.F90
+++ b/field_manager/field_manager.F90
@@ -1193,7 +1193,7 @@ do i = 1, num_elem
   if (val_name(1:1) .eq. squote) then  !{
 
     if (val_name(length:length) .eq. squote) then
-      val_name = val_name(2:length-1)
+      val_name = val_name(2:length-1)//repeat(" ",len(val_name)-length+2)
       val_type = string_type
     elseif (val_name(length:length) .eq. dquote) then
       call mpp_error(FATAL, trim(error_header) // ' Quotes do not match in ' // trim(val_name) //       &


field_manager/field_manager.F90
```diff
@@ -1193,7 +1193,7 @@ subroutine new_name ( list_name, method_name_in , val_name_in)
   if (val_name(1:1) .eq. squote) then  !{
 
     if (val_name(length:length) .eq. squote) then
-      val_name = val_name(2:length-1)
+      val_name = val_name(2:length-1)//repeat(" ",len(val_name)-length+2)
```
thomas-robinson: This seems like a fancy way of saying ```//"  "```

wrongkindofdoctor: The repeat ensures that the de-quotation-marked string has 2 blank spaces appended to it to match original string size. `//" "` would only append 1 blank space. `//" "//" "` would achieve the same result, but is less explicit, and more error-prone (easy to accidentally add too many spaces in between quotes).

PR #273, opened 2020-01-30
User/jml/add check for append mode

wrongkindofdoctor:
**Description**
This fix adds a logical attribute "mode_is_append" to the domain-decomposed netcdf file object type, and a check for whether this attribute is .true. to permit registering axes to the file object without writing the axes to the file itself when it is opened in "append" mode. This change supports cases when file creation and axis registration occur separately from data write procedures.

**How Has This Been Tested?**
This has been testing with MOM6_solo regression test cases tc0--tc3 in the .testing suite compiled with gcc9.2 on Skylake processors, and gcc7.2, and intel19 on Gaea

**Checklist:**
- [x] `make distcheck` passes


fms2_io/netcdf_io.F90
```diff
@@ -404,6 +405,7 @@ function netcdf_file_open(fileobj, path, mode, nc_format, pelist, is_restart) &
   character(len=256) :: buf
   logical :: is_res
 
+  fileobj%is_root = .false.
```
thomas-robinson: Won't this set the root to `.false.`?  This seems out of place for checking if the file is in append mode.

wrongkindofdoctor: No, all this does is initialize it to .false. 
fileobj%is_root is set on line 445 after getting the pelist and querying whether mpp_pe() matches the io_root.

PR #280, opened 2020-02-04
User/jml/update fms2io tests

wrongkindofdoctor:
**Description**
This PR contains updates to the fms2_io test suite in test_fms. 

**How Has This Been Tested?**

**Checklist:**
- [x] `make distcheck` passes


colingladueNOAA:
Per FMS meeting discussion on 2/13/2020, this is being pushed off to the next release, 2020.02.

test_fms/fms2_io/test_fms2_io.sh
```diff
@@ -1,3 +1,12 @@
 #!/bin/sh
-export PATH="$PATH:../bats/bin"
-bats test_fms2_io.bats
+
+# This is part of the GFDL FMS package. This is a shell script to
+# execute tests in the test_fms/field_manager directory.
```
edwardhartnett: This comment should be updated to reflect the actual directory.

test_fms/fms2_io/test_fms2_io.sh
```diff
@@ -1,3 +1,12 @@
...
+
+# Ed Hartnett 11/29/19
```
edwardhartnett: My name should be removed, and the author's substituted, since I did not create this file. ;-)

edwardhartnett:
This PR seems to correctly remove BATS from the test_fms/fms2_io directory, and thus fixes #358.

I hope this PR can be merged soon! ;-)

test_fms/fms2_io/test_fms2_io.sh
```diff
@@ -12,5 +12,7 @@
 rm -f *.nc > /dev/null 2>&1
 rm -f *.nc.* > /dev/null 2>&1
 rm -f logfile.*.out > /dev/null 2>&1
+rm -f input.nml > /dev/null 2>&1
+cp input_base.nml input.nml
```
edwardhartnett: It's actually better to remove these files in CLEANFILES in Makefile.am. That way, after the test is run, the files are still there for debugging purposes. Only when the user wants to clean them do they get erased.

wrongkindofdoctor:
I built and ran the test suite on the skylake dev box, and the log shows the following error:
```
Using            6  MPI ranks.
NOTE from PE     0: MPP_DOMAINS_SET_STACK_SIZE: stack size set to    32768.
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
  X-AXIS =   96
  Y-AXIS =   96
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
  X-AXIS =   96
  Y-AXIS =   96
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
  X-AXIS =   96
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
  X-AXIS =   96
  Y-AXIS =   96
  X-AXIS =   96
  Y-AXIS =   96
  Y-AXIS =   96
 Cubed-sphere domain decomposition
whalo =    2, ehalo =    2, shalo =    2, nhalo =    2
  X-AXIS =   96
  Y-AXIS =   96

FATAL from PE     1: NetCDF: Not a valid data type or _FillValue type mismatch

application called MPI_Abort(MPI_COMM_WORLD, 1) - process 1

FATAL from PE     5: NetCDF: Not a valid data type or _FillValue type mismatch

application called MPI_Abort(MPI_COMM_WORLD, 1) - process 5

FATAL from PE     0: NetCDF: Not a valid data type or _FillValue type mismatch
```
I'll add that I can't figure out how to get a stack trace written to the log, even after adding `-O0 -g` to the CFLAGS and FCFLAGS environment variables  before running `configure`

thomas-robinson:
@wrongkindofdoctor you may need to add `-traceback` also

wrongkindofdoctor:
-traceback does not seem to be an option. I added -fbacktrace to FFLAGS, but this still didn't work.

thomas-robinson:
@wrongkindofdoctor -traceback is an intel compiler option.  Have you tried it with intel?

edwardhartnett:
This problem is may be occurring because the Unidata netcdf team recently added a check for the type of the fill value.

The fill value type must match the data type it applies to. So if the variable is NC_DOUBLE, then the _FillValue for that variable must also be NC_DOUBLE.

This was always a documented rule, but it was not enforced by the library. Recently they added enforcement, and this has caused some issues such as this in the user community.

So find your nf90_put_att() calls which writes the fill values, and make sure the types match the variables they apply to.

marshallward:
You likely won't see a traceback unless the program creates or gets a signal to abort.  Since (I think) it's going through `MOM_error` after detecting some invalid inputs, it's just reporting the error and exiting as if nothing had gone wrong.

There are some platform-specific functions to force a traceback - [this is what we did in MOM5 for example](https://github.com/mom-ocean/MOM5/blob/75e267c03b68b1c862eb28831808e6fdc932ff20/src/shared/mpp/include/mpp_util_mpi.inc#L56-L59) - but usually I just set a break point in GDB (or whatever debugger) and look at it in there.

marshallward:
Slight amendment to my previous post: I see now that `MOM_error` is going through `mpp_error` and finally `MPI_Abort`, which is supposed to send a nonzero error code (looks hard-set to 1 in most cases) to the launcher process.

I would still guess that something is not sending a nonzero to the launcher, but cannot see what it could be.

I've noticed this myself quite a bit lately, and might be worth understanding a bit better, but might be better handled in a separate issue.

bensonr:
Besides the 'use ifcore / call tracebackqq' for intel, gfortran supports
'call backtrace'.

Directly from an Intel support forum:

A traceback is provided only when an exception is raised or the Fortran
run-time system detects an error condition.

test_fms/fms2_io/Makefile.am
```diff
@@ -1,33 +1,40 @@
...
 
 # This is the source code for the test.
-test_atmosphere_io_SOURCES = argparse.F90 setup.F90 test_atmosphere_io.F90
-test_atmosphere_io_LDADD = ${top_builddir}/libFMS/libFMS.la
-EXTRA_DIST = test_atmosphere_io.sh test_atmosphere_io.bats input.nml
+test_fms2_io_SOURCES = argparse.F90 test_fms2_io.F90 create_atmosphere_domain.inc \
+                       create_land_domain.inc create_ocean_domain.inc \
+                       ocean_restart_file_test.inc land_compressed_restart_file_test.inc \
+                       land_unstructured_restart_file_test.inc
+test_atmosphere_io_SOURCES = argparse.F90 setup.F90 test_atmosphere_io.F90 \
```
edwardhartnett: This line (20) has an extra space at the end of the line, which causes autoreconf -i to issue this warning:

`test_fms/fms2_io/Makefile.am:20: warning: whitespace following trailing backslash`

wrongkindofdoctor:
I'm not sure why this is still failing. Make check passes externally, I have both tests set to "skip",  but travis-ci is not registering the skip command in test_atmosphere_io.sh.

test_fms/fms2_io/test_atmosphere_io.sh
```diff
@@ -1,2 +1,12 @@
 #!/bin/sh
-bats ${srcdir}/test_fms/fms2_io/test_atmosphere_io.bats
+# This is part of the GFDL FMS package. This is a shell script to
+# execute tests in the test_fms/fms2_io directory.
+
+# Set common test settings.
+. ../test_common.sh
+
+# link to the input namelist
+cp ${srcdir}/test_fms/fms2_io/input_base.nml input.nml
```
uramirez8707: @wrongkindofdoctor In travis `make distcheck` fails because `cp: cannot create regular file 'input.nml': File exists` 

The test passes if you remove this line.

edwardhartnett:
OK, perhaps instead you want to remove input.nml from git. Then the copy
will work as intended...

wrongkindofdoctor:
Okay, this PR is FINALLY ready for review. Thanks, everyone!

test_fms/fms2_io/input.nml
```diff
@@ -1 +1 @@
-aljds;falkjsd
+blah;blahblah
```
colingladueNOAA: Is this meant to be here?

wrongkindofdoctor: Yes, unless others are opposed to having a bit of fun with a dummy file.

colingladueNOAA: I approve

wrongkindofdoctor:
> This is good for now, but at some point we have to fix/improve the broken tests ...

Yes. This fix is a stopgap to allow the test suite to run while I work on solving the issue(s) with the netcdf data type mismatches.

test_fms/fms2_io/input.nml
bensonr: I understand the need for input.nml to exist because mpp_init reads it into the input_nml_file character array for use with namelist internal file reads (plus documenting it via the logfile).  But I think the cleanest solution is to have the test script "cat" a dummy namelist entry into the input.nml file:
cat input.nml < EOF
  &dummy /
EOF

edwardhartnett:
Can this be merged soon please?

Currently your master branch is broken. This is something you should not tolerate.

edwardhartnett:
Thanks!!!

PR #286, opened 2020-02-06
potential fix for some axis_utils2 issues

menzel-gfdl:
**Description**
Fixes #277

**How Has This Been Tested?**
Tested in a 2-day ocean-ice regression run on Gaea C3 with Intel 16.

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #284, opened 2020-02-05
Allow calendar "365_DAYS" in addition to "365_DAY"

nikizadehgfdl:
- The calendar in NCAR files are called "365_DAYS". E.g.,
             double TIME(TIME) ;
                    TIME:axis = "T" ;
                    TIME:calendar_type = "NO_LEAP" ;
                    TIME:modulo = " " ;
                    TIME:time_origin = "31-DEC-1899 00:00:00" ;
                    TIME:units = "days since 1900-12-31 18:00:00" ;
                    TIME:calendar = "365_DAYS" ;

      I think supporting these files was the reason  to put "365_DAY" (note the missing "S") in
      time_manager/get_cal_time.F90
      However, I don't know how it works in  fms1.
      There must be a magic or a bug in original time_interp_external.F90 call to
         call mpp_get_atts(time_axis,units=units,calendar=calendar_type)
      that drops the trailing "S" in the calendar_type
      and matches it with the option "365_DAY" rather than "365_DAYS"!

- Without this fix the CORE forcing models crash like:

    FATAL from PE    36: get_cal_time: "365_days" is not an acceptable calendar attribute. acceptable calendars are:  noleap, 365_day, 360_day, julian, no_calendar, thirty_da
y_months, gregorian

    fms_MOM6_SIS2_com  0000000001922F90  mpp_mod_mp_mpp_er          69  mpp_util_mpi.inc
    fms_MOM6_SIS2_com  0000000001C298EA  get_cal_time_mod_         219  get_cal_time.F90
    fms_MOM6_SIS2_com  0000000001C48BC5  time_interp_exter         603  time_interp_external2.F90
    fms_MOM6_SIS2_com  0000000001B8252A  data_override_mod         826  data_override.F90
    fms_MOM6_SIS2_com  0000000001B88ABE  data_override_mod         673  data_override.F90
    fms_MOM6_SIS2_com  00000000006783BB  ice_spec_mod_mp_g         105  ice_spec.F90
    fms_MOM6_SIS2_com  00000000004CADBF  ice_model_mod_mp_        2510  ice_model.F90
    fms_MOM6_SIS2_com  000000000041447E  coupler_main_IP_c        1762  coupler_main.F90
    fms_MOM6_SIS2_com  000000000040A717  MAIN__                    612  coupler_main.F90

PR #289, opened 2020-02-07
Writes regional diagnostics to hostiry files.  Resolves #281.

### LINKED ISSUE #281: Regional diag output needs update for non-domain decomposed write
**Describe the bug**
Regional diagnostic output does not work because it is it being opened as a domain decomposed file.  In the case of regional output, every core should be writing its own file.

**To Reproduce**
Run with regional diagnostic output

**Expected behavior**
The files should be written on a per-core basis and not using the io_domain

**System Environment**
This is all environments.

**Additional context**

thomas-robinson:
**Description**
The regional diagnostics have a 2d domain, but the IO is done on a per core basis and not on an IO domain.  This PR checks if the 2D domain has an IO domain.  If it does, then the file is opened as a domain decomposed file.  If there is no IO domain (`io_domain` is not associated), then the file is opened by the code without using the domain.  The X and Y axes are not handles in any special way.

Fixes #281 

**How Has This Been Tested?**
The details of the test are in #281.  I added the suggest diagnostics to the diag table as @nikizadehgfdl suggested.  

**Checklist:**
- [x] `make distcheck` passes


thomas-robinson:
This should be pulled after #283 #284 #285 #286

PR #292, opened 2020-02-07
Allows region diagnostic history files to be written 

### LINKED ISSUE #281: Regional diag output needs update for non-domain decomposed write

thomas-robinson:
**Description**
The regional diagnostics have a 2d domain, but the IO is done on a per core basis and not on an IO domain.  This PR checks if the 2D domain has an IO domain.  If it does, then the file is opened as a domain decomposed file.  If there is no IO domain (`io_domain` is not associated), then the file is opened by the code without using the domain.  The X and Y axes are not handles in any special way.

Fixes #281 
Replaces #289 

**How Has This Been Tested?**
The details of the test are in #281.  I added the suggest diagnostics to the diag table as @nikizadehgfdl suggested.  

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -57,7 +57,7 @@ use,intrinsic :: iso_c_binding, only: c_double,c_float,c_int64_t, &
   USE mpp_domains_mod, ONLY: domain1d, domain2d, mpp_define_domains, mpp_get_pelist,&
        &  mpp_get_global_domain, mpp_get_compute_domains, null_domain1d, null_domain2d,&
        & domainUG, null_domainUG, CENTER, EAST, NORTH, mpp_get_compute_domain,&
-       & OPERATOR(.NE.), mpp_get_layout, OPERATOR(.EQ.)
+       & OPERATOR(.NE.), mpp_get_layout, OPERATOR(.EQ.), mpp_get_io_domain
   USE mpp_mod, ONLY: mpp_npes, mpp_pe, mpp_root_pe, mpp_get_current_pelist
   USE diag_axis_mod, ONLY: diag_axis_init, get_diag_axis, get_axis_length,&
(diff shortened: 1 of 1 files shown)


thomas-robinson:
@nikizadehgfdl  and @menzel-gfdl please test this one instead of #289.  #289 has updates that are not being merged in.

thomas-robinson:
@underwoo I retested to confirm that this is working.

diag_manager/diag_output.F90
```diff
@@ -214,11 +216,37 @@ SUBROUTINE diag_output_init(file_name, FORMAT, file_title, file_unit,&
...
+          case (10 : 99)
+               mype_string = "00"//trim(mype_string)
+          case (100 : 999)
+               mype_string = "0"//trim(mype_string)
+          case (1000 : 9999)
+               mype_string = trim(mype_string)
+          case default
+               CALL error_mesg('diag_output_init', "The PE number is above 9999, so you can not add 4 digits to the "//&
+                    "end of the file name "//trim(file_name) ,FATAL)
+        end select
```
underwoo: This can be accomplished using a formatted `write` statement, similar to:

```Fortran
write (mype_string, '(I0.4)') mype
```

menzel-gfdl: Also, is there a check to make sure that the run is not using > 10,000 ranks?

thomas-robinson: I increased the string length to 9, so it will be able to work if the PE number is greater than 9999.  It will write a minimum of 4 digits to `mype_string`.  @underwoo  and I tested this with gfortran and ifort.

PR #293, opened 2020-02-07
More get valid fix

uramirez8707:
**Description**
This modifies logic in function get_valid of fms2_io/netcdf_io.F90 that broadcast valid to the other ranks

1. initializes valid%has_fill, valid%has_missing, valid%has_range as false for all ranks
2. root ranks goes in the file and fills  valid%has_fill, valid%has_missing, valid%has_range if it is in the file 
3. if each valid%has_fill, valid%has_missing, valid%has_range is true, it broadcast it to the other pes

**How Has This Been Tested?**
GAEA on c4 with intel17 compiler on prod,repro,debug with and without openmp and all experiments from awg_xml. 

It **FINALLY** reproduces xanadu!

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/fms2_io/netcdf_io.F90
+++ b/fms2_io/netcdf_io.F90
@@ -1578,7 +1578,7 @@ function get_valid(fileobj, variable_name) &
       endif
     endif
     valid%has_range = has_min .and. has_max
-  endif
+  endif !if (fileobj%is_root)
 
   if (valid%has_range) then !If it found has_range broadcast to the rest of the pes
     call mpp_broadcast(valid%has_range, fileobj%io_root, pelist=fileobj%pelist)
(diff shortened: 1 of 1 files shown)


uramirez8707:
@wrongkindofdoctor  Please have @menzel-gfdl take a look before merging

fms2_io/netcdf_io.F90
```diff
@@ -1576,18 +1578,20 @@ function get_valid(fileobj, variable_name) &
       endif
     endif
     valid%has_range = has_min .and. has_max
-  endif
+  endif !if (fileobj%is_root)
 
-  call mpp_broadcast(valid%has_range, fileobj%io_root, pelist=fileobj%pelist)
-  if (valid%has_range) then
+  if (valid%has_range) then !If it found has_range broadcast to the rest of the pes
```
menzel-gfdl: @wrongkindofdoctor @uramirez8707 
I don't think this will work.  The broadcasts of valid%has_range, valid%has_fill, and valid%has_missing must
happen before the if statements or I think the code can hang (when the root has the value of .true. for one of these flags).

PR #294, opened 2020-02-07
Revert "More get valid fix"

wrongkindofdoctor:
Reverts NOAA-GFDL/FMS#293

After @menzel-gfdl voicing his concerns about hanging, I'm reverting this PR as a precaution.


thomas-robinson:
@wrongkindofdoctor this is fine.  Should we put it through testing first?  Can we confirm with @menzel-gfdl that this will hang?

PR #320, opened 2020-02-24
Fix: get_valid and is_valid

### LINKED ISSUE #321: Data Override Does not Reproduce Answers when fill value and missing value are present and equal.
**Describe the bug**
Valid values are not being read properly by time_interp_external2 when both a missing value and fill value are present in the variable and they are of equal value. This seems to be a rare occurrence in a variable for data_override, but it does exist. This is resulting in answer changes from Xanadu.

**To Reproduce**
Run AM4.1 with e1c0d9d01d844938adc0d18afa09532f336bcdfe or earlier.

**Expected behavior**
The valid values contained within the Netcdf file to be read in as valid values.

**System Environment**
Gaea C4 Intel16 - Prod and Debug
- Compiler(s): Intel16

**Additional context**
The NUG convention stipulates that a minimum or maximum value should be defined if a fill value is defined. The implementation of this in the `get_valid` call in netcdf_io.F90 seems to be the culprit.

GFDL-Eric:
Fixes to get_valid and is_valid to assign minimum and maximum valid values without creating impossible situations. @menzel-gfdl @uramirez8707 

**Description**
Prior to this fix, the get_valid function would auto-assign a minimum or maximum value according to the fill value or missing value in the file, which is consistent with the NUG convention. However, the implementation was incorrect, and resulted in situations in which it was impossible for any values to be valid. This situation occurred because the specific netcdf file referenced in data_override contained both a "missing value" and a "fill value" attribute for the variable, and those attributes contained the same value. The fix moves has_min and has_max attributes inside the valid type and now prohibits missing value from influencing the minimum or maximum value, in accordance with the NUG convention. Furthermore, either a minimum or maximum value will make the range valid (aka an open range is allowed and tested within `is_valid`, and there are also validity tests for fill value and missing values.

Fixes #321 

**How Has This Been Tested?**
Tests have used the data_override unit tests (in test_fms). The fix was tested with the unit test, and should accommodate scenarios for every combination of "min_val, max_val, valid_range, _FillValue, and missing_value" contained in a given netcdf file.

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/fms2_io/netcdf_io.F90
+++ b/fms2_io/netcdf_io.F90
@@ -81,6 +81,8 @@ endtype FmsNetcdfFile_t
 !> @brief Range type for a netcdf variable.
 type, public :: Valid_t
   logical :: has_range !< Flag that's true if both min/max exist for a variable.
+  logical :: has_min !< Flag that's true if both min exists for a variable.
+  logical :: has_max !< Flag that's true if both max exists for a variable.
   logical :: has_fill !< Flag that's true a user defined fill value.
   logical :: has_missing !< Flag that's true a user defined missing value.
   real(kind=real64) :: fill_val !< Unpacked fill value for a variable.
@@ -1478,14 +1480,12 @@ function get_valid(fileobj, variable_name) &
   real(kind=real64) :: scale_factor
   real(kind=real64) :: add_offset
   real(kind=real64), dimension(2) :: buffer
-  logical :: has_max
-  logical :: has_min
   integer :: xtype
 
   if (fileobj%is_root) then
     varid = get_variable_id(fileobj%ncid, variable_name)
-    has_max = .false.
-    has_min = .false.
+    valid%has_max = .false.
+    valid%has_min = .false.
     valid%has_fill = .false.
     valid%has_missing = .false.
     valid%has_range = .false.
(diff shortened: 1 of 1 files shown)


uramirez8707:
@nikizadehgfdl Can you test this? With this am4 **should** reproduce

thomas-robinson:
@GFDL-Eric You did not describe how this was tested.  Did you run @nikizadehgfdl experiment?

fms2_io/netcdf_io.F90
```diff
@@ -1503,86 +1503,80 @@ function get_valid(fileobj, variable_name) &
...
+        valid%max_val = buffer(1)*scale_factor + add_offset
+        valid%has_max = .true.
+      endif
+      if (attribute_exists(fileobj%ncid, varid, "valid_min")) then
+        call get_variable_attribute(fileobj, variable_name, "valid_min", buffer(1))
+        valid%min_val = buffer(1)*scale_factor + add_offset
+        valid%has_min = .true.
+      endif
+    endif
+    valid%has_range = valid%has_min .or. valid%has_max
```
thomas-robinson: If there is a range, only one of these has to be defined?  It seems like this should be an `.and.`.  Shouldn't valid%has_range be set to .true. in the if block under 1508?

fms2_io/netcdf_io.F90
```diff
@@ -1503,86 +1503,80 @@ function get_valid(fileobj, variable_name) &
...
+
```
thomas-robinson: I think a comment is missing here describing what is happening.  There was a comment here before, but now it's gone/partially moved.

fms2_io/netcdf_io.F90
```diff
@@ -1621,11 +1615,22 @@ elemental function is_valid(datum, validobj) &
 
   valid_data = .true.
   if (validobj%has_range) then
-    valid_data = rdatum .ge. validobj%min_val .and. rdatum .le. validobj%max_val
-  elseif (validobj%has_fill) then
-    valid_data = rdatum .ne. validobj%fill_val
-  elseif (validobj%has_missing) then 
-    valid_data = rdatum .ne. validobj%missing_val
+    if (validobj%has_min .and. .not. validobj%has_max) then
```
thomas-robinson: Is this possible with the new code?  If there is a range, don't you set these flags both to `.true.`

GFDL-Eric:
@thomas-robinson I've been running my tests using the data_override unit tests (in test_fms). Previous tests (with master) revealed that no valid values were bring read in by time_interp_external2, and therefore answers were not reproducing, so Uriel, Ray, and I had a brainstorming session on Friday afternoon to diagnose the issue and chart the path forward. The issue revolved around the way get_valid auto-assigns minimum or maximum values according to the fill value or missing value, which is something that should occur according to the NUG? convention (is that right @menzel-gfdl?). The implementation in the new I/O was resulting in impossible situations, but only for the specific netcdf file being tested because it contained both a "missing value" and a "fill value" attribute for the variable. I've tested the fix with the unit test and it works, and should accommodate scenarios for every combination of "min_val, max_val, valid_range, _FillValue, and missing_value" contained in a given netcdf file. I haven't tested with Niki's AM4.1 setup... I can try to get it set up correctly but it may just be easier for Niki to test it with this branch?

thomas-robinson:
You should add some of this to the description, because you didn't fill in the **How has this been tested?** part of the template.
> @thomas-robinson I've been running my tests using the data_override unit tests (in test_fms). Previous tests (with master) revealed that no valid values were bring read in by time_interp_external2, and therefore answers were not reproducing, so Uriel, Ray, and I had a brainstorming session on Friday afternoon to diagnose the issue and chart the path forward. The issue revolved around the way get_valid auto-assigns minimum or maximum values according to the fill value or missing value, which is something that should occur according to the NUG? convention (is that right @menzel-gfdl?). The implementation in the new I/O was resulting in impossible situations, but only for the specific netcdf file being tested because it contained both a "missing value" and a "fill value" attribute for the variable. I've tested the fix with the unit test and it works, and should accommodate scenarios for every combination of "min_val, max_val, valid_range, _FillValue, and missing_value" contained in a given netcdf file. I haven't tested with Niki's AM4.1 setup... I can try to get it set up correctly but it may just be easier for Niki to test it with this branch?

fms2_io/netcdf_io.F90
GFDL-Eric: I suppose this could be a topic of discussion: does a range exist if it is "open" (aka "undefined") on one end? My interpretation was that an open range should exist (and be tested). The old is_valid test only tested for a closed range, but if we're defining a minimum value, shouldn't we test to see if the values are above that minimum, regardless of whether a maximum is defined? The choice of setting `valid%has_range` to `.true.` when has_min is defined but has_max isn't (or v.v.) is also open to discussion, in other words, should `valid%has_range` be a strict "closed definition" or a more "open definition"? I tend to prefer the latter. Under that definition, I wanted to wait until those scenarios were tested (1515 and 1520 blocks) before assigning `valid%has_range`.

fms2_io/netcdf_io.F90
GFDL-Eric: As per my comment from the get_valid call, in the open range definition of `valid%has_range` this is a possibility (and potentially a frequent one with _FillValue defining a min or max)... obviously if we want to set `valid%has_range` to the closed definition I'd have to rework this section (and the others).

fms2_io/netcdf_io.F90
```diff
@@ -1503,86 +1503,80 @@ function get_valid(fileobj, variable_name) &
...
+      endif
+    endif
+    valid%has_range = valid%has_min .or. valid%has_max
+
 
-    !Get default max/min from missing_value.  These could be overwritten by
-    !vaild_range/valid_min/valid_max/_Fill_Value.
     if (attribute_exists(fileobj%ncid, varid, "missing_value")) then
       call get_variable_attribute(fileobj, variable_name, "missing_value", buffer(1))
       xtype = get_variable_type(fileobj%ncid, varid)
```
menzel-gfdl: @GFDL-Eric I think we can delete this xtype = line now.

GFDL-Eric: Resolved with 10096f4

thomas-robinson:
@GFDL-Eric Please open an issue that describes the problem.

fms2_io/netcdf_io.F90
GFDL-Eric: Added several new comment lines describing the behavior of sections in get_valid

GFDL-Eric:
Issue #321 opened

nikizadehgfdl:
Nice work everybody! This update brings back  xanadu answers for AM4.1.   Thanks.

I should figure out why ESM4 still changes answers.

PR #342, opened 2020-03-18
Modify get_global_io_domain_indices to output buffer of axis data

### LINKED ISSUE #341: Modify get_global_io_domain_indices to output buffer of axis data
When writing x/y axis data, I have to: 

1.  call `get_global_io_domain_indices` to get the starting and ending indices for each dimension
2. allocate a buffer the size of the difference between the ending and starting dimensions
3. add the data to the buffer 

and I have to do this for every axis, for every restart, for every model component which is leads to a lot of unnecessary copy and pasted code ...

It will be **nice**, if `get_global_io_domain_indices` created the buffer already for you and it was one of the subroutines arguments, similar to how to `get_compute_domain_dimension_indices`  already works: 
```F90
allocate(indices(e-s+1))
  do i = s, e
    indices(i-s+1) = i
  enddo
```

N/A

**Additional context**
N/A

uramirez8707:
**Description**
This adds `indices` as optional argument to `get_global_io_domain_indices`
If present, the subroutine allocates `indices` to the size of the difference between the ending and starting dimensions and it add the data. 

Fixes #341 

This also removes tabs in fms2_io/fms_netcdf_domain_io.F90 and fms2_io/netcdf_io.F90
**How Has This Been Tested?**
GAEA on intel with AM4.0 and AM4.1

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/fms2_io/fms_netcdf_domain_io.F90
+++ b/fms2_io/fms_netcdf_domain_io.F90
@@ -370,8 +370,8 @@ function open_domain_file(fileobj, path, mode, domain, nc_format, is_restart) &
     else
       success = netcdf_file_open(fileobj, combined_filepath, mode, nc_format, pelist, &
                                  is_restart)
-		!If the file is combined and the layout is not (1,1) set the adjust_indices flag to false
-		if (success .and. (io_layout(1)*io_layout(2) .gt. 1)) fileobj%adjust_indices = .false.
+!If the file is combined and the layout is not (1,1) set the adjust_indices flag to false
+      if (success .and. (io_layout(1)*io_layout(2) .gt. 1)) fileobj%adjust_indices = .false.
     endif
   endif
   if (.not. success) then
@@ -714,22 +714,22 @@ subroutine domain_offsets(data_xsize, data_ysize, domain, xpos, ypos, &
                               position=xpos)
   ! If the xpos is east and the ending x index is NOT equal to max allowed, set extra_x_point to true
   if (present(extra_x_point)) then
-	 if ((xpos .eq. east) .and. (iec .ne. xmax)) then 
-		extra_x_point = .true.
-    	else
-		extra_x_point = .false.
-	endif
+     if ((xpos .eq. east) .and. (iec .ne. xmax)) then 
+        extra_x_point = .true.
+     else
+        extra_x_point = .false.
+     endif
   endif
 
   call mpp_get_compute_domain(io_domain, ybegin=jsc, yend=jec, ysize=yc_size, &
                               position=ypos)
   ! If the ypost is north and the ending y index is NOT equal to max allowed, set extra_y_point to true
   if (present(extra_y_point)) then
-	 if ((ypos .eq. north) .and. (jec .ne. ymax)) then 
-		extra_y_point = .true.
-	 else
-		extra_y_point = .false.
-   	 endif
+      if ((ypos .eq. north) .and. (jec .ne. ymax)) then 
+         extra_y_point = .true.
+      else
+         extra_y_point = .false.
+      endif
   endif
 
   buffer_includes_halos = (data_xsize .eq. xd_size) .and. (data_ysize .eq. yd_size)
--- a/fms2_io/fms_netcdf_domain_io.F90
+++ b/fms2_io/fms_netcdf_domain_io.F90
@@ -749,10 +749,11 @@ subroutine get_global_io_domain_indices(fileobj, dimname, is, ie, indices)
   character(len=*), intent(in) :: dimname !< Name of dimension variable.
   integer, intent(out) :: is !< Staring index of I/O global domain.
   integer, intent(out) :: ie !< Ending index of I/O global domain.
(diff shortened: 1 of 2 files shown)


fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -743,15 +743,16 @@ end subroutine domain_offsets
...
 !> @brief Get starting/ending global indices of the I/O domain for a domain decomposed
 !!        file.
-subroutine get_global_io_domain_indices(fileobj, dimname, is, ie)
+subroutine get_global_io_domain_indices(fileobj, dimname, is, ie, indices)
 
   type(FmsNetcdfDomainFile_t), intent(in) :: fileobj !< File object.
   character(len=*), intent(in) :: dimname !< Name of dimension variable.
   integer, intent(out) :: is !< Staring index of I/O global domain.
   integer, intent(out) :: ie !< Ending index of I/O global domain.
+  integer, dimension(:), allocatable, intent(inout), optional :: indices
```
wrongkindofdoctor: Add a doxygen-style description of the indices argument

thomas-robinson: This implies that indices always has to be on the heap.  Are you sure that's what you want?  Also, I want to agree with Jess about doxygen commenting on what this variable is.

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -743,15 +743,16 @@ end subroutine domain_offsets
...
 
   type(domain2d), pointer :: io_domain
-  integer :: dpos
+  integer :: dpos, i
```
thomas-robinson: Sylistically speaking, `i` should be on its own line

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -767,6 +768,18 @@ subroutine get_global_io_domain_indices(fileobj, dimname, is, ie)
...
+
+  if (present(indices)) then 
+    if(allocated(indices)) then
+      deallocate(indices)
+    endif
+    allocate(indices(ie-is+1))
+    do i = is, ie
+      indices(i-is+1) = i
+    enddo
+  endif
```
thomas-robinson: There should be a comment explaining what's happening here.

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -767,6 +768,18 @@ subroutine get_global_io_domain_indices(fileobj, dimname, is, ie)
       call error("input dimension is not associated with the domain.")
     endif
   endif
+
+  if (present(indices)) then 
+    if(allocated(indices)) then
+      deallocate(indices)
+    endif
+    allocate(indices(ie-is+1))
```
thomas-robinson: It seems odd that you are reallocating a variable that's sent in.  This seems like some behind the scenes magic sauce that could cause danger.

fms2_io/fms_netcdf_domain_io.F90
uramirez8707: > You really need to add the doxy comment for the `indices` variable.
> 
> I don't think deallocating and reallocating a variable sent in and changing the size of the dimension is a good idea. It should come in with the correct size IMO.

The code doesn't know the size of `indices` before  `get_global_io_domain_indices` is called. The size of `indices` is defined by `is`, `ie`, which are obtained after the ` mpp_get_global_domain` call

I could (should?) change it from `inout` to just `out`

thomas-robinson: > I could (should?) change it from `inout` to just `out`
I think that makes a lot more sense.  It needs to be vary clear that this array **WILL** be changed.

fms2_io/fms_netcdf_domain_io.F90
thomas-robinson: @uramirez8707 If you change `indices` to `out`, then then if it's already allocated, I think you should call a fatal instead of just plowing over whatever was already sent in.  The caller should be handing you an unallocated array.

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -370,8 +370,8 @@ function open_domain_file(fileobj, path, mode, domain, nc_format, is_restart) &
     else
       success = netcdf_file_open(fileobj, combined_filepath, mode, nc_format, pelist, &
                                  is_restart)
-		!If the file is combined and the layout is not (1,1) set the adjust_indices flag to false
-		if (success .and. (io_layout(1)*io_layout(2) .gt. 1)) fileobj%adjust_indices = .false.
+!If the file is combined and the layout is not (1,1) set the adjust_indices flag to false
+      if (success .and. (io_layout(1)*io_layout(2) .gt. 1)) fileobj%adjust_indices = .false.
```
thomas-robinson: Is this still a tab?

uramirez8707: no it's all spaces. same with bellow

PR #348, opened 2020-03-20
change compute_global_checksum.inc to support `kind=real32`

### LINKED ISSUE #347: fsm2io: compute_global_checksum.inc doesn't support `kind=real32`
**Describe the bug**
@laurenchilutti  reported the following error when implementing the new io into GFDL_atmos_cube_sphere master: 

`
FATAL from PE     7: unsupported type
`

The fatal happens in `compute_global_checksum.inc`. 
The variable that it is trying to calculate the checksum for is a `kind=real32`. 
The part of the code that handles this is type is commented out: https://github.com/NOAA-GFDL/FMS/blob/8f3d7ab3184495fc0bfac7fc1e9a1a951c98c144/fms2_io/include/compute_global_checksum.inc#L90-L98

Uncommenting this lines, gets us pass this issue ...

**To Reproduce**
Run the GFDL_atmos_cube_sphere test with [atmos_cube_sphere with new io](https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/tree/fms2_io_update), [FMS with nesting changes ](https://github.com/bensonr/FMS/tree/zhi_nest_merge)+ #342 

**Expected behavior**
The code should be able to calculate checksums for floats. 

**System Environment**
This happens in every system. 

**Additional context**
N/A

uramirez8707:
**Description**
This comments out the code in compute_global_checksum that handles the `kind=real32`

Fixes #347 

**How Has This Been Tested?**
Ran the GFDL_atmos_cube_sphere test with [atmos_cube_sphere with new io](https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/tree/fms2_io_update), [FMS with nesting changes ](https://github.com/bensonr/FMS/tree/zhi_nest_merge)+ #342 + these changes in GAEA, debug mode, with  intel/16.0.3.210

**Checklist:**
- [x] `make distcheck` passes


uramirez8707:
@laurenchilutti please test this with the nesting/normal case in prod

fms2_io/include/compute_global_checksum.inc
```diff
@@ -87,7 +87,8 @@ function compute_global_checksum_2d(fileobj, variable_name, variable_data, is_de
         chksum_val = mpp_chksum(buf_int64)
       endif
       deallocate(buf_int64)
-   type is (real(kind=real32))
+   type is (real(kind=real32)
+#ifdef OVERLOAD_R4)
```
thomas-robinson: 1. Why are there `#ifdef`s?
2. It looks like there is a ) that doesn't belong

menzel-gfdl: `mpp_chksum` can only handle real(kind=real32) variables if you compile with `-DOVERLOAD_R4`.

thomas-robinson: @menzel-gfdl Then should there is a fatal called if it's not defined?

menzel-gfdl: Yes there is one:
```
  class default
    call error("unsupported type.")
end select
```

The `type is (real(kind=real32))` is inside the `#ifdef`.

thomas-robinson: OK, it is now.  In this example, it wasn't.  Maybe there should be a comment that says that if `OVERLOAD_R4` isn't, then you will end up here.  

If you have an else, you could give an error that is more descriptive than "unsupported type".  It could say that "you are trying to use a real*4 without defining OVERLOAD_R4"

uramirez8707: so like?
```F90
type is (real(kind=real32))
  #ifdef OVERLOAD_R4
      ... 
  #else
   call error("you are trying to use a real*4 without defining OVERLOAD_R4")
  #endif 
type is (real(kind=real64))
```

thomas-robinson: @uramirez8707 yes that is more descriptive than "unsupported type."  Telling the user to compile with -DOVERLOAD_R4 would be better too because it gives the user a solution to their problem.

uramirez8707:
I think this is ready.

PR #369, opened 2020-04-02
Changing "OBJEXT" to "lo" in Makefile.am files for more correct makefiles

### LINKED ISSUE #368: Makefile.am use of $(OBJEXT) in dependencies to libtool libraries should be .lo instead
Currently in our Makefile.am files we have stuff like this:

```
# Each mod file depends on the .o file.
fms2_io_mod.mod: fms2_io.$(OBJEXT)
fms_io_utils_mod.mod: fms_io_utils.$(OBJEXT)
netcdf_io_mod.mod: netcdf_io.$(OBJEXT)

```
However, I have recently learned from work on netcdf-fortran that this is not quite correct and may cause problems on parallel builds (by parallel I mean make -j, not parallel I/O).

Instead, we should be using .lo

This is well explained by netCDF user @skosukhin in this netcdf-fortran PR: https://github.com/Unidata/netcdf-fortran/pull/238

As Sergey says there:

> If an object file is part of a libtool library, the dependencies on Fortran module files should mention .lo files and not .$(OBJEXT) files because .$(OBJEXT) files are not part of the dependency graph of the library according to make. This is, however, not the case for objects that get linked into an executable directly, which is why new dependencies in nf_test mention .$(OBJEXT) files.

I will not attempt to do all of this in one PR, because that will touch every Makefile.am at the same time. Instead I'll do a few directories at a time and the change can be more gradual.

edwardhartnett:
Fixes #368 

In this issue I said I would do this in batches, but instead I just did it all at once. It's like eating potato chips - it's just hard to stop once you get started! ;-)


underwoo:
@colingladueNOAA and @thomas-robinson please run this PR on the skylake box using `make -j`.  I would like to know if this does resolve the build issues we have seen on systems with large number of processors.  One issue this PR still fails to resolve are instructions to Make on how to create the Fortran module files (.mod is a built-in suffix).  I am working on something that should fully resolve the parallel build issue we have seen, and allow Make to create the Fortran module files.  It is taking some time, mainly b/c of ensuring it will work with all the compilers.

colingladueNOAA:
I will test this on Skylake.

thomas-robinson:
@underwoo Do you want to just with -j with different numbers, or just -j?

underwoo:
Using `-j` should use all available processors to do the build.
--
Seth Underwood

S/W Development Process Lead
Modeling Systems Group
GFDL/NOAA/DOC
201 Forrestal Road
Princeton, NJ 08540-6649

(609) 452-5847 Office
(304) 376-9002 Cell
(609) 987-5063 Fax
Seth.Underwood@noaa.gov

colingladueNOAA:
On skylake using Ed's branch:

module load intel impi netcdf
autoreconf -i
./configure FCFLAGS="-fno-alias -auto -safe-cray-ptr -ftz -assume byterecl -i4 -r8 -nowarn -sox -traceback -qopenmp `nf-config --fflags` -O0 " CPPFLAGS="`nc-config --cflags` -I/opt/intel/2019_up2/impi/2019.2.187/intel64/include/" FC="mpiifort" CC="mpiicc"
make -j

does build without error. Subsequent rebuilds after make distclean are also successful. 

This same procedure fails on the current master branch of FMS.

edwardhartnett:
@underwoo these changes are inspired by a bunch of suggested changes to the netcdf-fortran build. 

There is also a section in there for generating the mod files. It's next on my list to bring over to FMS, but perhaps you are already working on on. What was suggested for netcdf-fortran is:

```
# If a module file is accidentally deleted, delete the corresponding object
# and rebuild it.
$(MODULES):
	@if test -z $<; then \
	  echo "Cannot find Fortran source file declaring module '$(basename $(@F))'." >&2; \
	elif test ! -f $@; then \
	  rm -f $<; $(MAKE) $(AM_MAKEFLAGS) $<; \
	fi

# If the rule above rebuilds an object in a subprocess of make, the current
# process might not know about it. Therefore, to enforce regeneration of the
# library, we introduce the following dependency.
EXTRA_libnetcdff_la_DEPENDENCIES = $(MODULES)
```

edwardhartnett:
If we can get this merged, I can move on to the next stage of improvements for the fortran build...

underwoo:
@bensonr we are waiting for an approval from you on this PR.

underwoo:
@edwardhartnett this will be merged on Monday along with a few other PRs.  You can see the list of PRs to be merged in the [2020.02 Project](https://github.com/NOAA-GFDL/FMS/projects/3), under the column _To Merge_.

colingladueNOAA:
@edhartnett  Could you please update this branch to be in line with the base branch so I can merge it in to master today? I do not have permission to update it for you. 

Thanks.

PR #388, opened 2020-04-17
update constants file by moving small_fac options

### LINKED ISSUE #389: Update constants.F90 to support SHiELD development
N/A

Updated constants/constants.F90 logic by moving small_frac options outside of current ifdef/endif pair.  Apply small_frac variable to radius and omega parameters in GFS_PHYS versions of constants.

none

**Additional context**
none

bensonr:
**Description**
Updates constants.F90 moving small_fac logic outside of ifdef/endif region. Necessary to allow SHiELD to utilize constants from FMS.

Fixes #389 

**How Has This Been Tested?**
Model tests run with SHiELD C3072 and C384 configurations

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/constants/constants.F90
+++ b/constants/constants.F90
@@ -37,9 +37,12 @@ public :: version
 
 real :: realnumber !< dummy variable to use in HUGE initializations
 
+!! The small_fac parameter is used to alter the radius of the earth to allow one to
+!! examine non-hydrostatic effects without the need to run full-earth high-resolution
+!! simulations (<13km) that will tax hardware resources.
 #ifdef SMALL_EARTH
 #if defined(DCMIP) || (defined(HIWPP) && defined(SUPER_K))
- real, public, parameter :: small_fac =  1._r8_kind / 120._r8_kind #only needed for supercell test
+ real, public, parameter :: small_fac =  1._r8_kind / 120._r8_kind   ! only needed for supercell test
 #elif defined(HIWPP)
  real, public, parameter :: small_fac = 1._r8_kind / 166.7_r8_kind
 #else
--- a/constants/constants.F90
+++ b/constants/constants.F90
@@ -37,12 +37,24 @@ public :: version
 
 real :: realnumber !< dummy variable to use in HUGE initializations
 
+#ifdef SMALL_EARTH
+#if defined(DCMIP) || (defined(HIWPP) && defined(SUPER_K))
+ real, public, parameter :: small_fac =  1._r8_kind / 120._r8_kind #only needed for supercell test
+#elif defined(HIWPP)
+ real, public, parameter :: small_fac = 1._r8_kind / 166.7_r8_kind
+#else
+ real, public, parameter :: small_fac = 1._r8_kind / 10._r8_kind
+#endif
+#else
+ real, public, parameter :: small_fac = 1._r8_kind
+#endif
+
 #ifdef GFS_PHYS
 ! SJL: the following are from fv3_gfsphysics/gfs_physics/physics/physcons.f90
-real,               public, parameter :: RADIUS = 6.3712e+6_r8_kind           !< Radius of the Earth [m]
-real(kind=r8_kind), public, parameter :: PI_8   = 3.1415926535897931_r8_kind  !< Ratio of circle circumference to diameter [N/A]
-real,               public, parameter :: PI     = 3.1415926535897931_r8_kind  !< Ratio of circle circumference to diameter [N/A] (REAL(KIND=8))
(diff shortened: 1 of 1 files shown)


constants/constants.F90
```diff
@@ -37,12 +37,24 @@ module constants_mod
...
+#ifdef SMALL_EARTH
+#if defined(DCMIP) || (defined(HIWPP) && defined(SUPER_K))
+ real, public, parameter :: small_fac =  1._r8_kind / 120._r8_kind #only needed for supercell test
+#elif defined(HIWPP)
+ real, public, parameter :: small_fac = 1._r8_kind / 166.7_r8_kind
+#else
+ real, public, parameter :: small_fac = 1._r8_kind / 10._r8_kind
+#endif
+#else
+ real, public, parameter :: small_fac = 1._r8_kind
```
thomas-robinson, line 49: 1. `small_fac` should have some kind of documentation to tell us what it is.  I realize it wasn't there before, but these values are very specific
2. Can `small_fac` have more digits?  It seems like the rounding of these numbers is left to the compiler instead of the programmer.  That would change the accuracy.

constants/constants.F90
```diff
@@ -37,12 +37,24 @@ module constants_mod
 
 real :: realnumber !< dummy variable to use in HUGE initializations
 
+#ifdef SMALL_EARTH
+#if defined(DCMIP) || (defined(HIWPP) && defined(SUPER_K))
+ real, public, parameter :: small_fac =  1._r8_kind / 120._r8_kind #only needed for supercell test
```
thomas-robinson, line 42: Should the comment on this have a ! instead of a #

constants/constants.F90
```diff
@@ -37,12 +37,24 @@ module constants_mod
...
+ real, public, parameter :: small_fac = 1._r8_kind
+#endif
+
 #ifdef GFS_PHYS
 ! SJL: the following are from fv3_gfsphysics/gfs_physics/physics/physcons.f90
-real,               public, parameter :: RADIUS = 6.3712e+6_r8_kind           !< Radius of the Earth [m]
-real(kind=r8_kind), public, parameter :: PI_8   = 3.1415926535897931_r8_kind  !< Ratio of circle circumference to diameter [N/A]
-real,               public, parameter :: PI     = 3.1415926535897931_r8_kind  !< Ratio of circle circumference to diameter [N/A] (REAL(KIND=8))
-real,               public, parameter :: OMEGA  = 7.2921e-5_r8_kind   !< Rotation rate of the Earth [1/s]
+real,               public, parameter :: RADIUS = 6.3712e+6_r8_kind * small_fac !< Radius of the Earth [m]
```
thomas-robinson, line 54: Is this answer changing?

constants/constants.F90
```diff
@@ -37,12 +37,24 @@ module constants_mod
...
+real(kind=r8_kind), public, parameter :: PI_8   = 3.1415926535897931_r8_kind    !< Ratio of circle circumference to diameter [N/A]
+real,               public, parameter :: PI     = 3.1415926535897931_r8_kind    !< Ratio of circle circumference to diameter [N/A]
+real,               public, parameter :: OMEGA  = 7.2921e-5_r8_kind / small_fac !< Rotation rate of the Earth [1/s]
```
thomas-robinson, line 57: Is this answer changing?

constants/constants.F90
bensonr: I will add documentation on what small_fac does.

constants/constants.F90
bensonr: Yes it should, although an error will only be triggered in the presence of the proper macros.

constants/constants.F90
bensonr: With the exception of those that are strictly typed, precision of all constants in this file is set at compile time.  small_fac is defined to be consistent with the parameters it is modifying.

constants/constants.F90
thomas-robinson: It's a good thing I had on my glasses then.

constants/constants.F90
bensonr: As this is gated by a #ifdef macro for GFS_PHYS, any answer changes will be limited to SHiELD and UFS models and is already in use by SHiELD.  The GFDL models will be unaffected.

constants/constants.F90
bensonr: As this is gated by a #ifdef macro for GFS_PHYS, any answer changes will be limited to SHiELD and UFS models and is already in use by SHiELD. The GFDL models will be unaffected.

PR #393, opened 2020-04-23
Adds unit test for mpp_pe

thomas-robinson:
**Description**
Adds a unit test for the function [mpp_pe](../blob/master/mpp/include/mpp_util.inc#L433)

Fixes #

**How Has This Been Tested?**
Ran `make check` and travis_ci

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ and this project uses `yyyy.rr[.pp]`, where `yyyy` is the year a patch is releas
 `rr` is a sequential release number (starting from `01`), and an optional two-digit
 sequential patch number (starting from `01`).
 
+## [2020.02] - 2020-05-01
+### Added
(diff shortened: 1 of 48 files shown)


thomas-robinson:
@edhartnett this should have been a WIP, but I forgot to set it as such.  I think get sidetracked, but now I think the test is done.  It no longer calls `mpp_init`.  I wrote a separate routine that does what `mpp_init` does to initialize MPI and set up the PE lists.  

These tests are being set up so that we can test what is called in `mpp_init`.  After the tests for these routines are created, we can then add an mpp_init test.

thomas-robinson:
@edhartnett we also have decided to merge tests as they are done because they don't affect the code (as you suggested).  This group of tests are being sort of lumped together as they are part of a group effort.

edwardhartnett:
@thomas-robinson good to know! I will get to work adding more tests. ;-)

edwardhartnett:
Glad to see this merged!

BTW it's not necessary or desirable for each test to get a .sh file. It leads to an unnecessary proliferation of files.

What we do in netcdf is group tests in a .sh file. So for example, put all your tests for each mpp module in a .sh file. Then you can test just one mod file conveniently from the command line, and also easily find tests related to a mod file, without having many different test shell scripts.

PR #395, opened 2020-04-23
Adds unit test for mpp_npes

thomas-robinson:
**Description**
Adds a unit test for [mpp_npes](https://github.com/NOAA-GFDL/FMS/blob/master/mpp/include/mpp_util.inc#L465) function

Fixes #

**How Has This Been Tested?**
Ran with `make check` using the intel19 compiler

**Checklist:**
- [x] `make distcheck` passes


thomas-robinson:
@colingladueNOAA and @uramirez8707 this test shouldn't rely on having input.nml. I think this fixes the problem.  

Also this code is merged with the mpp_pe test in #393

thomas-robinson:
> Uriel pointed out that this does not pass a skylake box make check (serial testing rather than parallel). I believe this needs to create an input.nml for itself if run first, and works on travis in parallel because one is being created by another test.

The current version passes `make check` for me

PR #398, opened 2020-04-27
Change write_restart to write out all meta_data before data

### LINKED ISSUE #396: Write_restart changes between writing meta_data and data
**Describe the bug**
I created a unit test that writes a restart using fms2io's `write_restart` and analysed it with Darshan. Results showed fms2io was writing twice the number of bytes as the actual size of the file and as fmsio. Part of the problem is because write_restarts changes between writing data and writing meta_data (the checksums)

See [here](https://github.com/NOAA-GFDL/FMS/blob/e6fd03b070eb1ba8ffbca740fd69681ee16d6be7/fms2_io/fms_netcdf_domain_io.F90#L549)

Write_restart needs to be modified to write meta data before data, to avoid this.

Fms2io implementation in model components should buffer the dummy axes and write the data with `write_restart`

**To Reproduce**
See above

**Expected behavior**
 Meta data should be written before all data 

**System Environment**
This applies to any system, but Darshan is installed in GAEA.  

**Additional context**
N/A

uramirez8707:
**Description**
write_restart now loops through the restart_variables, calculates the checksum and writes out the metadata. Then it loops through the restart_variables (again!) and writes out the data. 

Fixes #396 

**How Has This Been Tested?**
I have a "unit" tests that writes a restart, using `write_restart` that has two restart variables (on a cube sphere grid), and I used Darshan to analyze performance. 

If I run it with current master, for each *tileX.nc file, Darshan outputs: 
`POSIX_BYTES_WRITTEN     8852488`

even though the size of the file is `4425424`

Data is in: /lustre/f2/darshan/2020/4/27/Uriel.Ra_current_master_id62720_4-27-56196-18121498855530832921_1.darshan

If you run it with this fix, for each *tileX.nc file, Darshan outputs: 
`POSIX_BYTES_WRITTEN     4427172`

The file created from current master and from this fix is exactly the same. 

**Checklist:**
- [X] `make distcheck` passes

### DIFF
--- a/fms2_io/fms_netcdf_domain_io.F90
+++ b/fms2_io/fms_netcdf_domain_io.F90
@@ -546,16 +546,14 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
   if (.not. fileobj%is_restart) then
     call error("file "//trim(fileobj%path)//" is not a restart file.")
   endif
+
+! Calculate the checksum and write it out
   do i = 1, fileobj%num_restart_vars
     if (associated(fileobj%restart_vars(i)%data0d)) then
-      call domain_write_0d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data0d, unlim_dim_level=unlim_dim_level)
+       cycle
     elseif (associated(fileobj%restart_vars(i)%data1d)) then
-      call domain_write_1d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data1d, unlim_dim_level=unlim_dim_level)
+       cycle
     elseif (associated(fileobj%restart_vars(i)%data2d)) then
-      call domain_write_2d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data2d, unlim_dim_level=unlim_dim_level)
       chksum = compute_global_checksum(fileobj, fileobj%restart_vars(i)%varname, &
                                        fileobj%restart_vars(i)%data2d, is_decomposed)
       if (is_decomposed) then
@@ -563,8 +561,6 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
                                          "checksum", chksum)
       endif
     elseif (associated(fileobj%restart_vars(i)%data3d)) then
-      call domain_write_3d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data3d, unlim_dim_level=unlim_dim_level)
       chksum = compute_global_checksum(fileobj, fileobj%restart_vars(i)%varname, &
                                        fileobj%restart_vars(i)%data3d, is_decomposed)
       if (is_decomposed) then
@@ -572,8 +568,6 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
                                          "checksum", chksum)
       endif
     elseif (associated(fileobj%restart_vars(i)%data4d)) then
-      call domain_write_4d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data4d, unlim_dim_level=unlim_dim_level)
       chksum = compute_global_checksum(fileobj, fileobj%restart_vars(i)%varname, &
                                        fileobj%restart_vars(i)%data4d, is_decomposed)
       if (is_decomposed) then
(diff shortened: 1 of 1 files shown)


fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -584,6 +578,29 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
       call error("this branch should not be reached.")
```
thomas-robinson, line 578: This is not a good error message

thomas-robinson:
@uramirez8707 I just want to make sure I understand what you're doing.  Instead of one loop, you've set up two loops.  The first loop finds the checksums and registers them as variable metadata.  Then you do your writes.  Do I have that right?

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,34 +546,28 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
...
+! Calculate the checksum and write it out
   do i = 1, fileobj%num_restart_vars
     if (associated(fileobj%restart_vars(i)%data0d)) then
-      call domain_write_0d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data0d, unlim_dim_level=unlim_dim_level)
+       cycle
     elseif (associated(fileobj%restart_vars(i)%data1d)) then
-      call domain_write_1d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data1d, unlim_dim_level=unlim_dim_level)
+       cycle
```
thomas-robinson, line 555: Can these be combined or removed since nothing is happening?

uramirez8707:
> @uramirez8707 I just want to make sure I understand what you're doing. Instead of one loop, you've set up two loops. The first loop finds the checksums and registers them as variable metadata. Then you do your writes. Do I have that right?

The first loops **calculates** the checksum and **writes** the metadata. 
Believe it or not, `register_variable_attribute` **doesn't** actually registers metadata, it just writes it .. (see https://github.com/NOAA-GFDL/FMS/blob/e6fd03b070eb1ba8ffbca740fd69681ee16d6be7/fms2_io/include/register_variable_attribute.inc#L43)

fms2_io/fms_netcdf_domain_io.F90
uramirez8707: They can't be removed because, if you have a 0d or 1d variable, it will end up in: 
```
    else
       call error("this branch should not be reached.")
     endif
```

thomas-robinson: The `if`s should be combined then instead of two different `if`s.

uramirez8707:
> * The first `if` and `elseif` should be combined.

Done!
> * There should be some comments to explain what's happening.

There is a comment at the top of the first loop: 
https://github.com/NOAA-GFDL/FMS/blob/657e59ae42db9076eca8ac15fb010a1e911e2e01/fms2_io/fms_netcdf_domain_io.F90#L550
And at the top of the second loop: 
https://github.com/NOAA-GFDL/FMS/blob/657e59ae42db9076eca8ac15fb010a1e911e2e01/fms2_io/fms_netcdf_domain_io.F90#L582

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,44 +546,59 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
...
+
+! Calculate the checksum and write it out
   do i = 1, fileobj%num_restart_vars
-    if (associated(fileobj%restart_vars(i)%data0d)) then
-      call domain_write_0d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data0d, unlim_dim_level=unlim_dim_level)
-    elseif (associated(fileobj%restart_vars(i)%data1d)) then
-      call domain_write_1d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data1d, unlim_dim_level=unlim_dim_level)
+    if (associated(fileobj%restart_vars(i)%data0d) .and. associated(fileobj%restart_vars(i)%data1d)) then
```
menzel-gfdl, line 552: I don't think this is right.  You should never have both data0d and data1d associated at the same time.  I think a better solution is to delete this entire branch of the if statement, and then delete the else branch on line 575.

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,44 +546,59 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
...
     elseif (associated(fileobj%restart_vars(i)%data4d)) then
-      call domain_write_4d(fileobj, fileobj%restart_vars(i)%varname, &
-                           fileobj%restart_vars(i)%data4d, unlim_dim_level=unlim_dim_level)
       chksum = compute_global_checksum(fileobj, fileobj%restart_vars(i)%varname, &
                                        fileobj%restart_vars(i)%data4d, is_decomposed)
       if (is_decomposed) then
         call register_variable_attribute(fileobj, fileobj%restart_vars(i)%varname, &
                                          "checksum", chksum)
       endif
     else
```
menzel-gfdl, line 575: I think this else branch can be deleted, see my other comment above.

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,44 +546,55 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
   if (.not. fileobj%is_restart) then
     call error("file "//trim(fileobj%path)//" is not a restart file.")
   endif
+
+! Calculate the checksum and write it out
```
thomas-robinson, line 550: ...to the file?  ...to the variable metadata?  ...to the screen?

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,44 +546,55 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
...
       chksum = compute_global_checksum(fileobj, fileobj%restart_vars(i)%varname, &
                                        fileobj%restart_vars(i)%data4d, is_decomposed)
       if (is_decomposed) then
         call register_variable_attribute(fileobj, fileobj%restart_vars(i)%varname, &
                                          "checksum", chksum)
       endif
+    endif
+  enddo
+
+! Write out all of the data
```
thomas-robinson, line 576: ...to the file? ...to the variable metadata? ...to the screen?

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -546,44 +546,55 @@ subroutine save_domain_restart(fileobj, unlim_dim_level)
...
+                           fileobj%restart_vars(i)%data2d, unlim_dim_level=unlim_dim_level)
+    elseif (associated(fileobj%restart_vars(i)%data3d)) then
+      call domain_write_3d(fileobj, fileobj%restart_vars(i)%varname, &
+                           fileobj%restart_vars(i)%data3d, unlim_dim_level=unlim_dim_level)
+    elseif (associated(fileobj%restart_vars(i)%data4d)) then
+      call domain_write_4d(fileobj, fileobj%restart_vars(i)%varname, &
+                           fileobj%restart_vars(i)%data4d, unlim_dim_level=unlim_dim_level)
     else
-      call error("this branch should not be reached.")
+      call error("save_domain_restart: this branch should not be reached.")
```
thomas-robinson, line 594: This error message is still not descriptive of the problem.  If there is a fatal, then I know that I shouldn't have gotten here.  The user will be puzzled as to why they got here or how to fix it.

colingladueNOAA:
@thomas-robinson Are you satisfied with the changes made to this PR?

PR #407, opened 2020-04-29
Fix for get_ascii_file_num_lines issue

### LINKED ISSUE #399: MPP function get_ascii_file_num_lines returns wrong value
**Describe the bug**
The MPP function get_ascii_file_num_lines returns a value that is one higher than the number of lines in the file. E.g. if there are 3 lines in an ASCII file, the function will return the value 4.

**To Reproduce**
Call the MPP function get_ascii_file_num_lines on an ASCII file with a known number of lines. E.g. run this sample program:

```
program test_get_ascii_lines
  use mpp_mod, only : get_ascii_file_num_lines, mpp_init, mpp_exit, INPUT_STR_LENGTH

  implicit none

  integer :: my_known_lines=5
  integer :: my_guess_lines_in_file
  character(len=10) :: file_5_line="ascii_5   " ! ascii_5 contents listed below the program

  call mpp_init()
  ! Check on 5 line file
  my_guess_lines_in_file = get_ascii_file_num_lines(trim(file_5_line), INPUT_STR_LENGTH)
  print *, "My Guess", my_guess_lines_in_file
  print *, "Known Lines", my_known_lines
  call mpp_exit()
end program test_get_ascii_lines
```

Contents of `ascii_5` file:
```
uno
dos
tres
quatro
cinco
```

**Expected behavior**
I expect the function to return the correct number of lines

**System Environment**
- OS: Skylake Box
- Compiler(s): Intel 18
- MPI type, and version (e.g. MPICH, Cray MPI, openMPI): Default
- netCDF Version: For both C and Fortran: Default
- Configure options: Any additional flags, or macros passed to configure

**Additional context**
I'm looking at the function now, and I'm wondering whether or not num_lines should be 0 here instead of 1?

https://github.com/NOAA-GFDL/FMS/blob/e6fd03b070eb1ba8ffbca740fd69681ee16d6be7/mpp/include/mpp_util.inc#L1413

GFDL-Eric:
Changed num_lines starting value to 0 from 1 to avoid miscount in the get_ascii_file_num_lines function in mpp.

**Description**
This PR addresses the miscount shown in #399 by setting the initial value of `num_lines` in the function `get_ascii_file_num_lines` to 0 (the previous value was 1).

Fixes #399 

**How Has This Been Tested?**
Makes the unit test that I've written for the get_ascii_file_num_lines function pass instead of fail, as it now outputs the correct number of lines for a file with a known line count. n.b. that unit test will be submitted in a future PR

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -1410,7 +1410,7 @@ end function rarray_to_char
              call mpp_error(FATAL, 'get_ascii_file_num_lines: Error opening file:' //trim(FILENAME)// &
                             '.  (IOSTAT = '//trim(text)//')')
           else
-             num_lines = 1
+             num_lines = 0
              do
                 read (UNIT=f_unit, FMT='(A)', IOSTAT=status) str_tmp
                 if ( status .lt. 0 ) exit


thomas-robinson:
@GFDL-Eric I updated to reflect that this does not pass `make distcheck`.  It doesn't appear to be as simple as you implied.

GFDL-Eric:
When I ran make distcheck, the failure seemed to be in the diag_manager test:

```
Test 1: Data array is too large in x and y direction
forrtl: severe (408): fort: (2): Subscript #1 of the array CONTENT has value 22 which is greater than the upper bound of 21
```

When I look at the input.nml file, it only has 21 lines, so I'm not sure if there's something funky going on with `input_nml_file` or what. I'll keep looking into it.

GFDL-Eric:
So I did some digging into read_ascii_file, and it seems that allocating an extra space is done on purpose. The lines of the file are currently read directly into each element of the `Content` array, so the last read needs a valid array index of `num_lines + 1`, however this value will never be used. I hope there is a workaround to this so that the get_ascii_file_num_lines doesn't have to be an erroneous value.

thomas-robinson:
@GFDL-Eric this isn't passing the CI.  Is there an intent to fix it?

GFDL-Eric:
Um, no, since I probably should've closed this PR as we decided that
returning the "incorrect" number of lines was actually the correct
behavior. I'll go ahead and close this.

GFDL-Eric:
Closing this PR as we are leaving the behavior of get_ascii_file_num_lines as is (e.g. it returns an incorrect number of lines, but does so in a self-consistent manner, while the fix would produce a different, Fortran-specific error with no known workaround).

PR #434, opened 2020-05-28
Fixes regional diagnostic error (issue # 409) by adding global attrib…

### LINKED ISSUE #409: regional diagnostics are different when using Ocean mask_table
**Describe the bug**
Updating FMS and FMScoupler to 2020.02-beta3 caused the ocean regional diagnostics be different when using a mask_table from none mask_table version. The region is different (has smaller longitudinal extent with the mask_table).
The number of diag_axis are different!

```
%  ncdump -h /lustre/f2/scratch/Niki.Zadeh/work/19480101.ocean_Bering_Strait_xanadu.nc | grep 'xh_sub01 = '
xh_sub01 = 6 ;
% ncdump -h /lustre/f2/scratch/Niki.Zadeh/work/19480101.ocean_Bering_Strait_no_mask_xanadu.nc | grep 'xh_sub01 = '  
xh_sub01 = 6 ;
% ncdump -h /lustre/f2/scratch/Niki.Zadeh/work/19480101.ocean_Bering_Strait_no_mask_2020.02.nc | grep 'xh_sub01 = '
xh_sub01 = 6 ;
% ncdump -h /lustre/f2/scratch/Niki.Zadeh/work/19480101.ocean_Bering_Strait_2020.02.nc | grep 'xh_sub01 = '
xh_sub01 = 3 ;
```
Honestly, I don't know what the  xh_sub01 value should be. But the region looks totally different.

Note: The non-mask_table file is bitwise identical to xanadu non-mask_table file!

**To Reproduce**
```
git clone -b RTS git@gitlab.gfdl.noaa.gov:mdt/xml.git mdt_xml_RTS
cd mdt_xml_RTS
fremake -x OM4_COBALT_postxanadu.xml -p ncrc4.intel16 -t prod MOM6_SIS2_compile
frerun -x OM4_COBALT_postxanadu.xml -p ncrc4.intel16 -t prod -r debug1 OM4p5_CORE2_IAF_COBALT_abio_csf_mle200
```
compare  the history file 19480101.ocean_Bering_Strait.nc between the 512 (non mask_table) and 432 (mask_table) runs.

**Expected behavior**
The history files should be identical between mask_table and non-mask_table runs
like it was in xanadu. At the least they regions should look the same.

**System Environment**
ncrc4.intel16

**Additional context**
Add any other context about the problem.  If applicable, include where any files
that help describe, or reproduce the problem exist.

ngs333:
Fixes errors in writing regional diagnostics using mask tables by registering global attribute "NumFilesInSet" and per-axis attribute "domain_decomposition" for regional X and Y axis.
NumFilesInSet is always set equal to 0. File diag_output.F90 was changed.

Fixes #409

**How Has This Been Tested?**
Ocean model output for regional diagnostic with mask tables were performed on Xanadu and 2020.02 code base (plus this fix) and compared.

**Checklist:**
- [X] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -212,7 +212,10 @@ CONTAINS
         if (.not.check_if_open(fileob)) then
                call open_check(open_file(fileobjND, trim(fname_no_tile)//".nc."//trim(mype_string), "overwrite", &
                             nc_format="64bit", is_restart=.false.))
-        endif
+               !< For regional subaxis add the NumFilesInSet attribute, which is added by fms2_io for (other)
+               !< domains with sufficient decomposition info. Note mppnccombine will work with an entry of zero.
+               call register_global_attribute(fileobjND, "NumFilesInSet", 0)
+       endif
        fnum_domain = "nd" ! no domain
        if (file_unit < 0) file_unit = 10
      endiF
@@ -401,12 +404,8 @@ integer :: domain_size, axis_length, axis_pos
                               case (-1)
                                    call register_variable_attribute(fptr, axis_name, "positive", "down")
                          end select
-                         !< For regional subaxis add the NumFilesInSet and domain_decomposition atibutes, which
-                         !< are automatically added by fms2_io for (other) domains for which it has sufficient
-                         !< decomposition info. Note mppnccombine will work with an entry of zero.
-                         if (.not. global_att_exists(fptr,"NumFilesInSet")) then
-                            call register_global_attribute(fptr, "NumFilesInSet", 0)
-                         endif
+                         !< For regional subaxis add the "domain_decomposition" attribute, which is added
+                         !< fms2_io for (other) domains with sufficient decomposition info.
                          call register_variable_attribute(fptr, axis_name, "domain_decomposition", &
                               (/gstart, gend, cstart, cend/))
                          call write_data(fptr, axis_name, axis_data(istart:iend) )
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -212,6 +212,7 @@ CONTAINS
         if (.not.check_if_open(fileob)) then
                call open_check(open_file(fileobjND, trim(fname_no_tile)//".nc."//trim(mype_string), "overwrite", &
                             nc_format="64bit", is_restart=.false.))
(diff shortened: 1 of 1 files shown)


diag_manager/diag_output.F90
```diff
@@ -212,6 +212,7 @@ SUBROUTINE diag_output_init(file_name, FORMAT, file_title, file_unit,&
         if (.not.check_if_open(fileob)) then
                call open_check(open_file(fileobjND, trim(fname_no_tile)//".nc."//trim(mype_string), "overwrite", &
                             nc_format="64bit", is_restart=.false.))
+               call register_global_attribute(fileobjND, "NumFilesInSet", 0)
```
thomas-robinson, line 215: This will add the attribute to any file with the `fileobjND`.  Will this add problems to files that don't have to be combined or use the X/Y axes?  

Can this be added where the axis is written?  Will it write it twice if the x-axis and y-axis are both register?  @menzel-gfdl maybe you have some insight into these questions.

wrongkindofdoctor: You can add a check `if (global_att_exists(fileobjND, "NumFilesinSet")) then...` to prevent potential errors from trying to register the attribute multiple times.

thomas-robinson: If this only needs to be done when there is a x or y axis, then perhaps it should be moved and checked as @wrongkindofdoctor suggested when the[ axis is being registered](https://github.com/NOAA-GFDL/FMS/blob/master/diag_manager/diag_output.F90#L389)

menzel-gfdl: It was put here to mimic the behavior of `mpp_open`, which we know is compatible with `mppnccombine`.  It is not clear to me if/why `mppnccombine` requires this attribute for regional files, since it usually sets this value to zero even when the region is split over multiple ranks.

thomas-robinson: I realize why this code exists.  But this will write the attribute for all of the files that aren't regional diagnostics.  I don't think we want that

thomas-robinson: Upon further consultation and review, this ia fine.

ngs333:
Per the decisions in the FMS2_IO meeting on Jun 1, I moved the registering of the global attribute NumFileInSet back tnto the function "diag_output_init()"

PR #444, opened 2020-06-02
Adds unit test for get_ascii_file_num_lines (now works with Travis)

GFDL-Eric:
**Description**
This is a suite of unit tests (currently 5 tests but may be expanded) for the get_ascii_file_num_lines mpp function.

Fixes # (issue)

**How Has This Been Tested?**
Ran make distcheck as well as make check in the mpp folder. This was with Intel19 on the Skylake box.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/mpp/test_mpp_get_ascii_lines2.sh
+++ b/test_fms/mpp/test_mpp_get_ascii_lines2.sh
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+#***********************************************************************
+#*                   GNU Lesser General Public License
+#*
+#* This file is part of the GFDL Flexible Modeling System (FMS).
+#*
+#* FMS is free software: you can redistribute it and/or modify it under
+#* the terms of the GNU Lesser General Public License as published by
+#* the Free Software Foundation, either version 3 of the License, or (at
+#* your option) any later version.
+#*
+#* FMS is distributed in the hope that it will be useful, but WITHOUT
+#* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+#* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+#* for more details.
+#*
+#* You should have received a copy of the GNU Lesser General Public
+#* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+#***********************************************************************
+
+# This is part of the GFDL FMS package. This is a shell script to
+# execute tests in the test_fms/mpp directory.
+
+# Ed Hartnett 11/29/19
+
+# Set common test settings.
+. ../test_common.sh
+
+skip_test="no"
+
+# Copy file for test.
+cp $top_srcdir/test_fms/mpp/base_ascii_5 ascii_5
+cp $top_srcdir/test_fms/mpp/base_ascii_25 ascii_25
+cp $top_srcdir/test_fms/mpp/base_ascii_0 ascii_0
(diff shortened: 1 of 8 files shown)


colingladueNOAA:
Eric,
Seth suggested to me that I should generate the ascii files I am using for testing during the running of the test rather than store them in the test_fms/mpp directory permanently. Do you think this would be a good idea here?

GFDL-Eric:
I certainly understand not wanting to store extra files in the repo, even ones of a small size such as these. My hesitation would be in generating the files from within the Fortran version of the test, as we would then be testing the capability of write_ascii, when that's not the point of this unit test. However, I think we could do it in the context of the test script (e.g. the test_mpp_get_ascii_lines2.sh script). Maybe something with shell functions like `echo` and the `>` operator?

colingladueNOAA:
You could try something like this in your bash file:

touch test_numb_base.nml
echo "&test_read_input_nml_nml" > test_numb_base.nml
echo "test_numb = 0" >> test_numb_base.nml
echo "/" >> test_numb_base.nml

Where ">" overwrites contents and ">>" appends to the end of the file

edwardhartnett:
If I may relate the netCDF experience: we do have many small text files in the distribution. Particularly in the ncdump directory, which is full of files called ref_SOMETHING.cdl. We test ncdump by creating a netCDF file in code, then running ncdump on the created file, and comparing the output to the ref file. For example, from ncdump/run_tests.sh, which uses a file called ref_tst_small.cdl:


```
#!/bin/sh
# This shell script runs the ncdump tests.
# Ed Hartnett, Dennis Heimbigner

if test "x$srcdir" = x ; then srcdir=`pwd`; fi
. ../test_common.sh

set -e
echo ""
echo "*** Testing ncgen and ncdump using some test CDL files."
echo "*** creating tst_small.nc from ref_tst_small.cdl..."
${NCGEN} -b -o tst_small.nc $srcdir/ref_tst_small.cdl
echo "*** creating tst_small.cdl from tst_small.nc..."
${NCDUMP} tst_small.nc > tst_small.cdl
diff -b -w tst_small.cdl $srcdir/ref_tst_small.cdl
```
(Note that there is some magic in the test_common.sh which is required to correctly set $srcdir for all out of path builds of the software.)

You also have to remember to include the ref file in the EXTRA_DIST target of Makefile.am, and of course the generated test file in CLEANFILES. 

I would also suggest you pick some prefix for text files that are in the repo for this reason (as we used "ref_" in netCDF). This makes it easy to spot these files.

test_fms/mpp/test_mpp_get_ascii_lines2.sh
```diff
@@ -0,0 +1,55 @@
...
+#* for more details.
+#*
+#* You should have received a copy of the GNU Lesser General Public
+#* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+#***********************************************************************
+
+# This is part of the GFDL FMS package. This is a shell script to
+# execute tests in the test_fms/mpp directory.
+
+# Ed Hartnett 11/29/19
```
thomas-robinson, line 25: @GFDL-Eric Ed didn't make this file, you did.

GFDL-Eric: Good catch! Since this is already merged I can edit this in either a tiny separate PR or lump it into a different pull (e.g. another PR that deals with the testing suite).

Issue #244, opened 2019-12-16
Can I add some tests for fms2_io and other modules in support of PIO integration?


edwardhartnett:
**Is your feature request related to a problem? Please describe.**
To support future IO work, I would like to add some tests for the I/O code, mpp/mpp_io.F90.

**Describe the solution you'd like**
I will add a .F90 test program, and unit test some of the functions in mpp_io.mod. Preferably with and without using MPI.

**Describe alternatives you've considered**
There is no alternative to good testing, if you want your code to be correct.

**Additional context**
I'm happy to keep these tests up to date with the upcoming new IO code, or else to write these tests on that branch, for the new code.

This is in support of #100

underwoo:
It might be best to use the new FMS code in branch `fms-io-dev`.  @wrongkindofdoctor can help direct your efforts in writing these tests, as they may already have some written.

underwoo:
The new fms_io code is now in the master branch.  We will do some cleanup on the MPP code, but feel free to write whatever check tests you would like.

edwardhartnett:
OK great! At the moment I am adding some new compression options to netcdf. (Look for them soon!)

I will circle around to this again and add some tests...

edwardhartnett:
This should not be closed. THere are more tests coming.

PR #461, opened 2020-06-15
More testing for fms2_io

edwardhartnett:
**Description**
More development of I/O tests.

Fixes # (issue)

Part of #244
**How Has This Been Tested?**
Tested on my Linux machine and travis.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/fms2_io/test_io_simple.F90
+++ b/test_fms/fms2_io/test_io_simple.F90
@@ -30,17 +30,36 @@ program test_io_simple
   use mpp_mod
   use setup
   use netcdf
+  implicit none
   
   type(Params) :: test_params
   type(domain2d) :: domain
   integer :: err
   type(domain2d), pointer :: io_domain
+  integer :: isc
+  integer :: iec
+  integer :: nx
+  integer :: jsc
+  integer :: jec
+  integer :: ny
+  integer :: isd
+  integer :: nxd
+  integer :: jsd
+  integer :: nyd
+  integer :: nx_east
+  integer :: isc_east
+  integer :: iec_east
+  integer :: ny_north
+  integer :: jsc_north
+  integer :: jec_north
   type(FmsNetcdfDomainFile_t) :: fileobj
   integer, parameter :: ntiles = 6
   integer :: ncchksz = 64*1024 
   character (len = 10) :: netcdf_default_format = "64bit"
   integer :: header_buffer_val = 16384
   integer :: ncid
+  character (len = 80) :: testfile
+  integer :: numfilesatt
 
   ! Initialize.
   call init(test_params, ntiles)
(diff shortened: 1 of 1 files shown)


edwardhartnett:
@thomas-robinson GNU.

But I now have access to an ifort compiler on one of my testing machines. I will add that to the mix today... ;-)

test_fms/fms2_io/test_io_simple.F90
```diff
@@ -69,22 +105,75 @@ program test_io_simple
...
 
-  ! Open a netCDF file and initialize the file object.
-  call open_check(open_file(fileobj, "test_io_simple.nc", "overwrite", &
-       domain, nc_format="64bit", is_restart=.false.))
+  do i = 1, 3
+     write(testfile,'(a,a,a)') 'test_io_simple_', trim(my_format(i)), '.nc'
+     
+     ! Open a netCDF file and initialize the file object.
+     call open_check(open_file(fileobj, testfile, "overwrite", &
+          domain, nc_format=my_format(1), is_restart=.false.))
```
uramirez8707, line 113: shouldn't this be `i` instead of 1?
```f90
domain, nc_format=my_format(i), is_restart=.false.))
```

test_fms/fms2_io/test_io_simple.F90
```diff
@@ -69,22 +105,75 @@ program test_io_simple
...
+     ! Close the file.
+     call close_file(fileobj)
+     
+     call mpi_barrier(mpi_comm_world, err)
+     call mpi_check(err)
+
+     ! Check for expected netcdf file.
+     if (mpp_pe() .eq. mpp_root_pe()) then
+        write(testfile,'(a,a,a)') 'test_io_simple_', trim(my_format(i)), '.tile1.nc'
+        err = nf90_open(testfile, nf90_nowrite, ncid)
```
uramirez8707, line 135: Perhaps you should check if the file_type is correct?

edwardhartnett: Thanks! Good catch! I will fix...

edwardhartnett:
OK, here's a question: does FMS always require netCDF to support netCDF/HDF5?

That is, is FMS intended to also work on classic-only netCDF builds?

If so, we must put a little magic in the configure.ac to detect whether netCDF supports netCDF/HDF, and then we must change this test to only attempt to use netCDF/HDF5 format when the underlying netcdf-c library supports it.

Alternatively, if you say that HDF5 is always to be supported, we need to add a check in configure.ac to ensure that is the case, and give the user a good error message at configure-time.

test_fms/fms2_io/test_io_simple.F90
edwardhartnett: OK, turns out that the command to test the file type is not present in the netCDF F90 API! Whoops! This is a mistake I made about 15 years ago. ;-)

Well, I have added a ticket to the netcdf-fortran project, and I will hopefully get around to adding the nf90_inq_format() function soon, but it won't be available until the next release of netcdf-fortran anyway, so we don't want to use it in the test yet. So we will just skip testing the format for now.

edwardhartnett:
Since #466 has been decided in favor of requiring HDF5, this test code may be merged.

edwardhartnett:
Can we get these changes merged please? I have some other PRs to put up as well...

rem1776:
Tested on skylake with intel/2020 and found no issues

edwardhartnett:
Has this been sufficiently reviewed and testing to merge?

PR #472, opened 2020-06-18
Added new macro for correct definition of transfer intrinsic mold

### LINKED ISSUE #471: checksums of 32-bit real quantities were not correct or reproducible
**Describe the bug**
The mpp_chksum interface that accepts 32-bit reals produces an integer checksum that is not correct or reproducible

**To Reproduce**
Compile FMS with the OVERLOAD_R4 macro and perform an mpp_chksum on an array of 32-bit reals on a single MPI-rank.  Then perform the same checksum reversing the order of the array.  

**Expected behavior**
I  expect a valid and reproducible integer checksum regardless of the data.

**System Environment**
gaea and hera using any Intel compiler version

**Additional context**

bensonr:
**Description**
Added a new macro to properly define the 'mold' argument to the F90 transfer intrinsic.  This new macro allows mpp_chksum to work correctly for both 32 and 64 bit inputs.

Fixes #471 

**How Has This Been Tested?**
A standalone unit test was developed to demonstrate and eventually solve the problem.  This unit test will become a part of the standard unit test suite in another commit/PR.  The update was also tested in a doubly-periodic model where it was previously failing.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_chksum.h
+++ b/mpp/include/mpp_chksum.h
@@ -23,7 +23,7 @@ function MPP_CHKSUM_( var, pelist , mask_val)
 !result is LONG_KIND, which will actually be int ifdef no_8byte_integers
   !optional mask_val is masked away in checksum_int.h function via PACK()
   integer(LONG_KIND) :: MPP_CHKSUM_
-  integer(LONG_KIND) :: mold(1)
+  integer(MPP_TRANSFER_KIND_) :: mold(1)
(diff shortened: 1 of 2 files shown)


colingladueNOAA:
@bensonr Could you please update this to be even with the base branch? Thank you.

rem1776:
Tested on skylake with intel/2020 and had no issues

PR #477, opened 2020-06-22
Added additional checks to prevent time_bnds from being overwritten

### LINKED ISSUE #463: diag_manager overwrites time_bnds for each variable that was register
**Describe the bug**
After diag-manager_end is called: 

1. It loops through the number of files and calls `closing_file`: 
https://github.com/NOAA-GFDL/FMS/blob/37a607ae03ee9e05d4404606e602f5a6796e5339/diag_manager/diag_manager.F90#L3697-L3699
2. In `closing_file` it loops through the num_fields and calls `writing_field`
https://github.com/NOAA-GFDL/FMS/blob/37a607ae03ee9e05d4404606e602f5a6796e5339/diag_manager/diag_manager.F90#L3727
...
https://github.com/NOAA-GFDL/FMS/blob/37a607ae03ee9e05d4404606e602f5a6796e5339/diag_manager/diag_manager.F90#L3760
3. In `writing_field`, `diag_data_out` is called
https://github.com/NOAA-GFDL/FMS/blob/37a607ae03ee9e05d4404606e602f5a6796e5339/diag_manager/diag_manager.F90#L3535
4. In `diag_data_out`, it loops through the num_fields **AGAIN** and it writes out the time_bnds, if`num == field`
https://github.com/NOAA-GFDL/FMS/blob/37a607ae03ee9e05d4404606e602f5a6796e5339/diag_manager/diag_util.F90#L2555-L2581

This means that time_bnds is going to be rewritten for every field because of the **FIRST**  loop through the num_fields in step 2 and the values for the last variables in `num_field` will win :) 
It looks like the same thing is happening with `Time` as well...

This may cause the problem in issue #448 if the last field was registered but data was never sent since `start_dif` will be set to the initial model time. 

**To Reproduce**
Run the awg_xml in 448. Time_bnds is wrong because send_data is never called on the last variable that gets registered (the last variable in the loop in step 2). 

I was able to reproduce the issue using the diag_table and input_nml in `/lustre/f2/scratch/gfdl/Uriel.Ramirez/work/c96L33_am4p0_1x12m0d_432x1a.o268649219`
where I only write 1 history file with two variables for two days. 

**Expected behavior**
`Time_bnds` should be written correctly and maybe only once?

**System Environment**
This occurs in any system. 

**Additional context**
N/A

rem1776:
**Description**
Added additional checks of dif values to prevent incorrect duplicate writes on time_bnds
Fixes #463 

**How Has This Been Tested?**
Tested through Travis CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_util.F90
+++ b/diag_manager/diag_util.F90
@@ -2556,8 +2556,8 @@ CONTAINS
        num = files(file)%fields(i)
        IF ( output_fields(num)%time_ops .AND. &
             input_fields(output_fields(num)%input_field)%register) THEN
-          ! dif needs to be between start_dif and end_dif to prevent duplicate writes on time_bnds
-          IF ( num == field .AND. dif >= start_dif .AND. dif <= end_dif) THEN
+          ! time needs to be between start_dif and end_dif to prevent duplicate writes on time_bnds
+          IF ( num == field .AND. files(file)%rtime_current >= start_dif .AND. files(file)%rtime_current <= end_dif) THEN
              ! Output the axes if this is first time-averaged field
              time_data(1, 1, 1, 1) = start_dif
              call diag_field_write (files(file)%f_avg_start, time_data(1:1,:,:,:), file_num=file, &
--- a/diag_manager/diag_util.F90
+++ b/diag_manager/diag_util.F90
@@ -2556,7 +2556,8 @@ CONTAINS
        num = files(file)%fields(i)
        IF ( output_fields(num)%time_ops .AND. &
             input_fields(output_fields(num)%input_field)%register) THEN
-          IF ( num == field ) THEN
+          ! dif needs to be between start_dif and end_dif to prevent duplicate writes on time_bnds
+          IF ( num == field .AND. dif >= start_dif .AND. dif <= end_dif) THEN
              ! Output the axes if this is first time-averaged field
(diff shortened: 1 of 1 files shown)


uramirez8707:
@thomas-robinson @rem1776 

This does not work because: 
Before `diag_data_out` is called, `middle_time` is defined using the `last_output` and `next_output`:
https://github.com/NOAA-GFDL/FMS/blob/764513e8b847c5d687925376e2f47d9cfda9b956/diag_manager/diag_manager.F90#L3531

This `middle_time` is an input to the `diag_data_out` call, which is used to defined `diff`. So the if statement that was added will always be true. 

One way to fix this, is to use `files(file)%rtime_current` instead of `diff` which is set to the correct time. This works for me ...

colingladueNOAA:
@rem1776  Hey how did you defeat travis?!

thomas-robinson:
@uramirez8707 did you test this with the case you had set up?

uramirez8707:
@thomas-robinson Sorry forgot to update the PR

I tested this with the unit test I had set up and I also ran a year run of AM4-c96L33_am4p0 experiment and I now get the correct time_bnds.

rem1776:
Made a new PR for this since I messed up the branch and it got deleted  
It's now in #512

Issue #100, opened 2019-07-23
adding support for PIO to FMS for better I/O performance


edhartnett:
PIO (https://github.com/NCAR/ParallelIO) is a HPC I/O library that has been used in CESM at NCAR, and has recently become an output option for WRF.

PIO allows the same code to use netCDF classic, netCDF-4 parallel, netCDF-4 sequential/compressed, or pnetcdf for input/output, on a file-by-file basis. It also allows designation of an arbitrary number of I/O processors to handle all I/O. It will also allow use of new technologies being added to the netCDF C library, like zarr, and to the HDF5 library, including some other forms of cloud storage. Applications like FMS will not have to change their code to use these new features.

Right now I am finishing a PIO-netCDF integration project. PIO will become available to existing netCDF code (C or Fortran), via the use of a mode flag in nc_create/nc_open. (Some code changes are required to set up the I/O system, define data decomposition, and do reads/writes of the distributed arrays.) 

My plan is to submit this in as a pull request to FMS. I hope that it will allow users to easily try out a variety of different strategies, and use what works. They can easily switch from netcdf classic to pnetcdf to HDF5, to zarr - this could be a run-time decision by the user, or easily changed at compile-time by changing the mode flag in nf_create/nf_open. Furthermore, PIO allows codes to scaled to thousands of processors or more, while still providing the performance of parallel I/O.

marshallward:
Could I suggest that you integrate this into the [with-parallel-netcdf](https://github.com/NOAA-GFDL/FMS/tree/with-parallel-netcdf) branch?  It is working in the ocean cases that we tested, has a proven performance benefit at the ~1500 MPI rank/300 Lustre OST regime, has been written up in a manuscript, and has already been submitted as a PR.  Since (AFAIK) PIO is heavily based on parallel netCDF, it should share a lot of common code.

If it's possible, I'd also request that you retain this native parallel netCDF support alongside a potential PIO implementation.

edhartnett:
@marshallward I am going to examine your branch closely.

However, the way forward is not for me to integrate more code to an unmerged branch, it is to get your branch merged to master and then for me to work with that.

It is not my intention to remove any capability, but to add PIO as another option.

[...]

PR #453, opened 2020-06-11
Adds levels to mpp_init to return at a certain point

thomas-robinson:
**Description**
Adds a return to mpp_init so that different tests can use different amounts of mpp_init.

Fixes #454

**How Has This Been Tested?**
Tested with `make check` using intel20 and the travis_ci

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #478, opened 2020-06-23
simple test for mpp_domains

edwardhartnett:
**Description**
More simple testing. There is a bunch of mpp_domain tests, but they are all skipped. Well, that's a bummer because I am working on this code for PIO integration, and I need some tests to figure out what is going on.

I hope that someday you fix your broken tests. In the meantime, here's a simple test, which works now, and which I can add to as needed to confirm that I understand what mpp domains are all about.

Part of #244 and #100 

**How Has This Been Tested?**
Tested on my GNU/Linux box.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/mpp/Makefile.am
+++ b/test_fms/mpp/Makefile.am
@@ -37,7 +37,8 @@ check_PROGRAMS = test_mpp \
   test_mpp_print_memuse_stats_stderr \
   test_mpp_print_memuse_stats_file \
   test_mpp_memutils_begin_2x \
-  test_mpp_memutils_end_before_begin
+  test_mpp_memutils_end_before_begin \
+  test_domains_simple
 
 # These are the sources for the tests.
 test_mpp_SOURCES = test_mpp.F90
@@ -50,6 +51,7 @@ test_mpp_print_memuse_stats_stderr_SOURCES=test_mpp_print_memuse_stats_stderr.F9
 test_mpp_print_memuse_stats_file_SOURCES=test_mpp_print_memuse_stats_file.F90
 test_mpp_memutils_begin_2x_SOURCES=test_mpp_memutils_begin_2x.F90
 test_mpp_memutils_end_before_begin_SOURCES=test_mpp_memutils_end_before_begin.F90
+test_domains_simple_SOURCES = test_domains_simple.F90
 
 # Run the test programs.
 TESTS = test_mpp_domains2.sh \
--- a/test_fms/mpp/test_domains_simple.F90
+++ b/test_fms/mpp/test_domains_simple.F90
@@ -19,8 +19,8 @@
 
 !> @file
 
-!> This is a simple test of MPP layouts and a simple domain. In
-!> testing it should be run with 4 pes.
+!> This is a simple test of MPP layouts and a simple 1D and 2D
+!> domain. In testing it should be run with 4 pes.
 
 !> @author Ed Hartnett 6/22/20
 program test_domains_simple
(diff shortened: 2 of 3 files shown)


bensonr:
The mpp_domains tests aren't broken.  The issue is with the github CI area being resource limited for many of these tests.  Basic functionality requires a minimum of 6 mpi-ranks for the cubed-sphere related tests and to ensure domain creation and halo updates for the cubed-sphere and nests are unaffected requires more than six, ideally.     There are tests that can run on as few as 2 mpi-ranks and I agree these should be enabled.  I will work to ensure that happens.

edwardhartnett:
Is there a limit on the number of MPI ranks that Travis allows? I believe I use as many as 16 in some netCDF testing...

bensonr:
@edwardhartnett - I stand corrected.  I was able to run a travis test with 16 mpi-ranks as well.  Guess I was given bad info.  Regardless, I have most of the original mpp_domains tests running under travis and will soon have them in CI

edwardhartnett:
@thomas-robinson I have added comments both at the top and throughout the test.

Also I have added more test code, and am now checking the values of the created domain objects.

edwardhartnett:
This PR is ready to merge.

colingladueNOAA:
Hi @edhartnett ,
Could you please have a look at this PR ( https://github.com/NOAA-GFDL/FMS/pull/453 ) and let me know which "test_level" you need for your mpp_init call for this test. We are trying to minimize how much of mpp_init is used by these tests, so we'd like to only use mpp_init up to the point that is necessary.

edwardhartnett:
Test level 5.

Part of what I want from these tests is to have mpp behave the same way it does in an application. So I want it to do a full init.

I wonder why you are trying to limit mpp_init in this way? Testing right now is very quick, so why mess with it?

thomas-robinson:
@colingladueNOAA this PR needs to run all of mpp_init, so the test level will not apply to this test.

thomas-robinson:
@edhartnett the limits on using mpp_init are for testing the routines/functions that mpp_init calls.  The functions called in mpp_init all require that MPI in initialized and the mpp_initialized flag is set to .true.  Only an mpp function can set that flag, so we have to at least call that much of mpp_init. 

It seems silly to call all of mpp_init to test the routines that are in mpp_init.  mpp_pe(), for example, is used in mpp_init.  But if you call all of mpp_init, then you run mpp_pe before you test it. That's why we are limiting the  mpp_init calls for some tests.

This test should just run mpp_init.

edwardhartnett:
I do not think you should be so concerned about running mpp_pe before you run it (again) to test it. The test still has value if it's called twice. 

Since this is an existing fortran code base, written without unit testing, no doubt some oddities will result in testing. For example, I'm a bit at a loss to know how to test that mpp_init() succeeded. Since the initialized flag is private, I can't check it in my test code, and there does not seem to be a getter function for that, which is fine.

So for the moment I'm not testing much except that mpp_init() can run without calling mpp_error() or segfaulting. Which is not much of a test, but useful in the debugger, and a worthwhile platform for future testing. ;-)

You should be careful about adding testing overhead in your production code. Best to keep all your test-related code separate, so that it can't get tangled up in serious bugs, performance issues, or just cluttering a codebase that's already complex enough.

thomas-robinson:
@edhartnett thanks for the input.  We are trying to make as few modifications to the base code as possible related to testing.  The overhead is very minimal and only confined to mpp_init.  There were a few other ideas floated, and we ultimately selected this course of action to take moving forward for testing the functions contained in mpp_init.

PR #494, opened 2020-06-25
now require netCDF to have been built with HDF5

### LINKED ISSUE #466: Does FMS require netcdf to support HDF5? Or does FMS work with netCDF classic-only builds?
(Some background: netcdf supports multiple formats, including HDF5 and those formats we refer to as "classic formats." But it is possible to build netCDF in a way that HDF5 is not required. In this case, only classic formats can be read and written.)

OK, here's a question: does FMS always require netCDF to support netCDF/HDF5?

That is, is FMS intended to also work on classic-only netCDF builds?

If so, we must put a little magic in the configure.ac to detect whether netCDF supports netCDF/HDF, and then we must change tests to only attempt to use netCDF/HDF5 format when the underlying netcdf-c library supports it.

Alternatively, if you say that HDF5 is always to be supported, we need to add a check in configure.ac to ensure that is the case, and give the user a good error message at configure-time.

edwardhartnett:
**Description**
Check to ensure that netCDF was built with HDF5. This is required by the code, so checking in the configure will ensure we don't get link or runtime errors.

Fixes #466

**How Has This Been Tested?**
My GNU machine.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -96,6 +96,18 @@ fi
 AC_CHECK_HEADERS([netcdf.h], [], [AC_MSG_ERROR([Can't find the netCDF C header file.  Set CPPFLAGS/CFLAGS])])
 AC_SEARCH_LIBS([nc_create], [netcdf], [], [AC_MSG_ERROR([Can't find the netCDF C library.  Set LDFLAGS/LIBS])])
 
+# Require netCDF-4 (with HDF5).
+AC_MSG_CHECKING([if netCDF was built with HDF5])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [[
+#include <netcdf_meta.h>
+#if !(NC_HAS_NC4)
+      choke me
+#endif]])], [nc_has_nc4=yes], [nc_has_nc4=no])
+AC_MSG_RESULT([$nc_has_nc4])
+if test $nc_has_nc4 = no; then
+  AC_MSG_ERROR([NetCDF must be built with HDF5.])
+fi
+
 # Check if Linux gettid is avaiable
 AC_CHECK_FUNCS([gettid], [], [])
 


rem1776:
Tested on skylake with no issues

edwardhartnett:
OK, this is the next one to be merged. 

It's possible for netcdf to be built without HDF5 - we call this a "classic-only" build.

Not only in fms2_io, but also in other parts of your code, you depend on netCDF being built with HDF5. So this change to the configure script will save someone from a confusing debugging session, where they try and figure out why FMS is not working with a classic-only build of netCDF.

colingladueNOAA:
@edhartnett Can this be altered to check for Fortran as well as C?

edwardhartnett:
HDF5 fortran is not required. Only the HDF5 C library is used by netCDF, even netcdf-fortran.

That's because all calls to the F90 API go through the F77 API, and all F77 API calls go through the netcdf-c C library.

edwardhartnett:
@colingladueNOAA did I sufficiently address your question about Fortran?

Can we get this merged?

colingladueNOAA:
Yes, I just need to gather that last review.

edwardhartnett:
I'm going to suggest again that you lower your review requirements. Clearly it is holding back your progress.

Review is a great quality sieve, and I am a huge fan. But your entire quality process must be optimized to be running smoothly and increasing productivity. By setting the review bar too high, you are slowing development by holding back perfectly good PRs from merging. (And this current PR is certainly not the only example.)

I would suggest that, with the FMS at the present moment, testing is far more important than review. You are much more likely to find bugs with unit tests than you are with review. Only in the context of very good unit testing, does a more stringent review process make sense. (To find the bugs and problems that testing cannot.)

I suggest you restrict yourself to self-review, plus the review of one senior programmer who is an expert on that are of code. (Reviews by non-experts on the code have been found to be of little value in finding bugs.) If there are no experts on that area of code, then you need to appoint one and have them start coming up to speed on the code (perhaps by writing unit tests), so that you have a trusted expert in every area of the code base.

So a good way forward for FMS would be to lower your reviewer requirements to 1, and increase unit testing quite dramatically.

Here's a diagram from a paper about software engineering that is coming out as an NCAR tech note one of these fine days. Code review is just one pitstop on the racetrack. By taking too long there, you slow the entire process.

PIO Library Development Process

Issue #503, opened 2020-07-01
FMS Cmake infrastructure is not exporting Fortran compiled mdules


ngs333:
**Describe the bug**
FMS Cmake infrastructure is not exporting Fortran compiled module files into the install directory.

**To Reproduce**
Build libFMS and install as per the CMakeproject instructions. Inspect the install directories.

**Expected behavior**
All the compiled .mod files should appear somewhere in the install directotories, preferable in directory  /include/

**System Environment**
All environments that libFMS is build

- Configure options: Any additional flags, or macros passed to configure
Configuration as described in the project cmake instructions.

**Additional context**

ngs333:
It should be sufficient modify the CMakeLists.txt so as to set a Fortran module directory (e.g. set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/modules) ) and install it.

ngs333:
There is code that fixes he issue on branch  https://github.com/ngs333/FMS/tree/cmake_fmods. I am not considering a pull request (if any) until the resolution of issue 504 .

PR #504, opened 2020-07-06
bugfix on cmake in the master.

aerorahul:
**Description**

@underwoo @bensonr 
@DusanJovic-NOAA

The current CMake build in the master is broken and unusable.
This is a bugfix PR and fixes the cmake build, exports an FMS package that can be used in applications such as UFS and MOM6.
The UFS has cmake build.  FMS is used as a library inside the UFS (and MOM6).  This will allow UFS (and MOM6) developers to build and install FMS as a library and use it when building UFS (and MOM6), rather than building FMS each time the UFS (and MOM6) is compiled.

No additional dependency is added in this PR.

Fixes: #503 and more unreported issues with cmake building, packaging and installing.

**How Has This Been Tested?**
This branch has been tested with the UFS and MOM6
The tests with UFS resulted in bit-identical output when using FMS was built using GNUMake, in-line with UFS as well as externally (this PR) as a library. 
For MOM6, see associated MOM6 branch [here](https://github.com/aerorahul/MOM6/tree/feature/cmake-master)
To build MOM6 with CMake, see instructions [here](https://github.com/aerorahul/mom6-tools/tree/develop/MOM6-cmake)  

`CMAKE_INSTRUCTIONS.md` has been updated with simple build, install and use instructions.

`.travis.yml` is also updated to be able to build with Travis-CI.

**Checklist:**
- [x] Any dependent changes have been merged and published in downstream modules - There are none.
- [x] New check tests, if applicable, are included - Travis-CI now has a `build_with_autotools` and `build_with_cmake` build matrix.
- [x] make distcheck passes - Yes! and so does make install. May be add that to your checklist?


ngs333:
I do not see the cray pointer compiled option (-fcray-pointer for gnu)  being used. 
Is there a reason for this? My understanding is that without this option the code can compile but it can lead to crashes as it is possible to incorrectly pass the pointer size to a subroutine is the cray pointer is used as an calling argument. Also, arguments for line length are recommended for compiling FMS.

aerorahul:
> I do not see the cray pointer compile option (-fcray-pointer for gnu) being used.
> Is there a reason for this? My understanding is that without this option the code can compile but it can lead to crashes as it is possible to incorrectly pass the pointer size to a subroutine is the cray pointer is used as an calling argument. Also, arguments for line length are recommended for compiling FMS.

@ngs333 
Thank you for your review and comments.
I did not have a complete list of compiler flags and options for `GCC`.  `Intel` flags were derived from the `CMake_Instructions.md` as well as the need to have bit-2-bit reproducibility with the UFS. Thank you for suggesting these flags.  If there are other flags that are missing from the `GCC` (or `Intel` for that matter), I would gladly include them.

aerorahul:
@ngs333 
`-fcray-pointer` and `-ffree-line-length-none` compiler flags have been added for `GNU Fortran`.
Thanks!

ngs333:
Hello. I am still reviewing. Probably more info tomorrow. In the mean time here is an issue I noticed:
The code does not run on an our dfault test box that is using cmake version 3.10.1. The first o
/libnetcdff.so 
CMake Error at CMakeLists.txt:232 (target_link_libraries):
  Object library target "fms_r4_c" may not link to anything.

Here is a longer list of the output:
bash-4.2$ cmake -DCMAKE_BUILD_TYPE=Debug -D64BIT=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_INSTALL_PREFIX:PATH=/home/mzuniga/cmake_install ..
-- Found MPI_C: /opt/intel/2019_up5/impi/2019.5.281/intel64/bin/mpiicc (found version "3.1") 
-- Found MPI_Fortran: /opt/intel/2019_up5/impi/2019.5.281/intel64/bin/mpiifort (found version "3.1") 
-- Found MPI: TRUE (found version "3.1") found components:  C Fortran 
DEBUGNetCDF_C_INCLUDE_FILE: /opt/netcdf/4.6.1/INTEL/include/netcdf.h
DEBUGNetCDF_Fortran_INCLUDE_FILE: /opt/netcdf/4.6.1/INTEL/include/netcdf.mod
DEBUGNetCDF_C_CONFIG_EXECUTABLE: /opt/netcdf/4.6.1/INTEL/bin/nc-config
DEBUGNetCDF_Fortran_CONFIG_EXECUTABLE: /opt/netcdf/4.6.1/INTEL/bin/nf-config
DEBUGNetCDF_C_LIBRARY: /opt/netcdf/4.6.1/INTEL/lib64/libnetcdf.so
DEBUGNetCDF_Fortran_LIBRARY: /opt/netcdf/4.6.1/INTEL/lib64/libnetcdff.so
-- Found NetCDF: /opt/netcdf/4.6.1/INTEL/include (found version "4.6.1") found components:  C Fortran 
-- FindNetCDF defines targets:
--   - NetCDF_VERSION [4.6.1]
--   - NetCDF_PARALLEL [FALSE]
--   - NetCDF_C_CONFIG_EXECUTABLE [/opt/netcdf/4.6.1/INTEL/bin/nc-config]
--   - NetCDF::NetCDF_C [SHARED] [Root: /opt/netcdf/4.6.1/INTEL] Lib: /opt/netcdf/4.6.1/INTEL/lib64/libnetcdf.so 
--   - NetCDF_Fortran_CONFIG_EXECUTABLE [/opt/netcdf/4.6.1/INTEL/bin/nf-config]
--   - NetCDF::NetCDF_Fortran [SHARED] [Root: /opt/netcdf/4.6.1/INTEL] Lib: /opt/netcdf/4.6.1/INTEL/lib64/libnetcdff.so 
CMake Error at CMakeLists.txt:232 (target_link_libraries):
  Object library target "fms_r4_c" may not link to anything.

ngs333:
I noted that the proposed code does not use the existing FMS autotools infrastructure to generate any configuration information.  As per my original cmake instructions the input to the autotools are those files with the  “.in.in” extension. The intent of using autotools and those files is to leverage on the existing infrastructure which has been used for configuration for quite a while. This is a big issue because compilation configuration for FMS can be complex . Anyways, the proposed code has implications for a few related things I am still considering. The proposed code allows the Cmake system to automatically choose more of the default flags and it seems the flags are being set more in recommended Cmake style. Though this is a plus on one hand, it may not be possible to override the flags by setting environment variables (e.g FFLAGS) and Cmake code modification may be required.

aerorahul:
> I noted that the proposed code does not use the existing FMS autotools infrastructure to generate any configuration information. As per my original cmake instructions the input to the autotools are those files with the “.in.in” extension. The intent of using autotools and those files is to leverage on the existing infrastructure which has been used for configuration for quite a while. This is a big issue because compilation configuration for FMS can be complex . Anyways, the proposed code has implications for a few related things I am still considering. The proposed code allows the Cmake system to automatically choose more of the default flags and it seems the flags are being set more in recommended Cmake style. Though this is a plus on one hand, it may not be possible to override the flags by setting environment variables (e.g FFLAGS) and Cmake code modification may be required.

You are correct, that the proposed changes to the build framework does not use autotools.
I never understood the reasoning to have both inter-twined. Isn't autotools enough on itself? 
Why does FMS need a cmake configuration on top of autotools?

You can pass additional flags during compilation via:
```
-DCMAKE_Fortran_FLAGS=$FFLAGS
```
No code modification is required.

aerorahul:
> CMAKE_EXPORT_COMPILE_COMMANDS

I can't see the root cause in the log you posted.  There is no error on L232.

aerorahul:
> aerorahul
> I think there are two camps of users will likely remain for a while: Those who would want to use autotools generated config input into cmake , and those who do not? Do you think it will be possible (now or perhaps in the future if seems technically possible) to put in a flag that allows the user to decide between the two methods of operation?
> 
> By the way, the above is also related to additional choice of flags.For gnu, as you have it it seems cmake is automatically using these compile flags:
> -ffpe-trap=invalid,zero,overflow,underflow
> In general its a great decision for cmake to have done that - even as a default. The problem is that that is not usually how FMS is compiled and run by most customers, and in fact I am told the run results may be different with those flags vs without.

@ngs333 
I don't know anyone (users or developers) who would want to have autotools  generate config input into cmake, and then cmake generate Makefiles.  That is just plain redundant.  A pure cmake configuration, allows users to build FMS inline with their projects that use cmake. It also allows people who don't use cmake to link to the libraries as well as build FMS as an offline external library e.g. NetCDF.  Autotools on itself is capable of doing the latter.  By making autotools as a necessary step of the configuration building for cmake, one cannot execute option 1 of building FMS inline with the application that uses cmake.

If there is a demonstrable need to support both, please do add that option, while keeping the pure cmake option intact.  But in my opinion, that has redundancy written all over it, not to mention the added burden of maintenance.

As for your second comment, the flags you mention `-ffpe-trap=invalid,zero,overflow,underflow` are debugging flags as identified by `CMAKE_Fortran_FLAGS_DEBUG` and are only used when FMS is compiled in `Debug` mode.  It is **not** used when FMS is built in `Release` configuration.  CMake parses the flags based on the `CMAKE_BUILD_TYPE` and appends the appropriate list from Release or Debug (or Bit or MinRelDebInfo, etc).

thomas-robinson:
@aerorahul how do you suggest I test this?  Just type `cmake`?  Also, I would like to know what compilers you tested this with.  

Additionally, for contribution purposes, although the check boxes may not be applicable to a particular PR, we still require that they are all checked.  You can add additional comments before or after.

aerorahul:
> @aerorahul how do you suggest I test this? Just type `cmake`? Also, I would like to know what compilers you tested this with.
> 
> Additionally, for contribution purposes, although the check boxes may not be applicable to a particular PR, we still require that they are all checked. You can add additional comments before or after.

FMS requires a compiler, MPI and NetCDF and now CMake.

If you are on a system that supports modules; then do the following (the instructions are for Intel; for GNU replace appropriately):
```
module load intel
module load impi
module load netcdf
```

If the modules do not set the following variables or you are not using modules, these environment variables are necessary (Again, these are for Intel; for GNU replace appropriately)

```
export FC=mpiifort
export CC=mpiicc
export CXX=mpiicpc

export NetCDF_ROOT=`nc-config --prefix`
```

This assumes that NetCDF C and Fortran libraries are installed under `NetCDF_ROOT`. `NETCDF_ROOT` is also acceptable.  If NetCDF C and Fortran libraries are installed separately under their own individual directories, then it will be necessary to set the additional environment variables:

```
export NetCDF_C_ROOT=`nc-config --prefix`
export NetCDF_Fortran_ROOT=`nf-config --prefix`
```

You are now ready to configure and build FMS with CMake.

```
git clone -b bugfix/cmake https://github.com/aerorahul/fms
rm -rf build && mkdir -p build && cd build
cmake -DCMAKE_INSTALL_PREFIX=../FMS_INSTALL ../fms
make VERBOSE=Y -j6
make install
```

Following a successful build, FMS should be installed under `FMS_INSTALL`.  The installation tree should look like:
```
$> tree -L 4 FMS_INSTALL
FMS_INSTALL
├── include_r4
│   ├── amip_interp_mod.mod
│   ├── ...
│   └── xgrid_mod.mod
└── lib
    ├── cmake
    │   └── fms
    │       ├── fms-config-version.cmake
    │       ├── fms-config.cmake
    │       ├── fms-targets-release.cmake
    │       └── fms-targets.cmake
    └── libfms_r4.a
```

There are ofcourse many options for FMS during configuration of the build, such as:
```
-DOPENMP   "Build FMS with OpenMP support" DEFAULT: OFF
-D32BIT         "Build 32-bit (r4) FMS library" DEFAULT: ON
-D64BIT         "Build 64-bit (r8) FMS library" DEFAULT: OFF

-DINTERNAL_FILE_NML     "Enable compiler definition -DINTERNAL_FILE_NML"     DEFAULT: ON
-DENABLE_QUAD_PRECISION "Enable compiler definition -DENABLE_QUAD_PRECISION" DEFAULT: ON
-DGFS_PHYS              "Enable compiler definition -DGFS_PHYS"              DEFAULT:OFF
-DLARGEFILE             "Enable compiler definition -Duse_LARGEFILE"         DEFAULT:OFF
```

As mentioned in the PR description, this branch has been tested with the UFS and MOM6; although much more rigorously with UFS.
The tests with UFS resulted in bit-identical output when using FMS was built using GNUMake, in-line with UFS as well as externally (this PR) as a library.
The UFS tests were carried out on NOAA Tier-1 platforms of Orion, WCOSS Dell Phase 3 and Hera with Intel Compilers.
The UFS was also compiled (but no regression testing was done) with GNU compilers on macOS and this branch of FMS. These UFS tests use the `r4` library. 
Additionally, MOM6 was built and run with Intel and GNU compilers on similar NOAA platforms and macOS.
MOM6 uses the `r8` library from FMS. 

I will update the PR description with the additional checks information, but it here for brevity:
[x] Any dependent changes have been merged and published in downstream modules - There are none.
[x] New check tests, if applicable, are included - There were no checks in the old cmake build. None are added here as this is a bugfix PR for the broken cmake build.
[x] make distcheck passes - `make distcheck` is autotools lingo. `make install` installs the distribution. The package is relocatable i.e. the package installation directory can be moved anywhere on the system, and is guaranteed to work.

aerorahul:
@ngs333 [](url) @thomas-robinson 
If it wouldn't be too much of a burden, I would like to attach the following patch to the PR.
This patch will aid in requesting specific components of FMS `R4` and `R8`.  It also provides a more detailed information on what targets does the FMS package provide.  A snippet of configuration resulting from

```
find_package(FMS REQUIRED COMPONENTS R4 R8)
``` 

is shown below:

```
-- Found FMS: "/Users/demo/FMSwork/install" (Version: "2020.2")
-- find_package(FMS) defines targets:
--   - FMS::fms_r4 [/Users/demo/FMSwork/install/lib/libfms_r4.a]
--   - FMS::fms_r8 [/Users/demo/FMSwork/install/lib/libfms_r8.a]
```

It also checks for valid components requested from user, and provides a list of valid components if not.

[diffComponents.txt](https://github.com/NOAA-GFDL/FMS/files/4949967/diffComponents.txt)

.travis.yml
```diff
@@ -38,10 +38,6 @@ addons:
 
 env:
   - FCFLAGS_ADD='' DISTCHECK_CONFIGURE_FLAGS=''
-  - FCFLAGS_ADD='' DISTCHECK_CONFIGURE_FLAGS='--without-openmp'
-  - FCFLAGS_ADD='' DISTCHECK_CONFIGURE_FLAGS='--enable-mixed-mode'
-  - FCFLAGS_ADD='-fdefault-real-8 -fdefault-double-8 -fcray-pointer -ffree-line-length-none' DISTCHECK_CONFIGURE_FLAGS='--disable-setting-flags'
-  - FCFLAGS_ADD='' DISTCHECK_CONFIGURE_FLAGS='--enable-mpp-io'
```
thomas-robinson, line 44: Why were these tests removed?  They should be added back.

aerorahul: `master` `.travis.yml` is now restored.
This uses `autotools` and does not test `cmake` build.

aerorahul: @thomas-robinson I have restored `.travis.yml` to its `master` form, but that will break the build test.
I don't see the point in `autotools` generating `cmake` files, then `cmake` generating `Makefile` for build.
Adding just those flags means nothing to `cmake`.

aerorahul:
The build fails because @thomas-robinson asked me to restore `.travis.yml` to enable the `tests`.
Please advise on how to proceed.

thomas-robinson:
@aerorahul did you change the travis CI to test using cmake instead of autotools?  It looks like that's why the tests are failing.  If you want the CI to pass using cmake, then we need to figure out how to get the CI to pass using the 4 autotools tests and your cmake test.

aerorahul:
@thomas-robinson I think the first thing to do is to de-couple autotools from cmake.
autotools should not be pre-requisite for cmake.
Besides, what tests is autotools running? I couldn't get that to build and run.

thomas-robinson:
@aerorahul I'm sorry I'm not the cmake expert here. I thought this PR decoupled the autotools and cmake. I ran cmake as your instructions suggested and everything worked without running autotools. 

Autotools runs through a bunch of tests if you compile with make check or make distcheck.

aerorahul:
> @aerorahul I'm sorry I'm not the cmake expert here. I thought this PR decoupled the autotools and cmake. I ran cmake as your instructions suggested and everything worked without running autotools.
> 
> Autotools runs through a bunch of tests if you compile with make check or make distcheck.

@thomas-robinson 
I think I have it figured. I have de-coupled autotools from cmake. The autotools works as well as cmake.
Travis-ci now does 2 builds; first with autotools and next with cmake.
The autotools build does the test with `make -j distcheck`
The cmake does an clean install.

bensonr:
Thanks @aerorahul that's exactly what we want - a CI system that will test
both as independent entities.

thomas-robinson:
@aerorahul it looks like only the cmake is running on the travis CI still.  It passes because the cmake build is correct, but the autotools build is not being run at all.  If you look at the other PRs in this repo, you can see that there are 4 differentiationst builds that occur with the auto tools build.  In this PR, only one build is done, and it appears to only be the cmake build.  I'm not sure how to fix it yet.

aerorahul:
> @aerorahul it looks like only the cmake is running on the travis CI still. It passes because the cmake build is correct, but the autotools build is not being run at all. If you look at the other PRs in this repo, you can see that there are 4 differentiationst builds that occur with the auto tools build. In this PR, only one build is done, and it appears to only be the cmake build. I'm not sure how to fix it yet.

@thomas-robinson Yeah. I fixed it. There is now a build matrix `build_with_autotools` and `build_with_cmake`.
And they both pass!

thomas-robinson:
@aerorahul I submitted a [PR](https://github.com/aerorahul/FMS/pull/1) to your fork that will run all of the tests.  The matrix is a great idea, but there are 5 autotools tests and it was only testing one.

aerorahul:
> Has the cmake build system as configured been tested to ensure a modification in a .inc or .h in one of the subdirectories (MPP, fms2-io, etc) will rebuild the .F90s with dependencies and the library as a whole?

@bensonr No, I have not tried that. Trying now. Stay tuned.

aerorahul:
@bensonr 
I compiled a clean copy of FMS with cmake.
Then, I modified `mpp/include/mpp_comm.inc` by adding a comment..
Finally, I type make again, and indeed, cmake detected a change. See below:

```
Dependee "../mpp/include/mpp_comm.inc" is newer than depender "CMakeFiles/fms_r4_f.dir/mpp/mpp.F90.o".
Clearing dependencies in "/Users/rmahajan/scratch/FMSwork/fms-aerorahul/build/CMakeFiles/fms_r4_f.dir/depend.make".
[ 10%] Built target fms_r4_c
Scanning dependencies of target fms_r4_f
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f CMakeFiles/fms_r4_f.dir/build.make CMakeFiles/fms_r4_f.dir/build
[ 11%] Building Fortran object CMakeFiles/fms_r4_f.dir/mpp/mpp.F90.o
```

So it works as expected.  Feel free to re-test and satisfy yourself.

bensonr:
@aerorahul I'm good with your test result.  If it didn't work, I would have suggested a new issue and approved the merge anyway.

edwardhartnett:
Sorry to jump in so late, but I just noticed this PR. (Which I think is a good one and should be merged).

The way we do this on netcdf-c, netcdf-fortran, and pio, is to have a completely separate build for cmake and autotools. That is, cmake does not run and get the autotools build's config.h. Instead, we maintain a separate config.h.cmake.in, which is used in the cmake build.

This is a bit of a pain because when we add something to config.h, we need to also add it to config.h.cmake.in. And this is something we occasionally forget or get wrong. But once you get your builds well set up, it happens rarely. And you just get used to it. When submitting any change to the autotools build system of netcdf-c, I know I must also match that in the CMake build system. If I introduce a code file, pre-processor symbol, or test in one build, I must also add it to the other. 

So I would agree with @aerorahul that the CMake build should not depend on the autotools build. And once you add the file config.h.cmake.in, that pretty much solves the problem. See netcdf-c for a good example.

mlee03:
@aerorahul 

I followed the instructions in CMAKE_INSTRUCTIONS.md and was able to build FMS on Hera.
However, I am having trouble building FMS on Gaea and get the following error message: 

CMake Error at cmake/FindNetCDF.cmake:210 (message):
  Unable to properly find NetCDF.  Found static libraries at:
  /opt/cray/pe/netcdf/4.6.3.2/INTEL/19.0/lib/libnetcdff.a but could not run
  nc-config:
Call Stack (most recent call first):
  CMakeLists.txt:79 (find_package)

On Hera, I also tried linking the successfully built FMS library to an external program with "find_package(FMS)" and "target_link_libraries(appName FMS::fms_r4)", but I also had to provide the path to FindNetCDF.cmake with
-DCMAKE_MODULE_PATH.  Could direction #4 be updated with more detailed instructions?

bensonr:
This is a known problem and is a configuration issue with the
vendor-supplied NetCDF packages.

aerorahul:
> @aerorahul 
> 
> 
> 
> I followed the instructions in CMAKE_INSTRUCTIONS.md and was able to build FMS on Hera.
> 
> However, I am having trouble building FMS on Gaea and get the following error message: 
> 
> 
> 
> CMake Error at cmake/FindNetCDF.cmake:210 (message):
> 
>   Unable to properly find NetCDF.  Found static libraries at:
> 
>   /opt/cray/pe/netcdf/4.6.3.2/INTEL/19.0/lib/libnetcdff.a but could not run
> 
>   nc-config:
> 
> Call Stack (most recent call first):
> 
>   CMakeLists.txt:79 (find_package)
> 
> 
> 
> On Hera, I also tried linking the successfully built FMS library to an external program with "find_package(FMS)" and "target_link_libraries(appName FMS::fms_r4)", but I also had to provide the path to FindNetCDF.cmake with
> 
> -DCMAKE_MODULE_PATH.  Could direction number 4 be updated with more detailed instructions?

As @bensonr says, Gaea's NETCDF installation is not correct. 

Instruction #4 is to find the FMS library and use it.  
These are not the instructions for finding and using NetCDF. 
If a user has a different FindNETCDF.cmake, it should work just as well (assuming it is written correctly)

aerorahul:
> @aerorahul
> 
> I followed the instructions in CMAKE_INSTRUCTIONS.md and was able to build FMS on Hera.
> However, I am having trouble building FMS on Gaea and get the following error message:
> 
> CMake Error at cmake/FindNetCDF.cmake:210 (message):
> Unable to properly find NetCDF. Found static libraries at:
> /opt/cray/pe/netcdf/4.6.3.2/INTEL/19.0/lib/libnetcdff.a but could not run
> nc-config:
> Call Stack (most recent call first):
> CMakeLists.txt:79 (find_package)
> 
> On Hera, I also tried linking the successfully built FMS library to an external program with "find_package(FMS)" and "target_link_libraries(appName FMS::fms_r4)", but I also had to provide the path to FindNetCDF.cmake with
> -DCMAKE_MODULE_PATH. Could direction number 4 be updated with more detailed instructions?

@mlee03 Added a blurb on instructions.
This dismissed reviews  that were already done.
@bensonr @ngs333 @thomas-robinson @wrongkindofdoctor  Please re-review.

PR #510, opened 2020-07-10
Adds test for mpp_init_logfile 

ngs333:
**Description**
This PR creates a unit test for the mpp interface mpp_init_logfile()

Fixes # (issue)

**How Has This Been Tested?**
On both the skylake server with intel compiler, and a home computer with gcc compiler, 
libFMS and the unit test applications were compiled. The unite tests for subdirectory FMS/est_fms/mpp were run via "make check" from the command line. Additionally, the unit test brought in with this PR was made made to run as expected under various conditions and the side effects on the files were verified to be as expected.

Note and warning: Interface mpp_init_logfile() did not have any documentation. I made an interpretation on how its expected to run and commented the interface to contain that documentation. In order to test from an external program, I made the interface public. Let me know if this is an issue.  

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -149,6 +149,21 @@
   end function stdlog
 
   !#####################################################################
+  ! <FUNCTION NAME="mpp_init_logfile">
+  !  <OVERVIEW>
+  !    Initialize the logfile.
+  !  </OVERVIEW>
+  !  <DESCRIPTION>
+  !    This function re-initializes the log files of the PEs. Re-initialization
+  !    means replacing if it already exists. It does not make new files if they don't
+  !    already exist. Only log files with names of format "<header>.pe_number.out",
+  !    where pe_number is the digit integer represeting the PE number, such as 000000 for the
+  !    root PE, 000001 for the next PE, etch. The <header> is the config file name prefix -
+  !    a module global normally set elsewhere, and frequetly is the integer simulation
+  !    timedate (e.g. 19480101).
+  !  </DESCRIPTION>
+  ! </FUNCTION>
+
   subroutine mpp_init_logfile()
   integer :: p
   logical :: exist
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -148,22 +148,14 @@
 11  call mpp_error( FATAL, 'STDLOG: unable to open '//trim(etcfile)//'.' )
   end function stdlog
 
-  !#####################################################################
-  ! <FUNCTION NAME="mpp_init_logfile">
-  !  <OVERVIEW>
-  !    Initialize the logfile.
-  !  </OVERVIEW>
-  !  <DESCRIPTION>
-  !    This function re-initializes the log files of the PEs. Re-initialization
-  !    means replacing if it already exists. It does not make new files if they don't
-  !    already exist. Only log files with names of format "<header>.pe_number.out",
-  !    where pe_number is the digit integer represeting the PE number, such as 000000 for the
-  !    root PE, 000001 for the next PE, etch. The <header> is the config file name prefix -
-  !    a module global normally set elsewhere, and frequetly is the integer simulation
-  !    timedate (e.g. 19480101).
-  !  </DESCRIPTION>
-  ! </FUNCTION>
-
+  
+  !> \brief Subroutine mpp_init_logfile re-initializes the log files of the PEs.
+  !! Re-initialization means replacing if it already exists. It does not make new
+  !! files if they don't already exist. It only re-initializes log files with names of
+  !! format "<header>.pe_number.out", where pe_number is the digit integer represeting
+  !! the PE number, such as 000000 for the root PE, 000001 for the next PE,
+  !! etc. The <header> is the config file name prefix - a module global set
+  !! elsewhere, and frequetly is the integer simulation timedate (e.g. 19480101).
   subroutine mpp_init_logfile()
   integer :: p
   logical :: exist
--- a/mpp/mpp.F90
+++ b/mpp/mpp.F90
@@ -230,9 +230,7 @@ private
   public :: read_ascii_file, read_input_nml, mpp_clock_begin, mpp_clock_end
   public :: get_ascii_file_num_lines
   public :: mpp_record_time_start, mpp_record_time_end
-  !! TODO:
   public :: mpp_init_logfile
-
 
   !--- public interface from mpp_comm.h ------------------------------
   public :: mpp_chksum, mpp_max, mpp_min, mpp_sum, mpp_transmit, mpp_send, mpp_recv
--- a/mpp/mpp.F90
+++ b/mpp/mpp.F90
@@ -230,6 +230,9 @@ private
   public :: read_ascii_file, read_input_nml, mpp_clock_begin, mpp_clock_end
   public :: get_ascii_file_num_lines
   public :: mpp_record_time_start, mpp_record_time_end
+  !! TODO:
+  public :: mpp_init_logfile
+
 
   !--- public interface from mpp_comm.h ------------------------------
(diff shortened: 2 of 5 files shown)


mpp/include/mpp_util.inc
```diff
@@ -149,6 +149,21 @@
...
+  !    This function re-initializes the log files of the PEs. Re-initialization
+  !    means replacing if it already exists. It does not make new files if they don't
+  !    already exist. Only log files with names of format "<header>.pe_number.out",
+  !    where pe_number is the digit integer represeting the PE number, such as 000000 for the
+  !    root PE, 000001 for the next PE, etch. The <header> is the config file name prefix -
+  !    a module global normally set elsewhere, and frequetly is the integer simulation
+  !    timedate (e.g. 19480101).
+  !  </DESCRIPTION>
+  ! </FUNCTION>
+
```
thomas-robinson, line 166: Any added documentation should be in doxygen format.  Please reformat

test_fms/mpp/test_mpp_init_logfile.F90
```diff
@@ -0,0 +1,44 @@
...
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+!***********************************************************************
+
+!> @file
+!! @brief Tests the mpp_init_logfile interface
+!! @author Miguel Zuniga
+!! @email gfdl.climate.model.info@noaa.gov
+
+program test_mpp_init_logfile
+#include <fms_platform.h>
```
thomas-robinson, line 26: You should use platforms_mod instead

mpp/mpp.F90
```diff
@@ -230,6 +230,9 @@ module mpp_mod
   public :: read_ascii_file, read_input_nml, mpp_clock_begin, mpp_clock_end
   public :: get_ascii_file_num_lines
   public :: mpp_record_time_start, mpp_record_time_end
+  !! TODO:
+  public :: mpp_init_logfile
+
```
thomas-robinson, line 235: Remove this whitespace

mpp/mpp.F90
```diff
@@ -230,6 +230,9 @@ module mpp_mod
   public :: read_ascii_file, read_input_nml, mpp_clock_begin, mpp_clock_end
   public :: get_ascii_file_num_lines
   public :: mpp_record_time_start, mpp_record_time_end
+  !! TODO:
+  public :: mpp_init_logfile
```
thomas-robinson, line 234: Are you still working on this?  Are you making a private routine public?

test_fms/mpp/test_mpp_init_logfile.F90
```diff
@@ -0,0 +1,44 @@
...
+  use mpp_mod, only : mpp_init, mpp_init_logfile
+  use mpp_mod, only : mpp_init_test_logfile_init
+
+  IMPLICIT NONE
+
+  integer :: my_mpp_init_level
+  integer :: err_no
+
+  !Initialize mpp but have mpp_init() exit before log files are initalized.
+  my_mpp_init_level = mpp_init_test_logfile_init - 1
```
thomas-robinson, line 37: Why are you subtracting 1?  Isn't there a `mpp_init_test` level that you can directly use?

test_fms/mpp/test_mpp_init_logfile.F90
```diff
@@ -0,0 +1,44 @@
...
+  call mpp_init( test_level = my_mpp_init_level )
+  
+  call mpp_init_logfile()
```
thomas-robinson, line 40: is anything actually tested, or is this routine just called?

mpp/mpp.F90
ngs333: I was hoping to not make it public. Is there a better way to get access to mpp_init_logfile from the unit test app.

test_fms/mpp/test_mpp_init_logfile.F90
ngs333: Most of the test work is actually done is done by the shell command calling  this the fortran test program. Like many I/o routines, mpp_init_logfile() is merely a side_effect routine. So what the shell comman does is test that it makes it side effects correctly and also that it doesn't make additional side effects that it could potentially make. This means, checking the files that it should replace, and checking for the lack of existence of files that it should not make but that whose names are iterated over in the do loop of the routine. If you look at the shell command that calls it, you should be able to see that it return a '1' when those conditions are not satisfied. (PS I checked that the CI summary shows "FAILED" is the shell command returns 1, and success if it returns 0.)

test_fms/mpp/test_mpp_init_logfile.F90
ngs333: I was subtracting one so that initialization stops before calling init_logfile. But given the 
challenge of keeping init_logfile private, and upon inspection of the mpp_init routine, it occurs to me that I can solve both issues just by not subtracting one and not calling init_log_file.
Thanks.

ngs333: I am not sure and am requesting input. If the init level is set to "mpp_init_test_logfile_init".
then mpp_init() actually does more work AFTER calling mpp_initlogfile: it writes to the logifle. So technically, to test the interface, one should isolate the interface to whatever extent possible, not after what it creates is used by another function.

mpp/mpp.F90
ngs333: This may remain a problem (with both pros and cons) see the issue below concerning  mpp_init_test_logfile_init

test_fms/mpp/test_mpp_init_logfile.F90
ngs333: Thats what I get for copy and pasting. I dont even use it.

test_fms/mpp/test_mpp_init_logfile.F90
ngs333: I updated the comments in both the Fortran test program and the corresponding  shell script. It better explains what they are doing together.

test_fms/mpp/test_mpp_init_logfile.F90
```diff
@@ -0,0 +1,46 @@
...
+  use mpp_mod, only : mpp_init, mpp_init_logfile
+  use mpp_mod, only : mpp_init_test_logfile_init
+
+  IMPLICIT NONE
+
+  integer :: err_no
+
+  !Initialize mpp but have mpp_init() exit right after log files are initalized.
+ 
+  call mpp_init( test_level =  mpp_init_test_logfile_init -1 )
```
thomas-robinson, line 39: The problem that I have with this is that you can't guarantee that it will work in the future.  If they decide to change the test levels to be multiples of 100, switch to reals, etc, then this code will no longer work.  You should just use mpp_test_level instead of subtracting 1

ngs333: I changed the code to use mpp_init_test_logfile_init instead of "mpp_init_test_logfile_init -1"

mpp/include/mpp_util.inc
ngs333: I think I have reformatted now!

mpp/mpp.F90
ngs333: I think its removed now!

mpp/mpp.F90
ngs333: Thanks for bringing up the issue in the meeting. Fortunately for this test is a non-issue now.

test_fms/mpp/test_mpp_init_logfile.F90
ngs333: Its changed so it not subtracting now. Corresponding changed in the shell script were made. Mpp_init_logfile was removed from public (i.e. put back to original form).

rem1776:
test included in #800

Issue #505, opened 2020-07-09
FMS modules compatibility with mpp_io and fms2_io


thomas-robinson:
**Is your feature request related to a problem? Please describe.**
There are problems with the implementation of fms2_io, so the diag_manager, xgrid, interpolator, and data_override need to work with both fms2_io and mpp_io.

**Describe the solution you'd like**
Add a compile-time switch for using fms2_io or mpp_io.  Update the auto tools to handle this flag.

**Describe alternatives you've considered**
There could be a namelist option to pick between the two IO options, but this would require more complicated coding.

**Additional context**
This is needed to work with models transitioning from the legacy IO

thomas-robinson:
@GFDL-Eric @uramirez8707 @rem1776 I updated this issue so we can all use it.  We don't need 5-6 different issues for the same thing.

PR #509, opened 2020-07-10
Diag manager combining 2020 and 2019 with -Duse_mpp_io flag

thomas-robinson:
**Description**
The diag_manager code from 2019.01.02 is added here and controlled with the compile-time flag -Duse_mpp_io.

Fixes #505 
Replaces #506

**How Has This Been Tested?**
`make check` on skylake with 
```bash
../configure FCFLAGS="-fno-alias -auto -safe-cray-ptr -ftz -assume byterecl -i4 -r8 -nowarn -sox -traceback -qopenmp `nf-config --fflags` -O0 " CPPFLAGS="`nc-config --cflags` -I/opt/intel/2019_up2/impi/2019.2.187/intel64/include/" FC="mpiifort" CC="mpiicc"
```
and 
```bash
../configure FCFLAGS="-fno-alias -auto -safe-cray-ptr -ftz -assume byterecl -i4 -r8 -nowarn -sox -traceback -qopenmp `nf-config --fflags` -O0 " CPPFLAGS="-Duse_mpp_io `nc-config --cflags` -I/opt/intel/2019_up2/impi/2019.2.187/intel64/include/" FC="mpiifort" CC="mpiicc"
```
Also tested with intel18 and intel19 on gaea using the experiment c96L33_am4p0_cmip6Diag

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #538, opened 2020-07-22
Add ifdef for use_mpp_io to diag manager modules

thomas-robinson:
**Description**
Uses and ifdef for `Duse_mpp_io` to enable the compile time option to use mpp_io with the diag_manager output.

Fixes #505 
Replaces #509 

**How Has This Been Tested?**
Tested on skylake machine with intel19.
Run with c96L33_am4p0_cmip6Diag both with and without `Duse_mpp_io`

**Checklist:**
- [x] `make distcheck` passes


diag_manager/diag_manager.F90
```diff
@@ -3945,9 +3945,11 @@ SUBROUTINE diag_manager_init(diag_model_subset, time_init, err_msg)
       ALLOCATE(input_fields(j)%output_fields(MAX_OUT_PER_IN_FIELD))
     END DO
     ALLOCATE(files(max_files))
+#ifndef use_mpp_io
```
underwoo, line 3948: Are the `DEALLOCATE` statements somewhere for these items?

diag_manager/diag_output.F90
```diff
@@ -1388,4 +1389,957 @@ SUBROUTINE set_diag_global_att(component, gridType, tileName)
   END SUBROUTINE set_diag_global_att
   ! </SUBROUTINE>
 
+!END MODULE diag_output_mod
+#else
```
underwoo, line 1393: Please combine the two parts as you have done in `diag_manager.F90`

diag_manager/diag_util.F90
```diff
@@ -18,6 +18,7 @@
 !***********************************************************************
 
 MODULE diag_util_mod
+#ifndef use_mpp_io
```
underwoo, line 21: Please combine the two as was done in `diag_manager.F90`.

diag_manager/diag_output.F90
thomas-robinson: The routines in this file are so vastly different, that I think it makes the most sense to keep them separate like this.  In the next FMS release, we are working on renaming the routines that use mpp_io so that they will called differently.  If something needs to be patched, it's likely to only need to be patched in one place in this module.

diag_manager/diag_util.F90
thomas-robinson: This is done with [df04785](https://github.com/NOAA-GFDL/FMS/pull/538/commits/df04785ce6907df1c769a37a5ed03113b55cdf65)

diag_manager/diag_manager.F90
thomas-robinson: `output_fields`, `input_fields`, `files`, `fileobjU`, `fileobj`, `fileobjND`, and `fnum_for_domain` are not deallocated anywhere that I can see.

underwoo: They really should be deallocated somewhere.

diag_manager/diag_output.F90
underwoo: If that is the plan, then I suggest placing the `#if`/`#endif` around the routines, modifying each just enough so both can use the same file header stuff (e.g. `use`, `#include`, etc.).

I also noticed in the code places where instead of deleting lines that are not needed, they are commented out (but no comment as to why they are kept).  I also have seen things like the following:
```fortran
IF ( <some_test> ) THEN
  ! Do nothing
ELSE
  STATEMENT1
  STATEMENT2
END IF
```

Using an `IF` like this causing issues later, as it is difficult to understand _why_ the `IF` tests for something that has no work.  These cases should be negated:

```fortran
IF ( .NOT.<some_test> ) THEN
  STATEMENT1
  STATEMENT2
END IF.
```

Easier to understand and maintain.

diag_manager/diag_manager.F90
thomas-robinson: I'm hesitant to do that because I don't know if anything will break.

thomas-robinson: I added deallocate statements for the fms2_io variables that are in here

diag_manager/diag_output.F90
thomas-robinson: The two different code bases use different `#include`s, so I think this is actually messier now.  I did it though. 
 
I think the clean up (ifs, commented code) could/should be done on a different PR since it is a separate issue.

thomas-robinson:
@underwoo I think I addressed all of your concerns.  I opened up a new issue for the code clean up.  I `#ifdef`ed the routines and includes in diag_output.F90.  I deallocated the fms2_io variables in diag_manager_end.

PR #541, opened 2020-07-23
Fix call for mpp-io flag

rem1776:
**Description**
Improves PR #526 by calling AC_DEFINE to set the flag instead of adding the flag through a hard coded string

Fixes #505 

**How Has This Been Tested?**
Tested with travis CI and on skylake with intel compiler

**Checklist:**
- [x] `make distcheck` passes


thomas-robinson:
@edwardhartnett can you take a look at this?  It's a response to your review of #526

edwardhartnett:
Looks good! Using AC_DEFINE causes the flag to be set, along with all the others, in config.h.

Not sure what the state of your cmake build is, but any change you make in autotools must also be reflected in the cmake build. So generally, when I see an AC_DEFINE in configure.ac, I wonder what is supposed to happen in CMake?

ALso does your cmake build have a config.h.cmake.in file? If so, you will also want to regenerate that each time you add an AC_DEFINE.

rem1776:
@edwardhartnett I believe they're still working on the cmake build and getting those mpp-io changes added as well

PR #556, opened 2020-08-12
mpp: BOZ literal assignment

### LINKED ISSUE #555: BOZ literal error in mpp
**Describe the bug**
BOZ literals (e.g. Z'00001000') in the mpp module fail when transferred to integers when compiled under GCC10.  This is because BOZ literals formally have no type and cannot be safely transferred to a typed variable without recasting.

This could be resolved, for example, by using `int(..., kind=...)` to transfer the value.

**To Reproduce**
Standard ./configure && make build using CC=gcc-10

**Expected behavior**
Model should build without error.

**System Environment**
- OS:
```
$ uname -a
Linux guadosalam 5.7.12-arch1-1 #1 SMP PREEMPT Fri, 31 Jul 2020 17:38:22 +0000 x86_64 GNU/Linux
```
- Compiler(s): Type and version [e.g. Intel 19.1]
```
$ gfortran -v
Using built-in specs.
COLLECT_GCC=gfortran
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/10.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,d --with-isl --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-install-libiberty --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-libunwind-exceptions --disable-werror gdc_include_dir=/usr/include/dlang/gdc
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 10.1.0 (GCC)
```
- MPI type, and version (e.g. MPICH, Cray MPI, openMPI)
```
$ ompi_info --version
Open MPI v4.0.4
```
- netCDF Version: For both C and Fortran
```
$ nc-config --version
netCDF 4.7.4
```
- Configure options: Any additional flags, or macros passed to configure
```
export CC=mpicc
export FC=mpifort
export FCFLAGS="-fdefault-real-8 -fdefault-double-8 -fcray-pointer"
```
**Additional context**
I have a PR prepared to resolve this issue.

marshallward:
**Description**
Because BOZ literals (e.g. Z'00001000') are typeless, the standard does
not allow them to be assigned to variables, which do have prescibed
types.

While this has generally not been a problem in the past, GCC 10 now
forbids such assignments, which raises errors in the FMS build.

This patch resolves the issue by replacing assignments with an
`int(..,kind=...)` transfer.

Although this can be avoided with the `-fallow-invalid-boz` flag, the
following patch should avoid the need for any new flags.

Fixes #555 

**How Has This Been Tested?**
This was tested in the MOM6 test suite.  Only the OpenMP build could be tested, due to newer changes in FMS which seem to require OpenMP.

**Checklist:**
- [x] `make distcheck` passes

I apologize but I am currently unable to get the `make distcheck` tests to pass.  The problems appear to be unrelated to the code change.  Hopefully someone on the FMS side can test this.  If the problem persists then I will look back into it.

### DIFF
--- a/mpp/mpp_domains.F90
+++ b/mpp/mpp_domains.F90
@@ -646,7 +646,7 @@ module mpp_domains_mod
   integer,                                save :: a_sort_len=0        ! len sorted memory list
   integer,                                save :: n_addrs=0           ! num memory addresses used
 
-  integer(LONG_KIND), parameter :: ADDR2_BASE=Z'0000000000010000'
+  integer(LONG_KIND), parameter :: ADDR2_BASE = int(Z'0000000000010000', kind=LONG_KIND)
   integer, parameter :: MAX_ADDRS2=128
   integer(LONG_KIND),dimension(MAX_ADDRS2),save :: addrs2_sorted=-9999  ! list of sorted local addrs
   integer,           dimension(-1:MAX_ADDRS2),save :: addrs2_idx=-9999  ! idx of addr2 assoicated w/ d_comm
@@ -671,10 +671,10 @@ module mpp_domains_mod
   integer,                                         save           :: n_comm=0            ! num communicators used
 
   !     integer(LONG_KIND), parameter :: GT_BASE=2**8
-  integer(LONG_KIND), parameter :: GT_BASE=Z'0000000000000100'  ! Workaround for 64bit int init problem
+  integer(LONG_KIND), parameter :: GT_BASE = int(Z'0000000000000100', kind=LONG_KIND)
 
   !     integer(LONG_KIND), parameter :: KE_BASE=2**48
-  integer(LONG_KIND), parameter :: KE_BASE=Z'0001000000000000'  ! Workaround for 64bit int init problem
+  integer(LONG_KIND), parameter :: KE_BASE = int(Z'0001000000000000', kind=LONG_KIND)
 
   integer(LONG_KIND) :: domain_cnt=0
 
--- a/mpp/mpp_parameter.F90
+++ b/mpp/mpp_parameter.F90
@@ -117,7 +117,7 @@ module mpp_parameter_mod
 ! combination with the flag parameter defined above to create a unique identifier for
 ! each Domain+flags combination. Therefore, the value of any flag must not exceed DOMAIN_ID_BASE.
 ! integer(LONG_KIND), parameter :: DOMAIN_ID_BASE=INT( 2**(4*LONG_KIND),KIND=LONG_KIND )
-  integer(LONG_KIND), parameter :: DOMAIN_ID_BASE=Z'0000000100000000' ! Workaround for 64bit init problem
+  integer(LONG_KIND), parameter :: DOMAIN_ID_BASE = int(Z'0000000100000000', kind=LONG_KIND)
   integer, parameter :: NON_BITWISE_EXACT_SUM=0
   integer, parameter :: BITWISE_EXACT_SUM=1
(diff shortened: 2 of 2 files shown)


thomas-robinson:
@marshallward did you test this with any other compilers other than gcc/gfortran?

thomas-robinson:
@marshallward the CI passed, and that runs `make distcheck` with a few different options.  Maybe you are missing some of the flags that you need to compile with `make distcheck`.  What was the specific failure that you saw?

marshallward:
Unfortunately I lost a lot of the error logs, but I can try to reproduce what I saw.

On my own machine with gcc10, `make distcheck` seemed to build successfully but failed on a few of the tests.  The errors were in netCDF and were related to `_FillValue`.  On Gaea with Intel 19, I could not get it to build, but the error was a bit too deep into other Makefiles or scripts and I could not work out the actual problem.

I will try to come back to it and see if I can replicate these errors.

marshallward:
So `CC=cc FC=ftn ./configure && make -j` works for me with Intel 19.0.5, but I cannot subsequently run `make distcheck` unless I again set `CC` and `FC`.

If I do `CC=cc FC=ftn make -j distcheck` I get this error:
```
/bin/sh ../../libtool  --tag=FC   --mode=link ftn  -g -FR -real_size 64 -fopenmp   -o test_diag_manager test_diag_manager.o ../../libFMS/libFMS.la 
libtool: link: ftn -g -FR -real_size 64 -fopenmp -o .libs/test_diag_manager test_diag_manager.o  ../../libFMS/.libs/libFMS.so -fopenmp -Wl,-rpath -Wl,/lustre/f2/dev/Marshall.Ward/dev/FMS/FMS-2020.02-dev/_inst/lib
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: attempted static link of dynamic object `../../libFMS/.libs/libFMS.so'
```
It looks like I somehow built a dynamic library when the build was expecting a static library.

Admittedly I did not give much thought to the flags, so maybe I was supposed to do something different here.

wrongkindofdoctor:
@marshallward On Gaea, you need to run `./configure --disable-shared` after setting your environment variables.

marshallward:
@wrongkindofdoctor I tried adding the flag but don't seem to be getting any success.  This is what I just tried:
```
CC=cc FC=ftn ./configure --disable-shared && make -j distcheck
```
This failed because `make` said CC was not set, so I tried this instead:
```
$ CC=cc FC=ftn ./configure --disable-shared && CC=cc FC=ftn make -j distcheck
```
which does start to run but then gives me the static library error.

I don't really understand how I'm supposed to use `make distcheck`.  But if you all would rather discuss this outside the ticket, or just drop the issue, then either is fine with me.

rem1776:
Tested successfully with gnu on Orion

PR #571, opened 2020-09-02
Topography IO namelist flag

### LINKED ISSUE #569: Namelist switch to select IO version for use in libFMS modules
fms2_io and mpp_io both exist in the FMS repository.  There is no way to test each version of IO side-by-side in the different modules.

Add a namelist option to the different modules in libFMS
- [x] diag_manager @thomas-robinson 
- [x] xgrid @thomas-robinson 
- [x] data_override @GFDL-Eric 
- [x] interpolator @uramirez8707 
- [x] amip_interp @uramirez8707 
- [x] topography @rem1776 
- [x] travis yaml and autotools build @rem1776 

The current alternative is a compile time switch which is an all-or-nothing solution

rem1776:
**Description**
Takes out #ifdefs from topography.F90 and adds in runtime conditionals to use mpp_io if a namelist flag is set

Fixes #569 

**How Has This Been Tested?**
Tested with make check on skylake with intel 20

**Checklist:**
- [x] `make distcheck` passes


uramirez8707 (changes requested):
@rem1776 You need to add `use_mpp_io` to 
https://github.com/NOAA-GFDL/FMS/blob/565678f21f6e1e8c1b7560a5652665abb494e652/topography/topography.F90#L104

to define it as a namelist parameter.

topography/topography.F90
```diff
@@ -198,14 +205,20 @@ end subroutine topography_init
...
+
+ function get_topog_mean_1d (blon, blat, zmean, use_mpp_io_arg)
 
    real, intent(in),  dimension(:)   :: blon, blat
    real, intent(out), dimension(:,:) :: zmean
    logical :: get_topog_mean_1d
+   logical,optional :: use_mpp_io_arg
 
 !-----------------------------------------------------------------------
+   if (use_mpp_io .OR. present(use_mpp_io_arg) .AND. use_mpp_io_arg) then
```
thomas-robinson, line 217: I think you are missing a set of parentheses
```fortran
if (use_mpp_io .OR. (present(use_mpp_io_arg) .AND. use_mpp_io_arg)) then
```
I think stylistically matching how other modules work would be:
```fortran
   logical,optional :: use_mpp_io_arg
   logical :: mpp_io_use
   if (present(use_mpp_io_arg) then
     mpp_io_use = use_mpp_io_arg
   else
     mpp_io_use = use_mpp_io
  endif
!-----------------------------------------------------------------------
￼   if (mpp_io_use) then
```

topography/topography.F90
```diff
@@ -225,13 +238,19 @@ end function get_topog_mean_1d
...
- function get_topog_mean_2d (blon, blat, zmean)
+ function get_topog_mean_2d (blon, blat, zmean, use_mpp_io_arg)
 
    real, intent(in),  dimension(:,:) :: blon, blat
    real, intent(out), dimension(:,:) :: zmean
    logical :: get_topog_mean_2d
+   logical,optional :: use_mpp_io_arg
 
 !-----------------------------------------------------------------------
+   if ( use_mpp_io .OR. present(use_mpp_io_arg) .AND. use_mpp_io_arg) then
```
thomas-robinson, line 249: Same issue as above

topography/topography.F90
```diff
@@ -225,13 +238,19 @@ end function get_topog_mean_1d
...
+       get_topog_mean_2d = get_topog_mean_2d_mpp_io (blon, blat, zmean)
+       return
+   endif !< use_mpp_io
```
thomas-robinson, line 252: This isn't done for fms2_io?

topography/topography.F90
```diff
@@ -284,13 +303,19 @@ end function get_topog_mean_2d
...
- function get_topog_stdev_1d (blon, blat, stdev)
+ function get_topog_stdev_1d (blon, blat, stdev, use_mpp_io_arg)
 
    real, intent(in),  dimension(:)   :: blon, blat
    real, intent(out), dimension(:,:) :: stdev
    logical :: get_topog_stdev_1d
+   logical,optional :: use_mpp_io_arg
 
 !-----------------------------------------------------------------------
+   if ( use_mpp_io .OR. present(use_mpp_io_arg) .AND. use_mpp_io_arg) then
```
thomas-robinson, line 314: Same as above

topography/topography.F90
```diff
@@ -284,13 +303,19 @@ end function get_topog_mean_2d
...
+       get_topog_stdev_1d = get_topog_stdev_1d_mpp_io (blon, blat, stdev)
+       return
+   endif !< use_mpp_io
```
thomas-robinson, line 317: Should this just be an `else`?

topography/topography.F90
```diff
@@ -310,13 +335,19 @@ end function get_topog_stdev_1d
...
    real, intent(in),  dimension(:,:) :: blon, blat
    real, intent(out), dimension(:,:) :: stdev
    logical :: get_topog_stdev_2d
+   logical,optional :: use_mpp_io_arg
 
 !-----------------------------------------------------------------------
+   if ( use_mpp_io .OR. present(use_mpp_io_arg) .AND. use_mpp_io_arg) then
+       get_topog_stdev_2d = get_topog_stdev_2d_mpp_io (blon, blat, stdev)
+       return
+   endif !< use_mpp_io
```
thomas-robinson, line 349: Same as previous comments

topography/topography.F90
```diff
@@ -1413,9 +1477,8 @@ subroutine interp_water_1d ( blon, blat, zout, do_ocean )
  real   , intent(in)  :: blon(:), blat(:)
  real   , intent(out) :: zout(:,:)
  logical, intent(in), optional :: do_ocean
-
  real :: xdat(ipts+1), ydat(jpts+1), zdat(ipts,jpts)
-
+
```
thomas-robinson, line 1481: remove this space

topography/topography.F90
uramirez8707: why is `use_mpp_io_arg` even needed? 

`get_topog_mean` is called outside of this subroutine, so `use_mpp_io_arg` is never going to be set anyway (unless we add it) ... and because the if statements are before the `topography_init` call (where use_mpp_io is read from the namelist) fms_io will always be used (at least in the first call)

I think this subroutines needs to be something like: 
(you can probably clean it up farther)
```F90
   if (.not. module_is_initialized) call topography_init()

   if ( any(shape(zmean(:,:)) /= (/size(blon(:))-1,size(blat(:))-1/)) ) &
        call error_mesg('get_topog_mean_1d','shape(zmean) is not&
            & equal to (/size(blon)-1,size(blat)-1/))', FATAL)

 if (use_mpp_io) then
   if ( open_topog_file_mpp_io(topog_file) ) then
       call interp_topog_1d_mpp_io ( blon, blat, zmean )
       get_topog_mean_1d  = .true.
   else
       get_topog_mean_1d = .false.
   endif
else !use_mpp_io
   if ( open_topog_file() ) then
       call interp_topog_1d ( blon, blat, zmean )
       get_topog_mean_1d = .true.
   else
       get_topog_mean_1d = .false.
   endif
endif !use_mpp_io
```
Then there is no need for `get_topog_mean_1d_mpp_io` ... 
` open_topog_file` is where the actual io calls happen.

You can probably get rid of `interp_topog_1d_mpp_io` as well and just wrapped the `input_data`/`input_data_mpp_io` calls (which is actual io) in if statements as well.

rem1776: I mainly added that argument because I thought we were supposed to be adding an argument to public functions to use mpp_io so the version can be selected when called.

My thought process with the mpp_io versions was to just have the fms2_io versions call it's mpp_io version and then any other calls from there would all mpp_io, but I see now that overcomplicated things a bit when only certain calls really need the mpp_io code.

topography/topography.F90
```diff
@@ -727,16 +859,28 @@ end function open_water_file
...
+ subroutine interp_topog_1d ( blon, blat, zout, flag, use_mpp_io_arg )
  real   , intent(in)  :: blon(:), blat(:)
  real   , intent(out) :: zout(:,:)
  integer, intent(in), optional :: flag
 
  real :: xdat(ipts+1), ydat(jpts+1)
  real :: zdat(ipts,jpts)
  real :: zout2(size(zout,1),size(zout,2))
+ logical, optional :: use_mpp_io_arg
+ logical :: mpp_io_use
```
thomas-robinson, line 871: Please document these variables with doxygen

topography/topography.F90
```diff
@@ -768,8 +912,21 @@ subroutine interp_topog_2d ( blon, blat, zout, flag )
  real :: zout2(size(zout,1),size(zout,2))
  integer :: js, je
  type (horiz_interp_type) :: Interp
+ logical, optional :: use_mpp_io_arg
+ logical :: mpp_io_use
```
thomas-robinson, line 916: Please document these variables with doxygen

thomas-robinson (changes requested):

thomas-robinson (dismissed):

topography/topography.F90
```diff
@@ -244,27 +248,29 @@ function get_topog_mean_2d (blon, blat, zmean, use_mpp_io_arg)
...
-   if ( open_topog_file() ) then
-       call interp_topog_2d ( blon, blat, zmean )
-       get_topog_mean_2d = .true.
+   if( mpp_io_use ) then
+     get_topog_mean_2d = open_topog_file_mpp_io(topog_file)
+     if ( get_topog_mean_2d ) call interp_topog_2d ( blon, blat, zmean,&
+                        use_mpp_io_arg=.true. )
    else
-       get_topog_mean_2d = .false.
+     get_topog_mean_2d = open_topog_file_mpp_io(topog_file)
```
bensonr, line 271: We are in the else, but still referencing an mpp_io function.

bensonr (changes requested):
I agree with @uramirez8707 with the use of the optional argument.  topography_init is the first functional call to each routine which will set the value of use_mpp_io via the namelist.  The optional argument allows the user to override the namelist setting.  I don't think this optional argument logic is needed.

topography/topography.F90
rem1776: Also fixed this call in [3ca8114](https://github.com/NOAA-GFDL/FMS/pull/571/commits/3ca81142d4dfa766fc9f9dca73b05211cef43acb)

topography/topography.F90
```diff
@@ -769,7 +807,6 @@ subroutine interp_topog_2d ( blon, blat, zout, flag )
  integer :: js, je
  type (horiz_interp_type) :: Interp
 
-    call input_data ( TOPOG_INDEX, xdat, ydat, zdat )
```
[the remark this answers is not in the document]
rem1776: I had mistakenly taken that out when I took out the optional arguments, It was added it back with https://github.com/NOAA-GFDL/FMS/pull/571/commits/b18827165e0487c7b75f0ac6c64d9baac17973c8

wrongkindofdoctor (approved):

uramirez8707 (approved):

thomas-robinson (approved):

bensonr (approved):

PR #605, opened 2020-10-27
FMS2io: get valid bug fix plus unit test

### LINKED ISSUE #585: fms2io: get_valid will not work properly if the fileobj is accessed by multiple pes
**Describe the bug**
In `get_valid` there is a ` if (fileobj%is_root)` if block:
https://github.com/NOAA-GFDL/FMS/blob/521a15135a99d1f2da7d82f238353945f82ce1dd/fms2_io/netcdf_io.F90#L1547 

Inside the if block, `get_variable_attribute` may be called. 
https://github.com/NOAA-GFDL/FMS/blob/521a15135a99d1f2da7d82f238353945f82ce1dd/fms2_io/netcdf_io.F90#L1558

In get_variable_attribute, after the attribute is read it tries to broadcast it to the other ranks (this only happens if the optional argument `broadcast` is not set to .false.)

https://github.com/NOAA-GFDL/FMS/blob/521a15135a99d1f2da7d82f238353945f82ce1dd/fms2_io/include/get_variable_attribute.inc#L75
but at that point only root_pe will ever make it to that broadcast so if the file obj is accessed my multiple pes it will probably hang there forever or just crash ... 

**To Reproduce**
1. call `open_file` using the optional [pelist](https://github.com/NOAA-GFDL/FMS/blob/521a15135a99d1f2da7d82f238353945f82ce1dd/fms2_io/netcdf_io.F90#L444) argument
2. call `get_valid` on a variable with any of the attributes `scale_factor`, `valid_range`, `missing_value`, etc ..
3. wait 

**Expected behavior**
The broadcast should be done correctly. 
The `get_variable_attribute` calls in `get_valid` should have `broadcast=.false.`. It is an easy fix. 

**System Environment**
This happens in any system. 

**Additional context**
N/A

uramirez8707:
**Description**
This PR:
1. Modifies get_valid so that there are no mpp_broadcasts inside a `if (fileobj%is_root)`
2. Adds a unit test to test get_valid and is_valid

Without the source code change the unit test fails when you are running with [2 cores](https://github.com/uramirez8707/FMS/blob/6a2e41c03b8ef5b32d0d6b205dce3a6c91f8c85c/test_fms/fms2_io/test_io_simple.sh#L41)

Fixes #585

**How Has This Been Tested?**
`make check` passes on skylake with intel20 and on travis 

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/fms2_io/test_get_is_valid.F90
+++ b/test_fms/fms2_io/test_get_is_valid.F90
@@ -1,0 +1,97 @@
+!***********************************************************************
+!*                   GNU Lesser General Public License
+!*
+!* This file is part of the GFDL Flexible Modeling System (FMS).
+!*
+!* FMS is free software: you can redistribute it and/or modify it under
+!* the terms of the GNU Lesser General Public License as published by
+!* the Free Software Foundation, either version 3 of the License, or (at
+!* your option) any later version.
+!*
+!* FMS is distributed in the hope that it will be useful, but WITHOUT
+!* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+!* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+!* for more details.
+!*
+!* You should have received a copy of the GNU Lesser General Public
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+!***********************************************************************
+
+!> @brief  This programs tests calls to fms2_io get_valid and is_valid
+
+program test_get_is_valid
+
+use   fms2_io_mod, only: fms2_io_init, open_file, FmsNetcdfFile_t, valid_t, &
+                         close_file, is_valid, get_valid
+use   mpp_mod    , only: mpp_init, mpp_exit, mpp_npes, mpp_get_current_pelist, &
+                         mpp_root_pe, mpp_pe, FATAL, mpp_error
+use   netcdf     , only: nf90_create, nf90_def_var, nf90_put_att, nf90_enddef, &
+                         nf90_close, nf90_clobber, nf90_64bit_offset, nf90_double
+
+use, intrinsic :: iso_fortran_env
+
+implicit none
+
+type(valid_t) :: valid_type               !> Fms2io valid object type
+type(FmsNetcdfFile_t) :: fileobj          !> Fms2io file obj
+real(kind=real64) :: sst(10,10)           !> Data
+integer :: mask_in(10,10)                 !> Array defining if sst(:,:) is valid (1) or not (0)
+integer :: ncid, varid, err               !> Netcdf integers needed
+integer, dimension(:), allocatable :: pes !> Current pelist
+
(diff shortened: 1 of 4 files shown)


test_fms/fms2_io/test_get_is_valid.F90
```diff
@@ -0,0 +1,97 @@
...
+implicit none
+
+type(valid_t) :: valid_type               !> Fms2io valid object type
+type(FmsNetcdfFile_t) :: fileobj          !> Fms2io file obj
+real(kind=real64) :: sst(10,10)           !> Data
+integer :: mask_in(10,10)                 !> Array defining if sst(:,:) is valid (1) or not (0)
+integer :: ncid, varid, err               !> Netcdf integers needed
+integer, dimension(:), allocatable :: pes !> Current pelist
+
+call mpp_init()
```
wrongkindofdoctor, line 42: This needs an mpp_init  test_level [parameter](https://github.com/NOAA-GFDL/FMS/blob/40c6cb9663f574b902ee8d832828be8f04c64feb/mpp/mpp.F90#L1368). 
You'll have to replace `mpp_exit()` with `MPI_FINALIZE(ierr)`.

wrongkindofdoctor (changes requested):

test_fms/fms2_io/test_get_is_valid.F90
uramirez8707: why? the `test_level` flag was added for when we were testing subroutines used by `mpp_init`, this is not doing that. 

This test is testing `get_valid` and `is_valid`.

wrongkindofdoctor: Ah, my mistake. Ignore this request.

wrongkindofdoctor (dismissed):

test_fms/fms2_io/test_get_is_valid.F90
```diff
@@ -0,0 +1,97 @@
...
+if (mpp_root_pe() .eq. mpp_pe()) then
+!> Create your own file with a variable to test with:
+   err = nf90_create('test_file.nc', ior(nf90_clobber, nf90_64bit_offset), ncid)
+   err = nf90_def_var(ncid, 'sst', nf90_double,varid)
+   err = nf90_put_att(ncid, varid, "_FillValue", real(999,kind=real64))
+   err = nf90_put_att(ncid, varid, "missing_value", real(999,kind=real64))
+   err = nf90_enddef(ncid)
+   err = nf90_close(ncid)
+endif
+
```
bensonr, line 57: May need an mpp_sync here to ensure the root pe has a chance to create/finalize the test file prior to all pes executing the open_file.

uramirez8707: That's a good idea. Thanks.

bensonr (approved):

wrongkindofdoctor (approved):

PR #651, opened 2020-12-29
Fix Makefile.am dependencies to not trigger rebuild if not needed

### LINKED ISSUE #239: CRAY variable in configure.ac should be more descriptive
The automake conditional variable `CRAY` should be more descriptive, as it can lead individuals to think this is a problem on Cray systems, and not just Cray systems that use the Cray Programming Environment, and do static linking by default.  

A name like: `CRAY_PRGENV_STATIC` may work.

**Other things to consider**
Where `CRAY` is used in the Makefiles will also need to be updated with the new name.

### LINKED ISSUE #648: Autoconf second `make` causes full rebuild
**Describe the bug**
If using the autoconf build, issuing a second `make` command with no changes will cause nearly a full re-build of the FMS library components.  This makes performing a `make install` more cumbersome as a rebuild is not needed.

**To Reproduce**
```
autoreconf -if
configure [options]
make
make # this should do nothing!
```

**Expected behavior**
The second call to `make` should print many lines of 'Nothing to be done for ...'.

underwoo:
**Description**
An earlier update modified the dependencies that would cause a second `make` to rebuild many of the source files.  This update fixes those dependencies.  This also adds a variable `MODDIR` to configure.ac for use in all Makefiles to find the required Fortran module files.

Fixes #648

**How Has This Been Tested?**
Mac OS, with brew installed netCDF, clang and gfortran.

**Checklist:**
- [x] `make distcheck` passes


wrongkindofdoctor (dismissed):

wrongkindofdoctor (dismissed):

underwoo:
Testing with the Intel compiler indicates the way Intel _always_ generates the `*.mod` files (even if it doesn't need to) is causing an issue (at times) with the parallel builds.  I think I have a _fix_ that I'm currently testing.  I converted this to a draft until I'm satisfied with the fix.

underwoo:
Work with the Intel compiler build is improving.  I believe I have the dependencies correct, and found a way to keep a rebuild from triggering due to Intel's regenerating the `*.mod` files.  However, the system I'm using (gaea) appears to be slow occasionally flushing the `*.mod` files to disk, leaving the file _in flight_ when the compiler attempts to use the `*.mod` file.  This is especially true when building in parallel on the NFS mounted home and Lustre file systems.

rem1776 (commented):
I'm not sure if this is an issue since this may just be from system/environment differences, but I tried compiling on gaea and the main code compiled fine but it looks like the test files fail during linking with this error: `ifort: error #10236: File not found:  '../../libFMS/.libs/libFMS.so'`

underwoo:
Thank you for reporting that.  I'll do another check on gaea, and I think I may know the problem -- gaea cannot use the `*.so` library the autotools build generates and thus is not available unless you add the `--enable-shared` option to `configure.

rem1776 (dismissed):

wrongkindofdoctor (dismissed):

bensonr (dismissed):

thomas-robinson:
@underwoo Did `make distcheck` pass for you?  If so, can you click the box?

thomas-robinson (dismissed):

rem1776 (approved):

bensonr (approved):
Since others are okay with the changes, I'm approving them.

wrongkindofdoctor (approved):

thomas-robinson (approved):

PR #700, opened 2021-03-04
Adds logic to look for a . in the timestamp in fms_io for compatibilty with fms2_io

### LINKED ISSUE #699: fms_io logic for adding a "." next to the time stamp
fms_io accepts a time stamp, and uses that time stamp to create a file name.  f[ms_io places a period between the timestamp and the name of the file](https://github.com/NOAA-GFDL/FMS/blob/master/fms/fms_io.F90#L2509).  fms2_io does not do this, so there may be some errors caused by mixing the IOs.

Add some logic to fms_io to allow for the old (passing `time_stamp` without a ".")  and new (passing `time_stamp` with a ".")  behavior

We considered just ignoring it since fms_io is no longer doing to be supported.  However, we decided to fix this in order to set up the unsupported situation.  We will not update anything in fms_io.F90 after the 2021.02 release.

thomas-robinson:
…

**Description**
Allows for fms_io to accept a timestamp that either has a "." on the end (compatible with fms2_io) or doesn't have a "." at the end (current fms_io behavior).  This will allow fms_io to create file names that are correct using the fms2_io interface. 

Fixes #699 

**How Has This Been Tested?**
Build with `make dist` with intel compiler on amd test machine.

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/fms/fms_io.F90
+++ b/fms/fms_io.F90
@@ -2508,7 +2508,7 @@ subroutine save_restart(fileObj, time_stamp, directory, append, time_level)
         if(len_trim(restartname)+len_trim(time_stamp) > 79) call mpp_error(FATAL, "fms_io(save_restart): " // &
           "Length of restart file name + time_stamp is greater than allowed character length of 79")
            has_dot = .false.
-           do i=1,len(trim(time_stamp))
+           do i=1,len(time_stamp)
               if (time_stamp(i:i) == ".") has_dot = .true.
            enddo
            if (has_dot) then
--- a/fms/fms_io.F90
+++ b/fms/fms_io.F90
@@ -2492,7 +2492,7 @@ subroutine save_restart(fileObj, time_stamp, directory, append, time_level)
   character(len=80)  :: restartname          ! The restart file name (no dir).
   character(len=336) :: restartpath          ! The restart file path (dir/file).
   integer :: i !< For looping
-  logical :: has_dot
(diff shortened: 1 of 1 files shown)


uramirez8707 (dismissed):

wrongkindofdoctor (dismissed):

fms/fms_io.F90
```diff
@@ -2506,7 +2507,15 @@ subroutine save_restart(fileObj, time_stamp, directory, append, time_level)
      if (PRESENT(time_stamp)) then
         if(len_trim(restartname)+len_trim(time_stamp) > 79) call mpp_error(FATAL, "fms_io(save_restart): " // &
           "Length of restart file name + time_stamp is greater than allowed character length of 79")
-        restartname = trim(time_stamp)//"."//trim(restartname)
+           has_dot = .false.
+           do i=1,len(trim(time_stamp))
+              if (time_stamp(i:i) == ".") has_dot = .true.
```
rem1776, line 2512: I think this should be `if ( trim(time_stamp)(i:i)  ..etc` or just remove the trim in the do loop, since the lengths wouldn't match up if any whitespace was trimmed

rem1776 (changes requested):

rem1776 (approved):

wrongkindofdoctor (approved):

PR #726, opened 2021-04-09
Introducing Grid2

GFDL-Eric:
**Description**
Adds grid2.F90 to the mosaic directory to eliminate the need for fms_io in these grid calls. Init and end routines were introduced (and put in fms) to minimize the number of file open calls.

Fixes # (issue)

**How Has This Been Tested?**
Tested in 2021.02-alpha1 lm4p1 land tests

**Checklist:**
- [ ] `make distcheck` passes


rem1776:
@GFDL-Eric It looks like it can't find the compiled modulefile for grid2 when compiling in the CI. I think if you change fms to be after mosaic in the top level makefile.am directory list it should fix it.

GFDL-Eric:
@rem1776 I guess that didn't work? I'm not sure if there's a different issue now?

rem1776:
@GFDL-Eric Looks like the same issue, just with a different module. I'm not sure if there's an ordering that would work or if it's some kind of circular dependency.

GFDL-Eric:
ok so the reordering sort of worked, but now I need to get rid of any fms dependencies in the entirety of the mosaic folder... classic

GFDL-Eric:
oof, I guess that includes fms_io mod since that's in the fms folder... which isn't going to work. I'm going to have to rethink this a bit. Ideas welcome!

GFDL-Eric:
I've tried a ton of stuff here including separating mosaic2.F90 and grid2.F90 into their own directory, to no avail. Maybe I'm missing something obvious with the Makefile or configure.ac?

rem1776:
I was just about to suggest those last changes. I got it to compile on gcc (hopefully it'll work with the ci too) with those changes plus
`nodist_include_HEADERS = $(MODFILES)`

And i also had to add paranthesis to the get_great_circle_algorithm header for some reason.

GFDL-Eric:
@rem1776 I implemented the changes and it seems like now the failure is outside the realm of grid2 and is just a data_override test failure? I'm not sure if this was failing on other PRs as well or if it's because of something I did.

rem1776:
Yeah it looks like it failed from
`FATAL from PE     0: NetCDF: Variable not found`
I'm not sure why it would fail for this PR though, I don't think the data_override test uses anything from this. I re-ran it before and it failed the same way so I don't think its the normal sporadic failures from input files but I'll see if it happens with gcc too

rem1776 (approved):
Nice that fixed it, guess i was wrong

PR #730, opened 2021-04-14
Workarounds for PGI

### LINKED ISSUE #704: FMS2021.01 PGI run crashes with meaningless  FATAL: unsupported type.
**Describe the bug**
Ocean-ice models crash with PGI (19) executable in FMS2020.04 and FMS2021.01 

Prod mode exec gives no output/error message or traceback of any sort. It just crashes with:
NOTE frsrun: error: nid01901: tasks 0-7: Segmentation fault

Debug mode exec gives:
FATAL from PE     6: unsupported type.

**To Reproduce**
```
module load fre/test
fremake -x /ncrc/home2/Niki.Zadeh/frerts/FMS2020.04_mom6_20210308/_0/OM4_extra.xml.20210308124657 -p ncrc4.pgi19 -t debug MOM6_SIS2_compile

frerun -x /ncrc/home2/Niki.Zadeh/frerts/FMS2020.04_mom6_20210308/_0/OM4_extra.xml.20210308124657 -p ncrc4.pgi19 -t debug MOM6_SIS2_Baltic --no-transfer -s
```
**Expected behavior**
Experiment should run fine with PGI as it does with Intel and GNU and as it did in xanadu.

**System Environment**
-p ncrc4.pgi19

**Additional context**
Prod  mode produces a core file , but pgdbg does not work on c4
```
pgdbg /lustre/f2/dev/Niki.Zadeh/FMS2021.01-beta1_mom6_20210218/MOM6_SIS2_compile/ncrc4.pgi19-prod/exec/fms_MOM6_SIS2_compile.x core
pgdbg-Error-Required tool java was not found
```

Why is there no traceback? I compiled with -g -traceback !

uramirez8707:
**Description**
This are hacks to get the ICE/OCEAN model working with pgi and fms2io

1. The call bellow was for some reason crashing [here](https://github.com/NOAA-GFDL/FMS/blob/main/fms2_io/include/get_variable_attribute.inc#L68) because pgi was not recognizing that 'cart' was a string. Changing it to ```cart(1:1)``` solved the "problem".
```F90
call get_variable_attribute(fileobj, axisname, "cartesian_axis", cart)
 ``` 
2. Similary the call bellow was failing because pgi was not recognizing that "variable_missing" was real. Changing it to variable_missing to a 1d variable magically solved the problem.
```F90
call get_variable_attribute(fileobj, variable_name, "_FillValue", variable_missing)
```
3. The code was also failing inside the read_data with `unsupported type` because PGI couldn't deal with the nested `select types`. Removing the `select type` in axis_utils and replacing it with simply a read_data call solved the "problem".
https://github.com/NOAA-GFDL/FMS/blob/81a5b6ea2559e2c31edbcab32a3230dfc31287be/axis_utils/axis_utils2.F90#L191-L198. 
4. PGI was failing with a seg_fault inside here. It seems like the pgi couldn't handle the mix of class * and the conversion to real. I couldn't find a way around this so I just created a "is_valid" interface #oldskool
https://github.com/NOAA-GFDL/FMS/blob/81a5b6ea2559e2c31edbcab32a3230dfc31287be/fms2_io/netcdf_io.F90#L1743-L1752

There may be other "issues" that were not hit with the experiment I was running. 

Fixes #704 

**How Has This Been Tested?**
`MOM6_SIS2_Baltic` experiment with pgi19 on gaea. We need to make sure all of the experiments run though...

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/data_override/data_override.F90
+++ b/data_override/data_override.F90
@@ -97,7 +97,7 @@ type override_type
    character(len=3)                 :: gridname
    character(len=128)               :: fieldname
    integer                          :: t_index                 !< index for time interp
-   type(horiz_interp_type), pointer :: horz_interp(:) =>NULL() !< index for horizontal spatial interp
+   type(horiz_interp_type), allocatable :: horz_interp(:) !< index for horizontal spatial interp
    integer                          :: dims(4)                 !< dimensions(x,y,z,t) of the field in filename
    integer                          :: comp_domain(4)          !< istart,iend,jstart,jend for compute domain
    integer                          :: numthreads
--- a/axis_utils/axis_utils2.F90
+++ b/axis_utils/axis_utils2.F90
@@ -80,9 +80,9 @@ contains
     cart = "N"
     if (variable_exists(fileobj, axisname)) then
       if (variable_att_exists(fileobj, axisname, "cartesian_axis")) then
-        call get_variable_attribute(fileobj, axisname, "cartesian_axis", cart)
+        call get_variable_attribute(fileobj, axisname, "cartesian_axis", cart(1:1))
       elseif (variable_att_exists(fileobj, axisname, "axis")) then
-        call get_variable_attribute(fileobj, axisname, "axis", cart)
+        call get_variable_attribute(fileobj, axisname, "axis", cart(1:1))
       endif
       axis_cart = uppercase(cart)
       if (axis_cart .eq. 'X' .or. axis_cart .eq. 'Y' .or. axis_cart .eq. 'Z' &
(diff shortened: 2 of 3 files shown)


data_override/data_override.F90
```diff
@@ -97,7 +97,7 @@ module data_override_mod
    character(len=3)                 :: gridname
    character(len=128)               :: fieldname
    integer                          :: t_index                 !< index for time interp
-   type(horiz_interp_type), pointer :: horz_interp(:) =>NULL() !< index for horizontal spatial interp
```
nikizadehgfdl, line 100: Just curious, is this also needed to bypass PGI issues? I never got the difference between pointer and allocatable, this is being allocated for sure on line 923. Why was it a pointer? Thanks heaven we don't have to consider _ALLOCATABLE :)

nikizadehgfdl (approved):

data_override/data_override.F90
bensonr: @nikizadehgfdl - at one time, the fortran standard did not allow allocatable elements within a derived data type.  That is why we have so many pointers being used as allocatable elements.  Now that the standard supports it, we can change them to properly be allocatable.

wrongkindofdoctor (approved):

GFDL-Eric (approved):
#believeinmagic

rem1776 (approved):

PR #750, opened 2021-05-20
Makes MAX_FIELDS_PER_FILE a namelist variable and allocates 

### LINKED ISSUE #747: Modify max_fields_per_file via diag_manager namelist?
The maximum number of variables allowed in a history file is controlled by:

https://github.com/NOAA-GFDL/FMS/blob/9d55115a331685e4c6e01f2dfb3b770a9f80fa37/diag_manager/diag_data.F90#L124

If I'm not mistaken, diagnostics are limited to 4D `(lon, lat, z, time)`.  I am working on updates to the radiation code that require an additional spectral dimension.  The work-around right now is to create one variable per spectral point, but this leads a very large number of variables and sometimes exceeds the hard limit set in the code above.  Unfortunately this limit is a compile-time constant, so if you want to change it you must recompile.  Can this instead be a run-time parameter accessible via the `diag_manager_nml` namelist?

Remove the `parameter` attribute and add this variable to the `diag_manager_nml` namelist.

Ideally allow 5D (and possible more) diagnostics, but I know that given the current state of the modules this is not feasible.

thomas-robinson:
**Description**
Changes the `max_fields_per_file` to a namelist variable, and allocated the `fields` with this value.

Fixes #747 

**How Has This Been Tested?**
`make` run with intel

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_data.F90
+++ b/diag_manager/diag_data.F90
@@ -183,7 +183,7 @@ use platform_mod
   TYPE file_type
      CHARACTER(len=128) :: name !< Name of the output file.
      CHARACTER(len=128) :: long_name
-     INTEGER, DIMENSION(max_fields_per_file) :: fields
+     INTEGER, allocatable, DIMENSION(:) :: fields
      INTEGER :: num_fields
      INTEGER :: output_freq
      INTEGER :: output_units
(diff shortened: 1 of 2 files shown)


thomas-robinson:
@menzel-gfdl can you test this?  Does it fulfill your request?

thomas-robinson:
@menzel-gfdl do you still want this?

bensonr (changes requested):
If this change goes through, the error message in diag_manager/diag_util.F90  at line 1356ff needs to be updated.

menzel-gfdl:
@thomas-robinson sorry I've been distracted by a lot of other projects.  What are your thoughts?  I proposed this because we kind of need 5d diagnostics (time, z, lat, lon, spectral dimension).  Right now our only option is to create one variable per spectral dimension (band, wavenumber, gpoint, etc.), which as you can imagine leads to a very high number of diagnostics.  Is there a better solution?

thomas-robinson:
@menzel-gfdl Yes there is a plan to support more dimensions in the diag manager rewrite.  Although the target is September, there is no guarantee that 5D support will be working by then.  If this PR solves your problem more immediately, then why not take it now.  I don't think there is a reason why this shouldn't be done.

rem1776:
closing since the rewritten diag manager does not have this limitation.

PR #766, opened 2021-06-22
Resolve "make check hangs with intel"

### LINKED ISSUE #765: `make check` hangs with intel
**Describe the bug**
`make check` hangs in intel.

It hangs on tests that are using an empty namelist. 
This because `max_length` gets initialized to -1:
https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/mpp/include/mpp_util.inc#L1436

and if the file is empty, it doesn't get into this do loop, 
https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/mpp/include/mpp_util.inc#L1450-L1465

and `max_length` ends up being 0. 
https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/mpp/include/mpp_util.inc#L1471

The code then tries to allocate a string array of length 0 and the code ends up hanging. 

**To Reproduce**
On skylake:
```
module load intel impi netcdf
autoreconf -install
./configure FC="mpiifort" CC="mpiicc"
make
make check
```
**Expected behavior**
Adding `max_length=1` here should the problem. 
https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/mpp/include/mpp_util.inc#L1449

to be consistent with `max_lines` which is 1 if there is an empty file. 
**System Environment**
- OS: skylake
- Compiler(s): intel20
- MPI type, and version  impi20
- netCDF Version: netcdf/4.6.1
- Configure options: 
This doesn't happen with gnu. 

**Additional context**
Add any other context about the problem.  If applicable, include where any files
that help describe, or reproduce the problem exist.

uramirez8707:
**Description**
Inits the max_length to 1 to avoid make check to hang when using an empty namelist or when calling `ascii_read` on an empty file.

Fixes #765 

**How Has This Been Tested?**
`make check` passes with intel on skylake and and it still passes on gnu. 

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -1447,6 +1447,7 @@ end function rarray_to_char
                             '.  (IOSTAT = '//trim(text)//')')
           else
              num_lines = 1
+             max_length = 1
              do
                 read (UNIT=f_unit, FMT='(A)', IOSTAT=status) str_tmp
                 if ( status .lt. 0 ) exit


bensonr (dismissed):

mpp/include/mpp_util.inc
```diff
@@ -1447,6 +1447,7 @@ end function rarray_to_char
                             '.  (IOSTAT = '//trim(text)//')')
           else
              num_lines = 1
+             max_length = 1
```
bensonr, line 1450: If we always have num_lines and max_length > 0 for any successfully opened file, how does a user know when the file is truly empty?  Could the hang have been avoided with better error checking of the results of the function?

wrongkindofdoctor (approved):

mpp/include/mpp_util.inc
uramirez8707: The output of `get_ascii_file_num_lines_and_length` is going to be 1 more than the actual number of lines and lengths. 

So an empty file will output 1 for num_lines and max_length to prevent an allocation of 0. 

I suppose we can add error checking here in `read_input_nml`
https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/fms2_io/fms_io_utils.F90#L496-L497

and here in `ascii_reads`

https://github.com/NOAA-GFDL/FMS/blob/caf4466241698a41e3272becbc2ed419a8d3c248/mpp/include/mpp_util.inc#L1332-L1333

bensonr: @uramirez8707 - what do you think this is best solution - error checking?  Or max_length=num_lines = 1?

uramirez8707: @bensonr 

I think we should leave them as `max_length=num_lines = 1`

I think it is cleaner than having to check for 0s every time `get_ascii_file_num_lines_and_length` is called.

bensonr (approved):

rem1776 (approved):

PR #772, opened 2021-07-01
Add gcc 11.1 check to configure script

### LINKED ISSUE #756: Calling `register_variable_attribute` with GCC 11 causes segmentation fault
**Describe the bug**
Some ice-ocean simulations crash during I/O when compiled with GCC 11.  The fail is in the calls to `register_variable_attribute`.

https://github.com/NOAA-GFDL/MOM6/blob/c432872bebda98b0acaff7d2378dc5a8523d3377/config_src/infra/FMS2/MOM_io_infra.F90#L1907-L1931

The program reaches this point in `register_variable_attribute_0d`:

https://github.com/NOAA-GFDL/FMS/blob/b9fc6515c7e729909e59a0f9a1efc6eb1d3e44d1/fms2_io/include/register_variable_attribute.inc#L40-L43

but then fails after exiting `get_variable_id` and just before or just after entering the `select type` block.

Instead, it just reports that the model has a segmentation fault at the end of the `netcdf_io.f90` file.  I'm sure this is nonsense and is either a bad jump induced by `select type` or corrupted stack pointer.

Commenting out the `select type` blocks (along with other relevant bits) does "fix" the problem, though of course it's no longer writing attributes.  So it does seem to be another `select type` compiler issue.

The experiment runs without error with GCC 8.4 and GCC 10.2.

**To Reproduce**
Build MOM6 dev/gfdl, target mkmf to the FMS2 infra, and FMS 2021.02 library.

Run the `Baltic_OM4_05` experiment from `MOM6-examples`.  it will fail during any time it attempts to write a netCDF attribute.

https://github.com/NOAA-GFDL/MOM6-examples/tree/dev/gfdl/ice_ocean_SIS2/Baltic_OM4_05

Single-core runs should be sufficient.

**Expected behavior**
Writing attributes to files should not cause a segmentation fault.

**System Environment**
- OS: Arch Linux (2021-05-30)
- Compiler(s): GCC 11.1 (GCC 8.4 and 10.2 work correctly)
- MPI: OpenMPI 4.0.5
- netCDF Version: netcdf 4.7.4, netcdf-fortran 4.5.3
- Configure options: Using mkmf template `linux-gnu.mk`, with BOZ and argument mismatch support flags: 

**Additional context**
The `include` intrinsics appear to be misreporting the paths of these include files and gdb is unable to find any of this content.  You will need to step through to confirm some of these details.

Backtrace below:
```
Thread 1 "MOM6" received signal SIGSEGV, Segmentation fault.
0x000055555822cba1 in netcdf_io_mod::register_variable_attribute_0d (fileobj=..., variable_name=..., attribute_name=..., attribute_value=..., str_len=8, _variable_name=4, 
    _attribute_name=9) at $HOME/MOM6-examples/src/FMS/fms2_io/netcdf_io.F90:1993
1993	end module netcdf_io_mod
(gdb) bt
#0  0x000055555822cba1 in netcdf_io_mod::register_variable_attribute_0d (fileobj=..., variable_name=..., attribute_name=..., attribute_value=..., str_len=8, _variable_name=4, 
    _attribute_name=9) at $HOME/MOM6-examples/src/FMS/fms2_io/netcdf_io.F90:1993
#1  0x000055555716e847 in mom_io_infra::write_metadata_axis (io_handle=..., axis=..., name=..., units=..., longname=..., cartesian=..., 
    sense=<error reading variable: Cannot access memory at address 0x0>, domain=..., data=..., edge_axis=<error reading variable: Cannot access memory at address 0x0>, 
    calendar=..., _name=4, _units=40, _longname=8, _cartesian=1, _calendar=0) at $HOME/MOM6-examples/src/MOM6/config_src/infra/FMS2/MOM_io_infra.F90:1864
#2  0x0000555556a1f23b in mom_io::create_file (io_handle=..., filename=..., vars=<error reading variable: value requires 92000 bytes, which is more than max-value-size>, 
    novars=31, fields=<error reading variable: value requires 137600 bytes, which is more than max-value-size>, threading=401, 
    timeunit=<error reading variable: Cannot access memory at address 0x0>, g=..., dg=<error reading variable: Cannot access memory at address 0x150>, gv=..., checksums=..., 
    extra_axes=<error reading variable: Location address is not set.>, global_atts=<error reading variable: Location address is not set.>, _filename=15)
    at $HOME/MOM6-examples/src/MOM6/src/framework/MOM_io.F90:306
#3  0x0000555556ce5b38 in mom_restart::save_restart (directory=..., time=..., g=..., cs=0x55555b0f5f60, 
    time_stamped=<error reading variable: Cannot access memory at address 0x0>, filename=..., gv=..., 
    num_rest_files=<error reading variable: Cannot access memory at address 0x0>, write_ic=<error reading variable: Cannot access memory at address 0x0>, _directory=200, 
    _filename=0) at $HOME/MOM6-examples/src/MOM6/src/framework/MOM_restart.F90:978
#4  0x0000555555a6029d in ocean_model_mod::ocean_model_save_restart (os=0x55555b696760, time=..., directory=..., filename_suffix=..., _directory=0, _filename_suffix=0)
    at $HOME/MOM6-examples/src/MOM6/config_src/drivers/FMS_cap/ocean_model_MOM.F90:745
#5  0x0000555555a603ad in ocean_model_mod::ocean_model_end (ocean_sfc=..., ocean_state=0x55555b696760, time=...)
    at $HOME/MOM6-examples/src/MOM6/config_src/drivers/FMS_cap/ocean_model_MOM.F90:713
#6  0x0000555555573dbd in coupler_main::coupler_end () at $HOME/MOM6-examples/src/coupler/coupler_main.F90:2034
#7  0x00005555555707a4 in coupler_main () at $HOME/MOM6-examples/src/coupler/coupler_main.F90:1148
#8  0x0000555555581c2a in main (argc=1, argv=0x7fffffffe958) at $HOME/MOM6-examples/src/coupler/coupler_main.F90:313
#9  0x00007ffff7582b25 in __libc_start_main () from /usr/lib64/libc.so.6
#10 0x000055555556d24e in _start ()
```

rem1776:
**Description**
Adds check in the configure script for GCC 11.1 and outputs a warning stating it is unsupported if found. This can be switched to a error if preferable but i figured since it still compiles it should just state that it is unsupported.

Fixes #756 

**How Has This Been Tested?**
Tested with intel and gnu 9.3.0 and 11.1.0 on amd

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -272,16 +272,6 @@ if test -n "$fc_version_info"; then
   FC_VERSION="$FC_VERSION ( $fc_version_info )"
 fi
 
-# Check if gcc is 11.1 for class(*) select type bug
-AC_MSG_CHECKING([if gcc 11.1 is loaded])
-if [ test -n "`$FC --version | grep GNU | grep 11.1`" ]; then
-  AC_MSG_RESULT([yes])
-  AC_MSG_RESULT([Warning: Compilation with GNU compilers version 11.1 is unsupported by this version\
- of FMS. Please use another compiler version])
-else
-  AC_MSG_RESULT([no])
-fi
-
 # Check if netcdf is version 4.7.4 to skip failing fms2_io tests
 AC_MSG_CHECKING([if netCDF 4.7.4 is loaded])
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
@@ -375,4 +365,10 @@ AC_OUTPUT()
 ## if tests are being skipped, output warning at end of output
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
+fi
+
+# Check if gcc is 11.1 for class(*) select type bug
+if [ test -n "`$FC --version | grep GNU | grep 11.1`" ]; then
+  AC_MSG_RESULT([Warning: Compilation with GNU compilers version 11.1 is unsupported by this version\
+ of FMS. Please use another compiler version])
 fi
--- a/configure.ac
+++ b/configure.ac
@@ -376,4 +376,3 @@ AC_OUTPUT()
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
 fi
-
--- a/configure.ac
+++ b/configure.ac
@@ -272,6 +272,16 @@ if test -n "$fc_version_info"; then
   FC_VERSION="$FC_VERSION ( $fc_version_info )"
 fi
 
+# Check if gcc is 11.1 for class(*) select type bug
+AC_MSG_CHECKING([if gcc 11 is loaded])
+if [ test -n "`$FC --version | grep GNU | grep 11\..\..`" ]; then
+  AC_MSG_RESULT([yes])
+  AC_MSG_ERROR([Compilation with gcc and gfortran 11 is unsupported by this version\
+ of FMS. Please use another compiler version])
+else
+  AC_MSG_RESULT([no])
+fi
+
 # Check if netcdf is version 4.7.4 to skip failing fms2_io tests
 AC_MSG_CHECKING([if netCDF 4.7.4 is loaded])
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
@@ -367,8 +377,3 @@ if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
 fi
 
-# Check if gcc is 11.1 for class(*) select type bug
-if [ test -n "`$FC --version | grep GNU | grep 11\..\..`" ]; then
-  AC_MSG_RESULT([Warning: Compilation with gcc and gfortran 11 is unsupported by this version\
- of FMS. Please use another compiler version])
-fi
--- a/configure.ac
+++ b/configure.ac
@@ -368,7 +368,7 @@ if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
 fi
 
 # Check if gcc is 11.1 for class(*) select type bug
-if [ test -n "`$FC --version | grep GNU | grep 11.1`" ]; then
-  AC_MSG_RESULT([Warning: Compilation with GNU compilers version 11.1 is unsupported by this version\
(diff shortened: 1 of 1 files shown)


wrongkindofdoctor (approved):

configure.ac
```diff
@@ -366,3 +366,9 @@ AC_OUTPUT()
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
 fi
+
+# Check if gcc is 11.1 for class(*) select type bug
+if [ test -n "`$FC --version | grep GNU | grep 11.1`" ]; then
```
bensonr, line 371: Do we know that 11.0 is safe?  I would think it best to just not support the 11 stream until we know 11.0 and 11.2 have the bug/bugfix.

rem1776: I agree, it's updated now to warn on any 11 versions.

wrongkindofdoctor (approved):

bensonr (approved):

configure.ac
```diff
@@ -366,3 +366,9 @@ AC_OUTPUT()
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
 fi
+
+# Check if gcc is 11.1 for class(*) select type bug
+if [ test -n "`$FC --version | grep GNU | grep 11\..\..`" ]; then
+  AC_MSG_RESULT([Warning: Compilation with gcc and gfortran 11 is unsupported by this version\
```
thomas-robinson, line 372: Should this be a warning or a fatal?  I don't think we should allow compiling with gcc11 at all.

thomas-robinson (changes requested):

configure.ac
rem1776: I had it as an error originally but figured since it still compiles successfully compilation should still be allowed. I'll switch it to an error, odds are the warning will get ignored half the time anyway.

bensonr (approved):

configure.ac
```diff
@@ -366,3 +376,4 @@ AC_OUTPUT()
 if [ test "`nc-config --version`" == "netCDF 4.7.4" ]; then
   AC_MSG_RESULT([Warning: netCDF v4.7.4 is loaded. Incompatible fms2_io tests will be skipped and there may be netCDF issues with model runs.])
 fi
+
```
thomas-robinson, line 379: Remove added whitespace

thomas-robinson (changes requested):

thomas-robinson (approved):

bensonr (approved):

configure.ac
marshallward: I'm not sure if there even was an 11.0 release: http://gcc.gnu.org/releases.html

PR #773, opened 2021-07-02
Fixes to mpp_broadcast_domains to give the data domain initial values

bensonr:
Fixes to domain broadcast routines in mpp to give the data domain initial values and ensure their values are correct when the data is unpacked after communication.  

**Description**
The bug could have affected all nested runs using the updated nesting in FV3 when using gnu compilers.  Intel generally gives all variables sane initial values (zero) and therefore doesn't always catch bugs of this nature.

Fixes # (issue)

**How Has This Been Tested?**
In addition to the Intel CI tests on gaea, the multiple test cases and configurations of the SHiELD model were run with both Intel and Gnu compilers.  The previously failing nest tests with gnu are now passing.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_domains_misc.inc
+++ b/mpp/include/mpp_domains_misc.inc
@@ -541,6 +541,10 @@ end subroutine init_nonblock_type
           domain%x%compute%end   = -1
           domain%y%compute%begin =  1
           domain%y%compute%end   = -1
+          domain%x%data   %begin = -1
+          domain%x%data   %end   = -1
+          domain%y%data   %begin = -1
+          domain%y%data   %end   = -1
           domain%x%global %begin = -1
           domain%x%global %end   = -1
           domain%y%global %begin = -1
@@ -582,6 +586,10 @@ end subroutine init_nonblock_type
              domain%list(listpos)%y%compute%end   = msg(5)
              domain%list(listpos)%tile_id(1)      = msg(6)
              if(domain%x(1)%global%begin < 0) then
+                domain%x(1)%data  %begin = msg(2)
+                domain%x(1)%data  %end   = msg(3)
+                domain%y(1)%data  %begin = msg(4)
+                domain%y(1)%data  %end   = msg(5)
                 domain%x(1)%global%begin = msg(2)
                 domain%x(1)%global%end   = msg(3)
                 domain%y(1)%global%begin = msg(4)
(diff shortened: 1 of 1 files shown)


bensonr:
@rem1776 @thomas-robinson  -  I'd like this fix to be considered for a 2021.02.01 patch release.

wrongkindofdoctor (approved):

uramirez8707 (dismissed):

rem1776:
@bensonr  Since this is for a patch, I changed the base branch for this to go directly to the 2021.02 release tag. This was off the current main, so the documentation update and other PR's that were merged in after are included here now as well. I believe you should be able to checkout the `release/2021.02` branch and cherry pick the commit in order for the other changes to be excluded, otherwise they'll just get merged into the patch.

bensonr:
@rem1776 - this needs to go in the patch and also be put into main.  Can we
put it in main and then cherry-pick to the patch?

rem1776:
> @rem1776 - this needs to go in the patch and also be put into main. Can we put it in main and then cherry-pick to the patch?

Yeah we can put it in main and cherry pick the patch from there (sorry other reviewers!).
Usually the patch changes are merged into main before the next release with any other PR's, so the only difference is when it's going in.

mpp/include/mpp_domains_misc.inc
```diff
@@ -597,6 +605,10 @@ end subroutine init_nonblock_type
                 endif
                 domain%ntiles = msg(12)
              else
+                domain%x(1)%data  %begin = msg(2) - msg(7)
+                domain%x(1)%data  %end   = msg(3) + msg(9)
```
uramirez8707, line 609: should this be `msg(3) + msg(8)`? which i think corresponds to the `ehalo`

mpp/include/mpp_domains_misc.inc
```diff
@@ -718,6 +738,10 @@ end subroutine init_nonblock_type
                 endif
                 domain_out%ntiles            = msg(12)
              else
+                domain_out%x(1)%data  %begin = msg(2) - msg(7)
+                domain_out%x(1)%data  %end   = msg(3) + msg(9)
```
uramirez8707, line 742: same comment as above

mpp/include/mpp_domains_misc.inc
bensonr: good catch

mpp/include/mpp_domains_misc.inc
bensonr: again...

uramirez8707 (approved):

wrongkindofdoctor (approved):

PR #778, opened 2021-07-14
replace hardwired input.nml with optional user specified path to input.nml

### LINKED ISSUE #779: relax the restriction on hard-wired references to `input.nml`
No.  Not a problem but poses a limitation.

`input.nml` is hard-wired and expected to exist in CWD.
For applications in DA with outer loops, this poses a limitation when there is a need to read a unique input.nml for that outer loop.

Moving `input.nml` in compiled code that uses FMS.

aerorahul:
**Description**

**THIS IS A DRAFT PR TO FACILITATE DISCUSSION**

Fixes #779 

`input.nml` is hard-wired and expected to exist in `CWD`.
For applications in DA with outer loops, this poses a limitation when there is a need to read a unique `input.nml` for that outer loop.

What is the solution being proposed in this PR:
We added a variable `input_nml_filename` in `mpp.F90`. Its default value is `input.nml`.
We added an optional argument `nml_filename` for the `subroutine fms_init()` in `fms.F90` 

During the initialization of FMS via `fms_init`, one can pass a path to their custom `input.nml`.  Absence of which is the current state of work.

We replaced all hard-wired instances of `input.nml` with the variable `input_nml_filename`.

This was tested in the CI.

**THE SECTION BELOW WILL BE UPDATED IF THIS DRAFT IS READY FOR A PR**

Include a summary of the change and which issue is fixed. Please also include
relevant motivation and context. List any dependencies that are required for
this change.

Fixes # 779

**How Has This Been Tested?**
Tested in the CI only (for now).

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/diag_integral/diag_integral.F90
+++ b/diag_integral/diag_integral.F90
@@ -69,7 +69,7 @@ use time_manager_mod, only:  time_type, get_time, set_time,  &
                              operator(+),  operator(-),      &
                              operator(==), operator(>=),     &
                              operator(/=)
-use mpp_mod,          only:  input_nml_file, get_unit
+use mpp_mod,          only:  input_nml_file, input_nml_filename, get_unit
 use fms_mod,          only:  open_file, error_mesg, &
                              check_nml_error, &
                              fms_init, &
@@ -326,7 +326,7 @@ real,dimension(:,:), intent(in), optional :: area_in
 !-------------------------------------------------------------------------------
 !    read namelist.
 !-------------------------------------------------------------------------------
-    if ( file_exists('input.nml')) then
+    if ( file_exists(input_nml_filename)) then
         read (input_nml_file, nml=diag_integral_nml, iostat=io)
         ierr = check_nml_error(io,'diag_integral_nml')
     endif
--- a/diag_manager/diag_manager.F90
+++ b/diag_manager/diag_manager.F90
@@ -197,6 +197,7 @@ use platform_mod
        & get_ticks_per_second
   USE mpp_io_mod, ONLY: mpp_open, mpp_close, mpp_get_maxunits
   USE mpp_mod, ONLY: mpp_get_current_pelist, mpp_pe, mpp_npes, mpp_root_pe, mpp_sum
+  USE mpp_mod, ONLY: input_nml_filename
 
 #ifdef INTERNAL_FILE_NML
   USE mpp_mod, ONLY: input_nml_file
@@ -3585,7 +3586,7 @@ CONTAINS
 #ifdef INTERNAL_FILE_NML
     READ (input_nml_file, NML=diag_manager_nml, IOSTAT=mystat)
 #else
-    IF ( file_exist('input.nml') ) THEN
+    IF ( file_exist(input_nml_filename) ) THEN
        nml_unit = open_namelist_file()
        READ (nml_unit, diag_manager_nml, iostat=mystat)
        CALL close_file(nml_unit)
--- a/fms/fms.F90
+++ b/fms/fms.F90
@@ -331,7 +331,7 @@ subroutine fms_init (localcomm, nml_filename )
  use mpp_mod,       only: input_nml_filename
 
  integer, intent(in), optional :: localcomm
- character(len=128), intent(in), optional :: nml_filename
+ character(len=1024), intent(in), optional :: nml_filename
  integer :: unit, ierr, io
  integer :: logunitnum
 
--- a/fms/fms.F90
+++ b/fms/fms.F90
@@ -136,7 +136,8 @@ use          mpp_mod, only:  mpp_error, NOTE, WARNING, FATAL,    &
                              mpp_set_stack_size,                 &
                              stdin, stdout, stderr, stdlog,      &
                              mpp_error_state, lowercase,         &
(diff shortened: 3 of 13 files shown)


aerorahul:
@bensonr 
Please comment if this is an appropriate way to do this or there are better alternatives.
TIA.

bensonr:
@aerorahul - is this fix to address the fact you won't have a file named input.nml at all, or is it the need to allow for different namelists based on your configuration?

aerorahul:
> @aerorahul - is this fix to address the fact you won't have a file named input.nml at all, or is it the need to allow for different namelists based on your configuration?

This is a fix so that we can provide a namelist file named differently (e.g. `input_outerloop1.nml`, etc.) and located not in CWD.
We still need __a__ namelist file to initialize FMS (unless using the defaults).  This PR makes the name and location of the namelist file user configurable during initialization of FMS.

aerorahul:
@bensonr does that address your question?

bensonr:
@aerorahul - the read_input_nml is actually uses a generic function read_ascii_file for the actual ingest of input.nml or files with a specific pattern.  You can actually accomplish what you want by using the read_ascii_file routine as it takes in a generic 'filename' argument.  To utilize this, you can:

a) create a separate variable length allocatable character array (VLACA), read data into it using read_ascii_file, and use that VLACA within the fortran namelist reads

b) deallocate input_nml_file, use read_ascii_file to repopulated it, and use the existing input_nml_file FLACA within the fortran namelist reads

c) change the interface to read_input_nml and add an additional optional argument for a filename - with logic to fail if this filename does not exist.  This may require some changes to model code, but shouldn't be any more intrusive than those proposed.

@GFDL-Eric - your thoughts?

GFDL-Eric:
@bensonr as it sounds to me like this is going to be supplemental to the existing "input_nml_file", I would recommend option a, which is ostensibly what read_input_nml does with a bit of window dressing. FWIW, the "ascii_read" routine from fms2_io should roll the get_ascii_file_num_lines_and_length and the read_ascii_file call into a convenient package:

```
character(len=:), dimension(:), allocatable :: my_input_nml
character(len=256) :: filename
call ascii_read(filename, my_input_nml)
```

aerorahul:
Thanks for the updates @bensonr @GFDL-Eric.
Since my knowledge is quite superficial on this, how do I get to call these before `fms_init` is called. Or are you suggesting we do this in our code after `fms_init`. In which case, we will need an `input.nml` in CWD to "initialize" as always and then over-write it with the above calls.

GFDL-Eric:
> Thanks for the updates @bensonr @GFDL-Eric.
> Since my knowledge is quite superficial on this, how do I get to call these before `fms_init` is called. Or are you suggesting we do this in our code after `fms_init`. In which case, we will need an `input.nml` in CWD to "initialize" as always and then over-write it with the above calls.

@aerorahul We are suggesting that this happen after fms_init, in which case, yes, you would still need the "standard" input.nml in CWD. Note that if you use a separate variable name (which was option a), your calls to access those parameters would come from that variable (as opposed to input_nml_file), which will likely change some of the calling routines you currently have in the model (but as the namelist is only generally read in once per nml section, hopefully it won't be too many overall changes).

aerorahul:
@GFDL-Eric 
In that case, we then need a `input.nml` in CWD which is essentially a dummy.  
And an actual one that we will use and contain values we wish to set and use.
Besides, we will read it in `my_input_nml` array, but that won't update the one `input_nml_file` structure which is used in FMS, the model and everywhere else where it is needed.

bensonr:
@aerorahul - I want to clarify that while FMS still has some capability for ingesting the input.nml via the open/read/close mechanism, that macro and functionality has already been removed from many of the FMS subsystems and will be fully removed from the FMS library soon.  The various subsystems within FMS initialization will still be happening from a consistent space and the input_nml_file will be the VLACA used with fortran NML reads in FMS.

As some background, on FMS/MPP initialization just after MPI_Init, a global MPP pelist is defined, logging is initialized, some MPP default behaviors are defined (via arguments to mpp_init) and then the input.nml from the CWD is read in by read_input_nml.  As mentioned above, it is the information in the CWD input.nml (ingested into the input_nml_file VLACA) that will be used for initializing all FMS subsystems.  With this procedure, there is not a simple way for one to ingest a different namelist file at startup.

We've outlined a procedure for you to either overwrite the input_nml_file VLACA or create a new VLACA, using a file of your choice, for your component(s) to use when performing NML reads.  If needed, we can discuss this further, but I'd recommend a verbal discussion to hash things out.

aerorahul:
One last reply and if that doesn't work, then we can talk.

So you would rather us do the following at our end:

```
use mpp_mod, only: input_nml_file
use fms2_io_mod, only: ascii_read

character(len=1024) :: my_input_nml_filename = "/path/to/my/custom/input.nml"
call fms_init()
call ascii_read(my_input_nml_filename, input_nml_file)
```

If thats it, then great.  I can close the PR.

Would you still be interested in removing the hardwired references to `input.nml` in the rest of the PR?

danholdaway:
Superseded by #787

rem1776:
as mentioned above, replaced by #787

PR #787, opened 2021-07-23
JCSDA changes

danholdaway:
**Description**

This a series of changes implemented by the JCSDA to enable data assimilation applications. 

The main changes are to allow for paths to the namelist files, field_table, and the INPUT directory. The solution used follows the one suggested provided to @aerorahul by @bensonr.

Other change include allowing more tracers for EMC's CMAQ assimilation and changes to MPP adjoint for data assimilation with MOM6

**How Has This Been Tested?**
We have run the the unit tests of the JEDI system

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included
- [ ] `make distcheck` passes

PR #807, opened 2021-09-02
JCSDA development without the adjoint changes

danholdaway:
**Description**

This a series of changes implemented by the JCSDA to enable data assimilation applications (without the adjoint part). 

The main changes are to allow for paths to the namelist files, field_table, and the INPUT directory. The solution used follows the one suggested provided to @aerorahul by @bensonr.

Other change include allowing more tracers for EMC's CMAQ assimilation and changes to MPP adjoint for data assimilation with MOM6

**How Has This Been Tested?**
We have run the the unit tests of the JEDI system

**Checklist:**
- [ ] `make distcheck` passes


danholdaway:
@bensonr hoping we can get this merged more quickly without the adjoint changes. Since it seems GMAO can't even build with the latest FMS it will be a long time before they can test those adjoint changes.

bensonr:
@danholdaway - with this new PR, do we still need #787 to remain open?

fms/fms_io.F90
```diff
@@ -7909,7 +7911,11 @@ subroutine get_mosaic_tile_grid(grid_file, mosaic_file, domain, tile_count)
     allocate(tile_id(ntileMe))
     tile_id = mpp_get_tile_id(domain)
     call read_data(mosaic_file, "gridfiles", grid_file, level=tile_id(tile) )
-    grid_file = 'INPUT/'//trim(grid_file)
+    if (.not. present(custom_path)) then
+      grid_file = 'INPUT/'//trim(grid_file)
+    else
+      grid_file = trim(custom_path)//'/'//trim(grid_file)
+    endif
```
bensonr, line 7918: There are entries in the grid_spec.nc for the <comp>_mosaic_dir, as well as the 'gridlocation' within the individual grid mosaics.  If one does not use relative paths that are specific to the rundir directory structure, these cannot be used.  However, there are NetCDF file manipulation tools that one could use during job creation to allow us to support this option as an alternative to the custom pathing option.  Thoughts?

fms/fms_io.F90
```diff
@@ -137,7 +137,7 @@ module fms_io_mod
 integer, parameter, private :: max_fields=400
 integer, parameter, private :: max_axes=40
 integer, parameter, private :: max_atts=20
-integer, parameter, private :: max_domains = 10
+integer, parameter, private :: max_domains = 1000
```
bensonr, line 140: 1000 is a bit extreme for the max_domain limit - is this a realistic limit?

danholdaway:
> @danholdaway - with this new PR, do we still need #787 to remain open?

I would suggest closing that PR. The conversation involving GMAO folks could move to an issue perhaps.

sanAkel:
@danholdaway (and others involved), I would like to say a few things:

1. I was _mentioned_ in the now [closed PR](https://github.com/NOAA-GFDL/FMS/pull/787) because building GEOSgcm **includes** building the coupled model, which in turn includes MOM5 and MOM6 ocean models. 
2. Reg. MOM6 in GEOS, as [this](https://github.com/NOAA-GFDL/FMS/pull/787#issuecomment-911836006) and others _above_ it should indicate, we have been **able to build** GEOSgcm, and also **run it.**

Therefore, 

> Since it seems GMAO can't even build with the latest FMS 

Is **incorrect**. I would appreciate if would please:
~~Since it seems GMAO can't even build with the latest FMS~~

Also note that:
- MOM6 in GEOS is up to date with `dev/gfdl` so you can't ask for a more _recent_ version!
- As for FV3, it is up to @wmputman.

Thanks!

fms/fms_io.F90
danholdaway: Would there be an issue with doing it this way? I fear that adding these extra steps into the workflow would add quite a bit of complexity. For example you don't have to run with these grid spec files, in most cases we create the grid on the fly. It's only the regional system that needs to read the grid. So the workflow would have to have extra conditionals on that and it could get rather complicated.

fms/fms_io.F90
danholdaway: Would 100 be OK?

danholdaway:
@sanAkel I was referring to the issues Matt was reporting. But it seems he got it building so that's great. In any case it would be a while before the adjoint could be tested with the branch we were trying to merge so we closed that and took out the change.

fms/fms_io.F90
bensonr: No, I was merely pointing out the capability for discussion.  As it seems to add complexity, I am okay with your proposed solution.

danholdaway: OK, thanks. It's good to know about the other solution.

fms/fms_io.F90
bensonr: Yes.  As the fms_io has been superseded by fms2_io and has moved into available but unsupported mode, keeping a sane limit is preferable.  A transition guide has been provided and can be found [here](https://github.com/NOAA-GFDL/FMS/blob/main/fms2_io/readme.md).  More detailed information on fms2_io interfaces can also be found on the documentation site:  http://noaa-gfdl.github.io/FMS/group__fms2__io.html

sanAkel:
> @sanAkel I was referring to the issues Matt was reporting. But it seems he got it building so that's great. In any case it would be a while before the adjoint could be tested with the branch we were trying to merge so we closed that and took out the change.

@danholdaway Thanks for the reply. I have nothing to do/say with the 
>adjoint could be tested with the branch

Whatever floats your boat and makes it go fast! 😉 
When you want it to be going in MOM6 modeled fluid, let me know. Thanks again!

fms/fms_io.F90
bensonr: @danholdaway - If you think 100 will cover your needs, can you make the change so we can review and get this ready for merge?

bensonr (approved):

fms/fms_io.F90
danholdaway: I've reduced to 100. I don't think there are any JEDI applications that instantiate more than 100 simultaneous geometry objects.

rem1776 (approved):

fms/fms_io.F90
```diff
@@ -7895,11 +7896,12 @@ end subroutine get_mosaic_tile_file_ug
 
 
   !#############################################################################
-  subroutine get_mosaic_tile_grid(grid_file, mosaic_file, domain, tile_count)
+  subroutine get_mosaic_tile_grid(grid_file, mosaic_file, domain, tile_count, custom_path)
     character(len=*), intent(out)          :: grid_file
     character(len=*), intent(in)           :: mosaic_file
     type(domain2D),   intent(in)           :: domain
     integer,          intent(in), optional :: tile_count
+    character(len=*), intent(in), optional :: custom_path
```
thomas-robinson, line 7904: Please add a doxygen description of this new variable per the [style guide functions section](https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#functions)

"Inline doxygen descriptions for all arguments, except the result variable."

thomas-robinson (commented):
This just needs a doxygen comment for the description of the new variable `custom_path`

fms/fms_io.F90
danholdaway: Will do. We'll have another PR coming up so will add it then.

Issue #674, opened 2021-02-01
Fms2io replace types with those defined in platform_mod


uramirez8707:
**Describe the bug**
One of the changes in https://github.com/NOAA-GFDL/FMS/pull/664, was replace 
```F90 
type is (integer(kind=int32)) 

```
to 
```F90
type is (integer(kind=i4_kind))
```
similarly with int64, real32, real64 to i8_kind, r4_kind, r8_kind in fms2io. 

This was not done in:
1.axis_utils2.F90
2. compressed_write.inc 
3. register_variable_attribute.inc
4. get_global_attribute.inc
5. test_fms/fms2_io/land_compressed_restart_file_test.inc
6. test_fms/fms2_io/land_unstructured_restart_file_test.inc
7. test_fms/fms2_io/setup.F90

**To Reproduce**
N/A

**Expected behavior**
Code should we consistent. 

**System Environment**
Any

**Additional context**
N/A

uramirez8707:
fyi @wrongkindofdoctor

wrongkindofdoctor:
*Sigh* Always something with these typedefs. Thanks for listing the files. I'll fix them.

PR #859, opened 2021-11-09
Clean up unused/uninitialized variables and other warnings

### LINKED ISSUE #190: Remove unused variables
Several sources/routines have unused variables.  These should be removed.  Most of the unused variables can be found using the `-Wunused` Gfortran option.

### LINKED ISSUE #192: Ensure variables initialized before use

rem1776:
**Description**
This PR mainly cleans up unused/uninitialized variables, implicit conversions, and character truncation warnings where possible(excluding old io routines).

For real comparisons I figured they have worked so far and might behave differently if changed, so probably wasn't worth the extra operations for proper difference comparisons. Besides that I also added a macro, MPP_TYPE_LOGICAL in order to initialize some output data to the correct type.

Fixes #190
Fixes #192 

**How Has This Been Tested?**
Tested on amd with gcc 9, intel/impi 2020_up2, netcdf 4.8.0

**Checklist:**
- [x] `make distcheck` passes


wrongkindofdoctor (approved):

random_numbers/random_numbers.F90
```diff
@@ -86,7 +86,7 @@ subroutine getRandomNumber_Scalar(stream, number)
     type(randomNumberStream), intent(inout) :: stream
     real,                     intent(  out) :: number
 
-    number = getRandomReal(stream%theNumbers)
+    number = real(getRandomReal(stream%theNumbers))
```
bensonr, line 89: getRandomReal is already coming back as a real(kind=8) value.  Is this to continue to allow r4 compilation?

bensonr (changes requested):
You modified the mpp code to add a MPP_LOGICAL_TYPE and have it set a value to either .false. or 0 when the macro was expanded in a a few of the mpp routines.  I'm not sure I understand why you did that - especially since it now has code taking loc(.false.)?

constants/constants.F90
```diff
@@ -61,8 +61,6 @@ module constants_mod
 ! by fms_init
 public :: version
 
-real :: realnumber !< dummy variable to use in HUGE initializations
-
```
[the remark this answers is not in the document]
rem1776: It's private by default so it shouldn't be used

data_override/data_override.F90
```diff
@@ -230,7 +230,7 @@ subroutine data_override_init(Atm_domain_in, Ocean_domain_in, Ice_domain_in, Lan
     call write_version_number("DATA_OVERRIDE_MOD", version)
 
 !  Initialize user-provided data table
-    default_table%gridname = 'none'
+    default_table%gridname = 'non'
```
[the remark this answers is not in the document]
rem1776: This was to prevent a warning from the string getting cut off, since gridname is only 3 characters long

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -26,7 +26,6 @@
 !> @addtogroup fms_netcdf_domain_io_mod
 !> @{
 module fms_netcdf_domain_io_mod
-use, intrinsic :: iso_fortran_env
```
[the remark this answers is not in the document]
uramirez8707: These can be removes. They were left over from when [we switch over from the intrinsic types `real32` instead of the portables types from platform.mod](https://github.com/NOAA-GFDL/FMS/issues/674)

mpp/mpp_efp.F90
```diff
@@ -257,8 +257,8 @@ function mpp_reproducing_sum_r4_2d(array, isr, ier, jsr, jer, EFP_sum, reproduci
 
   array_r8 = array
 
-  sum = mpp_reproducing_sum_r8_2d(array_r8, isr, ier, jsr, jer, EFP_sum, reproducing, &
-                            overflow_check, err)
+  sum = int(mpp_reproducing_sum_r8_2d(array_r8, isr, ier, jsr, jer, EFP_sum, reproducing, &
+                            overflow_check, err), r4_kind)
```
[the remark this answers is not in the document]
rem1776: the result sum is r4 so this just makes the cast explicit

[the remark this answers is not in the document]
bensonr: Yes - but you are casting to an "int"

mpp/include/system_clock.h
```diff
@@ -33,13 +33,13 @@ subroutine system_clock_mpi( count, count_rate, count_max )
       if(first_call_system_clock_mpi)then
          first_call_system_clock_mpi=.false.
          mpi_count0 = MPI_WTime()
-         mpi_tick_rate = 1.d0/MPI_WTick()
+         mpi_tick_rate = real(1.d0/MPI_WTick())
```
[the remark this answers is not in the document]
rem1776: The double(1.d0) makes the result quad precision if compiled with 8 byte reals so this just makes the cast explicit

mpp/mpp_efp.F90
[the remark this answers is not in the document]
rem1776: My bad! That should be real, will be fixed in the next commit

rem1776:
> You modified the mpp code to add a MPP_LOGICAL_TYPE and have it set a value to either .false. or 0 when the macro was expanded in a a few of the mpp routines. I'm not sure I understand why you did that - especially since it now has code taking loc(.false.)?

It was just to initialize the data before use, needed the macro to be able to initialize to a logical if needed. Would the address from LOC change from that? I thought it would stay the same just hold the initial values instead.

amip_interp/amip_interp.F90
```diff
@@ -893,7 +892,7 @@ subroutine amip_interp_init()
 
     tice_crit_k = tice_crit
     if ( tice_crit_k < 200. ) tice_crit_k = tice_crit_k + TFREEZE
-    ice_crit = nint((tice_crit_k-TFREEZE)*100.)
+    ice_crit = int((tice_crit_k-TFREEZE)*100., I2_KIND)
```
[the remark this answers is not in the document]
rem1776: fixed in 2a7946f158d2786eb22ad01976b7165dd149676f

field_manager/field_manager.F90
```diff
@@ -4714,7 +4699,6 @@ subroutine initialize  !{
 !+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 !        local variables
 !+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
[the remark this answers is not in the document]
rem1776: fixed in 2a7946f158d2786eb22ad01976b7165dd149676f

mpp/mpp_domains.F90
```diff
@@ -1783,7 +1776,7 @@ module mpp_domains_mod
   interface mpp_global_field_ad
      module procedure mpp_global_field2D_r8_2d_ad
      module procedure mpp_global_field2D_r8_3d_ad
-     module procedure mpp_global_field2D_r8_4d_ad
+     module procedure mpp_global_field2d_r8_4d_ad
```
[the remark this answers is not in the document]
rem1776: fixed in 2a7946f158d2786eb22ad01976b7165dd149676f

random_numbers/random_numbers.F90
rem1776: Yeah that'll just cast to default so it'll always be the same kind

exchange/xgrid.F90
```diff
@@ -464,8 +460,8 @@ module xgrid_mod
 ! Include variable "version" to be written to log file.
 #include<file_version.h>
 
- real(r8_kind), parameter                              :: EPS = 1.0d-10
- real(r8_kind), parameter                              :: LARGE_NUMBER = 1.d20
+ real(r8_kind), parameter                              :: EPS = 1.0e-10_r8_kind
+ real(r8_kind), parameter                              :: LARGE_NUMBER = 1.e20_r8_kind
```
thomas-robinson, line 464: Is this "more correct" than using `d`?

exchange/xgrid.F90
```diff
@@ -3272,6 +3262,7 @@ subroutine get_side1_from_xgrid(d, grid_id, x, xmap, complete)
   integer(i8_kind), dimension(MAX_FIELDS), save :: d_addrs=-9999
   integer(i8_kind), dimension(MAX_FIELDS), save :: x_addrs=-9999
 
+  d = 0
```
thomas-robinson, line 3265: 2 Things here:

1. This seems very dangerous.  The variable comment for `d` says "recieved xgrid data".  Are you sure that the undefined behavior of the compilers was to return `d` as all 0?  I think a better idea would be to change `d` to `intent (inout)` or just `intent(in)` because it's never modified.  It's difficult to tell what the compiler actually does when this sort of thing is happening.
2. If `d` is a real, it should be
``` fortran
d = 0.
```

exchange/xgrid.F90
```diff
@@ -4752,6 +4742,7 @@ subroutine get_side1_from_xgrid_ug(d, grid_id, x, xmap, complete)
...
```
thomas-robinson, line 4745: Same comment as above

mosaic2/grid2.F90
```diff
@@ -368,7 +368,7 @@ subroutine get_grid_cell_area_SG(component, tile, cellarea, domain)
                 'Illegal component name "'//trim(component)//'": must be one of ATM, LND, or OCN')
         end select
         ! convert area to m2
-        cellarea = cellarea*4.*PI*radius**2
+        cellarea = real(cellarea*4.*PI*radius**2, r4_kind)
```
thomas-robinson, line 371:
```fortran
cellarea = real(cellarea*4.*PI*radius**2., r4_kind)
```
You need a `2.` instead of `2` to ensure it is a real

mosaic2/grid2.F90
```diff
@@ -600,7 +600,7 @@ subroutine get_grid_comp_area_SG(component,tile,area,domain)
         deallocate(nest_tile_name)
      end select ! version
      ! convert area to m2
-     area = area*4.*PI*radius**2
+     area = real(area*4.*PI*radius**2, r4_kind)
```
thomas-robinson, line 603:
``` fortran
cellarea = real(cellarea*4.*PI*radius**2., r4_kind)

```

mpp/include/mpp_update_domains2D.h
```diff
@@ -227,6 +227,7 @@
       pointer( ptr_in,  field3D_in  )
       pointer( ptr_out, field3D_out )
 
+      field_out = 0
```
thomas-robinson, line 230: I don't know about this one...
1.  Are you sure this is the compiler behavior for all compilers?  Will this change answers?
2. This code is very confusing with the pointers.  Are you sure this isn't somehow getting set in mpp_redistribute?

mpp/include/mpp_update_domains2D.h
```diff
@@ -257,6 +258,7 @@
       MPP_TYPE_ :: d_type
       integer(i8_kind) :: floc_in, floc_out
 
+      field_out = 0
```
thomas-robinson, line 261: same as above

mpp/include/mpp_update_domains2D.h
```diff
@@ -344,6 +346,7 @@
       pointer( ptr_in,  field3D_in  )
       pointer( ptr_out, field3D_out )
 
+      field_out = 0
```
thomas-robinson, line 349: same as above

mpp/include/mpp_update_domains2D.h
```diff
@@ -367,6 +370,7 @@
...
```
thomas-robinson, line 373: same as above

mpp/mpp_efp.F90
```diff
@@ -257,8 +257,8 @@ function mpp_reproducing_sum_r4_2d(array, isr, ier, jsr, jer, EFP_sum, reproduci
 
   array_r8 = array
 
-  sum = mpp_reproducing_sum_r8_2d(array_r8, isr, ier, jsr, jer, EFP_sum, reproducing, &
-                            overflow_check, err)
+  sum = real(mpp_reproducing_sum_r8_2d(array_r8, isr, ier, jsr, jer, EFP_sum, reproducing, &
```
thomas-robinson, line 260: This needs extensive testing.

mpp/mpp_efp.F90
```diff
@@ -697,7 +697,7 @@ subroutine mpp_efp_list_sum_across_PEs(EFPs, nval, errors)
     do n=1,NUMINT ; EFPs(i)%v(n) = ints(n,i) ; enddo
     if (present(errors)) errors(i) = overflow_error
     if (overflow_error) then
-      write (mesg,'("mpp_efp_list_sum_across_PEs error at ",i6," val was ",ES12.6, ", prec_error = ",ES12.6)') &
+      write (mesg,'("mpp_efp_list_sum_across_PEs error at ",i6," val was ",ES13.6, ", prec_error = ",ES13.6)') &
```
thomas-robinson, line 700: Is this a change to an error message?

mpp/mpp_utilities.F90
```diff
@@ -29,6 +29,7 @@
 !> @{
 module mpp_utilities_mod
 
+implicit none
```
thomas-robinson, line 32: OOOOOHHHHHHHH BOY!!!!!

time_manager/time_manager.F90
```diff
@@ -1193,8 +1193,8 @@ function time_type_to_real(time)
 
 if(.not.module_is_initialized) call time_manager_init
 
-time_type_to_real = dble(time%days) * 86400.d0 + dble(time%seconds) + &
-     dble(time%ticks)/dble(ticks_per_second)
+time_type_to_real = real(dble(time%days) * 86400.d0 + dble(time%seconds) + &
+     dble(time%ticks)/dble(ticks_per_second), kind=r8_kind)
```
thomas-robinson, line 1197: Did this create a warning?  Everything is a double?

time_manager/time_manager.F90
```diff
@@ -2941,7 +2936,7 @@ function length_of_year_julian()
 
 type(time_type) :: length_of_year_julian
 
-length_of_year_julian = set_time((24 / 4) * 60 * 60, 365)
+length_of_year_julian = set_time(21600, 365)
```
thomas-robinson, line 2939: Why is this hardcoded/changed?

thomas-robinson (changes requested):
Setting `intent(out)` variables to 0 comes with an assumption of the compiler's behavior.  This needs to be discussed and investigated.

mosaic2/grid2.F90
bensonr: The 2 is an exponent, so making it a real could change answers due to f-p representations for "2" vs "2."

mosaic2/grid2.F90
bensonr: The 2 is an exponent, so making it a real could change answers due to f-p representations for "2" vs "2."

mpp/mpp_efp.F90
bensonr: The return from the mpp_reproducing_sum is a real(kind=8) and sum is defined as a real(kind=4).  With this addition, the cast is no longer implicit.

diag_manager/diag_output.F90
```diff
@@ -477,16 +475,15 @@ FUNCTION write_field_meta_data ( file_unit, name, axes, units, long_name, range,
 
     logical :: is_time_bounds !< Flag indicating if the variable is time_bounds
     CHARACTER(len=256) :: standard_name2
-    CHARACTER(len=1280) :: att_str
     TYPE(diag_fieldtype) :: Field
     LOGICAL :: coord_present
-    CHARACTER(len=40) :: aux_axes(SIZE(axes))
```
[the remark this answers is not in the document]
rem1776: Yeah it gets assigned to a larger string so this prevents it from getting cut off

exchange/xgrid.F90
rem1776: I think so, `d` will double whatever the current default real precision is so it could give you 16 byte reals, this will always be 8 bytes.

bensonr: Can you give us a reference to this as it has not been my experience.

rem1776: I had thought that from testing it before, but I just realized that it's actually compiler specific. I just tested it with:
```
print *, sizeof(1.e0), sizeof(1.d0)
```
With default 8 byte reals, GCC prints 8 and 16 and intel prints 8 and 8.

I guess this should still use the `_r8_kind` because that'll always be 8.

exchange/xgrid.F90
rem1776: I believe it's saying received because this function is receiving it. It's address gets sent to another routine where the data is set.

I think any compiler can return d with whatever happens to be in the address at the time (unless compiler flags are used to set all new values to 0) so this just sets it to 0 to make sure there's no random undefined values. I did this because it came up as an uninitialized warning but I made a similar change in #848, with gnu the undefined logical values were making the code fail so it needed to be initialized.

mpp/include/mpp_update_domains2D.h
rem1776: It shouldn't change answers unless the answers are dependent on uninitialized values. It most likely is getting set in mpp_redistribute, it's just good practice to set to some initial value and gets rid of the warnings.

time_manager/time_manager.F90
rem1776: double doesn't necessarily mean r8_kind so this had a warning but i think you'd only get it with gnu

time_manager/time_manager.F90
```diff
@@ -2927,11 +2925,8 @@ end function length_of_year_thirty
 function length_of_year_gregorian()
 
 type(time_type) :: length_of_year_gregorian
-integer :: days, seconds
 
-days = days_in_400_year_period / 400
-seconds = 86400*(days_in_400_year_period/400. - days)
-length_of_year_gregorian = set_time(seconds, days)
+length_of_year_gregorian = set_time(20952, 365)
```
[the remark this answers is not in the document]
rem1776: just thought it made more sense than doing useless calculations when we already know the results (same for below)

rem1776:
> Setting `intent(out)` variables to 0 comes with an assumption of the compiler's behavior. This needs to be discussed and investigated.

I mean with any compiler they'll be undefined by default. On intel(maybe gnu too) you can use flags to change the behavior to set uninitialized values to zero, which would make this redundant but this just makes sure the output data is 'clean' before values are set.

mpp/mpp_efp.F90
rem1776: Yeah, format code was throwing a warning, this just changes how the number is printed

rem1776:
@thomas-robinson @bensonr Updated this with fixes mainly for the initialization macro and int/real `0`'s. I should have covered all the comments, let me know if any other changes need to go in as well.

mpp/include/mpp_update_domains2D.h
thomas-robinson: You're right, this code *should have* initialized `field_out` because it is an `intent(out)` variable.  I completely agree with everything you said, but this is going to need extensive testing to make sure it's not changing answers.  I wouldn't be surprised if there was some code somewhere using the uninitialized values, and initializing with 0 changes the answers.  It's done in multiple places, which gives multiple chances for answer changes.

mpp/mpp_efp.F90
thomas-robinson: Will removing the implicit cast change answers/rounding though? Is the behavior of the explicit cast the same as the behavior of the implicit cast?  Can we guarantee the behavior between compilers?  Is it part of the standard? 
I have a repressed memory of @menzel-gfdl and I "fixing" something similar to this involving `mpp_reproducing_sum`, and it changed answers when variables were being explicitly cast.

time_manager/time_manager.F90
thomas-robinson: Leave the math in a comment at least so we know where these numbers come from.

thomas-robinson (changes requested):

exchange/xgrid.F90
thomas-robinson: So if you do this, then it's going to change the precision and change answers?

rem1776: regardless, it will be a 8bit real. With intel theres no change. With gnu it'll be implicitly casted so there shouldn't be a change, at least there wasn't when i tested it.

thomas-robinson (approved):
Let's test this beast

fms2_io/netcdf_io.F90
```diff
@@ -2190,11 +2189,13 @@ subroutine read_restart_bc(fileobj, unlim_dim_level)
...
                                 fileobj%restart_vars(i)%data2d, &
-                                fileobj%restart_vars(i)%bc_info)
+                                fileobj%restart_vars(i)%bc_info, &
+                                unlim_dim_level = unlim_dim_level)
     else if (associated(fileobj%restart_vars(i)%data3d)) then
        call scatter_data_bc (fileobj, fileobj%restart_vars(i)%varname, &
                                 fileobj%restart_vars(i)%data3d, &
-                                fileobj%restart_vars(i)%bc_info)
+                                fileobj%restart_vars(i)%bc_info, &
+                                unlim_dim_level = unlim_dim_level)
```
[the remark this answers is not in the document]
rem1776: @uramirez8707 Could you take a look at this before it goes in? It adds that optional variable to the scatter call that was otherwise unused

horiz_interp/horiz_interp_type.F90
```diff
@@ -169,7 +169,7 @@ subroutine stats ( dat, low, high, avg, miss, missing_value, mask )
       ! Force use of "scalar", integer pointer mpp interface
       call mpp_send(buffer_real(1),plen=3,to_pe=root_pe, tag=COMM_TAG_1)
       buffer_int(1) = miss
-      buffer_int(2) = npts
+      buffer_int(2) = int(npts)
```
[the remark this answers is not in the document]
rem1776: this was fixed with 2a7946f1

mosaic/grid.F90
```diff
@@ -952,7 +950,7 @@ subroutine define_cube_mosaic ( component, domain, layout, halo, maskmap )
 
   ! ---- local vars
   character(len=MAX_NAME) :: varname
-  character(len=MAX_FILE) :: mosaic_file
+  character(len=MAX_FILE + len(grid_dir)) :: mosaic_file
```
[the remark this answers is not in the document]
rem1776: Yeah this was to append the grid_dir string below

fms2_io/netcdf_io.F90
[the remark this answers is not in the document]
uramirez8707: I think this should be okay.

rem1776:
Looks like this is good to go without another merge

horiz_interp/horiz_interp_type.F90
```diff
@@ -117,8 +117,8 @@ subroutine stats ( dat, low, high, avg, miss, missing_value, mask )
  real, intent(in), optional :: missing_value
  real,    intent(in), optional :: mask(:,:)
 
- real :: dsum, npts, buffer_real(3)
- integer :: pe, root_pe, npes, p, buffer_int(2)
+ real :: dsum, buffer_real(3)
+ integer :: pe, root_pe, npes, p, buffer_int(2), npts
```
bensonr, line 121: Is there a chance npts can overflow as an integer here?

bensonr (approved):
While I have one new comment, I'll still allow the PR to go forward.

horiz_interp/horiz_interp_type.F90
rem1776: Hard to say, we can keep it in mind if there's any issues with the tag.

PR #866, opened 2021-12-06
Dm diag_yaml

uramirez8707:
**Description**

- Adds two public subroutines to `fms_diag_yaml_mod`, `diag_yaml_object_init` which fills in the diag_yaml_obj and `diag_yaml_object_end`, which destroys the diag_yaml_obj. They are called in diag_manager_init and diag_manager_end respectively. 
- Removes any c structs that were meant to hold the diag_table.yaml file and made the corresponding changes to the Makefile.am
- All the code in `fms_diag_yaml_mod` is hidden inside the -Duse_yaml macro because yaml_parser_mod won't compile without it

Fixes # (issue)

Still to do:
- Error checking: check that diag_table entries are valid?
- Test diag_yaml_object_init

**How Has This Been Tested?**
`make check` still passes with intel/gnu
`make check` passes with --with-yaml on skylake with intel

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_yaml_object.F90
```diff
@@ -34,31 +34,13 @@ module fms_diag_yaml_object_mod
...
-                                      !! this field is present.
-    character (kind=c_char) :: file_duration_units (3)!< The file duration units
-end type diag_yaml_files_struct
+
+type sub_region_type
+     character (len=:), allocatable :: grid_type !< Flag indicating the type of region,
+                                                      !!acceptable values are "latlon" and "index"
+     real :: lat_lon_region (NUM_REGION_ARRAY) !< Bounds of the regional section to capture if in "latlon" mode
+     integer :: index_region (NUM_REGION_ARRAY) !< Bounds of the regional section to capture if in "index" mode
+end type sub_region_type
```
bensonr, line 43: For the sub-regions, I want to ensure I understand the intent.  This storage object will accommodate the use of four pairs of lat-lon values or i-j indexes.  What I don't see is some way of constraining or limiting the vertical extent nor specifying the particular tile for the index option.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+  character(len=255) :: buffer !< String buffer to read in to
+
+  call get_value_from_key(file_id, par_id, trim(key_name), buffer, is_optional= is_optional)
+  allocate(character(len=len_trim(buffer)) :: value_name)
+  value_name = trim(buffer)
+
+end subroutine diag_get_value_from_key
+
+!< @brief gets the lat/lon of the sub region to use in a diag_table yaml
+subroutine get_region_latlon(diag_yaml_id, region_id, lat_lon_region)
```
bensonr, line 271: Do we want this to be subregion?

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+  call get_value_from_key(diag_yaml_id, region_id, "yend", lat_lon_region(4), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "zbegin", lat_lon_region(5), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "zend", lat_lon_region(6), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "ubegin", lat_lon_region(7), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "uend", lat_lon_region(8), is_optional=.true.)
+
+end subroutine get_region_latlon
+
+!< @brief gets the indexes of the sub region to use in a diag_table yaml
+subroutine get_region_index(diag_yaml_id, region_id, index_region)
```
bensonr, line 288: Do we want this to be subregion?

bensonr (commented):
There was a discussion on renaming of region to sub-region.  I see much of the code still uses the region nomenclature with a smattering of sub-region.  For consistency sake we should refer to it all the same.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+  integer, intent(out) :: index_region (8) !< sub region
+
+  call get_value_from_key(diag_yaml_id, region_id, "xbegin", index_region(1), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "xend", index_region(2), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "ybegin", index_region(3), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "yend", index_region(4), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "zbegin", index_region(5), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "zend", index_region(6), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "ubegin", index_region(7), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "uend", index_region(8), is_optional=.true.)
```
thomas-robinson, line 300: What is `u` referring to?  Is this documented somewhere and I missed it?

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+!< @brief gets the lat/lon of the sub region to use in a diag_table yaml
+subroutine get_region_latlon(diag_yaml_id, region_id, lat_lon_region)
+  integer, intent(in)  :: diag_yaml_id       !< Id of the diag_table yaml file
+  integer, intent(in)  :: region_id          !< Id of the region block to read from
+  real,    intent(out) :: lat_lon_region (8) !< sub region
+
+  call get_value_from_key(diag_yaml_id, region_id, "xbegin", lat_lon_region(1), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "xend", lat_lon_region(2), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "ybegin", lat_lon_region(3), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "yend", lat_lon_region(4), is_optional=.true.)
```
thomas-robinson, line 279: Latitude is the Y direction.  Longitude is the X direction.  This labeling is misleading.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+  call get_value_from_key(diag_yaml_id, region_id, "zbegin", lat_lon_region(5), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "zend", lat_lon_region(6), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "ubegin", lat_lon_region(7), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, region_id, "uend", lat_lon_region(8), is_optional=.true.)
```
thomas-robinson, line 283: What is `u`?  Wind?

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+
+  do i = 1, size(diag_yaml%diag_files, 1)
+    if(allocated(diag_yaml%diag_files(i)%file_global_meta)) deallocate(diag_yaml%diag_files(i)%file_global_meta)
+  enddo
+  if(allocated(diag_yaml%diag_files)) deallocate(diag_yaml%diag_files)
+
+  do i = 1, size(diag_yaml%diag_fields, 1)
+    if(allocated(diag_yaml%diag_fields(i)%var_attributes)) deallocate(diag_yaml%diag_fields(i)%var_attributes)
+  enddo
+  if(allocated(diag_yaml%diag_fields)) deallocate(diag_yaml%diag_fields)
```
thomas-robinson, line 143: This looks like work that could be shared between 2 threads because one loop doesn't depend on the other.

thomas-robinson (changes requested):

diag_manager/fms_diag_yaml.F90
uramirez8707: u is suppose to be the fourth dimension (i.e spectral dimension, species or whatever). I wasn't sure what naming convention to use here.

uramirez8707: Maybe instead of "{x/y/z/u}_begin" and "{x/y/z/u}_end", we can do: "dim{1,2,3,4}_begin" and "dim{1,2,3,4}_end" which is more general?

diag_manager/fms_diag_yaml_object.F90
uramirez8707: > What I don't see is some way of constraining or limiting the vertical extent

I am not sure I understand what you mean. The `*_region` arrays store the bounds for all four of the possible dimensions: [dim1_min, dim1_max, dim2_min, dim2_max, dim3_min, dim3_max, dim4_min, dim4_max].

> specifying the particular tile for the index option.

Yes, I forgot to add the "tile". It will be a required key if using the "index" approach. This will be added in the next commit.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
- contains
- procedure :: title => get_title        !< Returns the title
- procedure :: basedate => get_basedate  !< Returns the basedate array
+  character(len=:), allocatable, private :: diag_title                   !< Experiment name
+  integer, private, dimension (basedate_size) :: diag_basedate           !< basedate array
+  type(diag_yaml_files_type), allocatable, private, dimension (:) :: diag_files!< History file info
+  type(diag_yaml_files_var_type), allocatable, private, dimension (:) :: diag_fields !< Diag fields info
+  contains
+  procedure :: title => get_title        !< Returns the title
+  procedure :: basedate => get_basedate  !< Returns the basedate array
```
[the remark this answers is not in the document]
uramirez8707: ooops

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,324 @@
...
+  var_count = 0
+  do i = 1, nfiles
+    call diag_yaml_files_obj_init(diag_yaml%diag_files(i))
+    call fill_in_diag_files(diag_yaml_id, file_ids(i), diag_yaml%diag_files(i))
+
+    nvars = 0
+    nvars = get_num_blocks(diag_yaml_id, "varlist", parent_block_id=file_ids(i))
+    allocate(var_ids(nvars))
+    call get_block_ids(diag_yaml_id, "varlist", var_ids, parent_block_id=file_ids(i))
+    do j = 1, nvars
```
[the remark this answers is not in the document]
uramirez8707: I think it is not thread safe. `var_count` is changed every iteration.

diag_manager/fms_diag_yaml_object.F90
bensonr: This is where confusion comes in on how a sub-region is being defined.  I know in the past we've always defined the sub-region to have sides of constant lat or lon.  I saw the eight entries here and thought maybe we were expanding the capability to create somewhat irregular sub-regiosn denoted by either (i,j) or (lat,lon) for each fo the corners of the sub-region se, sw, ne, nw.

diag_manager/fms_diag_yaml.F90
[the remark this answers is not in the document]
thomas-robinson: `nvars` is set outside of this loop.  That shouldn't affect the loop.

If you made `nvars` (and `var_ids`) private, then you could parallelize the outer loop.  That might make more sense.  Otherwise you would be spinning up threads in a loop which is inefficient.

diag_manager/fms_diag_yaml.F90
thomas-robinson: I think the more general approach is a good idea.  If you document what u is, or name that `other_dim` or something like that, it would also be fine.  Or you could document that `u` is a utility 4th dimension.  Either way is up to you.

diag_manager/fms_diag_yaml.F90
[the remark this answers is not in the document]
uramirez8707: [`var_count` is changed inside the second do loop](https://github.com/NOAA-GFDL/FMS/blob/01f039fd0963ae0ae3e95eaa027c15e10584a640/diag_manager/fms_diag_yaml.F90#L122) and it is what it is used to index the `diag_yaml%diag_fields`. If multiple threads are changing it, I think it could lead to a race condition?

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -34,31 +34,13 @@ module fms_diag_yaml_object_mod
...
-                                      !! NOTE: The file_duration_units field must also be present if
-                                      !! this field is present.
-    character (kind=c_char) :: file_duration_units (3)!< The file duration units
-end type diag_yaml_files_struct
+
+type sub_region_type
+     character (len=:), allocatable :: grid_type !< Flag indicating the type of region,
+                                                      !!acceptable values are "latlon" and "index"
+     real :: lat_lon_region (NUM_REGION_ARRAY) !< Bounds of the regional section to capture if in "latlon" mode
+     integer :: index_region (NUM_REGION_ARRAY) !< Bounds of the regional section to capture if in "index" mode
```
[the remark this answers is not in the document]
uramirez8707: I don't understand. How will the code know which sub_region_type is inside in  each `diag_yaml%diag_files(:)` type?

parser/yaml_parser.F90
```diff
@@ -281,7 +281,7 @@ subroutine get_value_from_key_0d(file_id, block_id, key_name, key_value, is_opti
           read(buffer,*, iostat=err_unit) key_value
           if (err_unit .ne. 0) call mpp_error(FATAL, "Key:"//trim(key_name)//" Error converting '"//trim(buffer)//"' to r8")
        type is (character(len=*))
-          key_value = buffer
+          key_value = trim(buffer)
```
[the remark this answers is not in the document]
uramirez8707: This in the parser, so it has to be done on main too. 
I added this check on main #868.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  integer              :: diag_yaml_id !< Id for the diag_table yaml
+  integer              :: nfiles       !< Number of files in the diag_table yaml
+  integer, allocatable :: file_ids(:)  !< Ids of the files in the diag_table yaml
+  integer              :: i, j         !< For do loops
+  integer              :: total_nvars  !< The total number of variables in the diag_table yaml
+  integer              :: var_count    !< The current number of variables added to the diag_yaml obj
+  integer              :: nvars        !< The number of variables in the current file
+  integer, allocatable :: var_ids(:)   !< Ids of the variables in diag_table yaml
+
+  diag_yaml_id = open_and_parse_file("diag_table.yaml")
```
thomas-robinson, line 100: can we rename this to diag_yaml or diag.yaml?  I don't want it to get confsed with the diag_table and its format

thomas-robinson (changes requested):
Rename the yaml file that is used for diagnostic input without  "*table*" in the name

wrongkindofdoctor (approved):

bensonr (approved):

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim4_end", index_sub_region(8), is_optional=.true.)
+
+end subroutine get_sub_region_index
+
+!< @brief gets the total number of variables in the diag_table yaml file
+!< @return total number of variables
+function get_total_num_vars(diag_yaml_id, file_ids) &
+result(total_nvars)
+
+  integer, intent(in) :: diag_yaml_id   !< Id for the diag_table yaml
```
thomas-robinson, line 313: I think you want this to be `diag_file_id` or `history_file_id`.  I thought this was the fortran unit number for the diag_table.yaml file.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim2_end", lat_lon_sub_region(4), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim3_begin", lat_lon_sub_region(5), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim3_end", lat_lon_sub_region(6), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim4_begin", lat_lon_sub_region(7), is_optional=.true.)
+  call get_value_from_key(diag_yaml_id, sub_region_id, "dim4_end", lat_lon_sub_region(8), is_optional=.true.)
+
+end subroutine get_sub_region_latlon
+
+!< @brief gets the indexes of the sub region to use in a diag_table yaml
+subroutine get_sub_region_index(diag_yaml_id, sub_region_id, index_sub_region)
```
thomas-robinson, line 292: There's gotta be a cleaner way to do this than have 2 identical subroutines but one is for real and one is for integer.  Can you have the `out` variable be `class(*)` and let `get_value_from_key` figure it out?  Is that subroutine polymorphic?

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -33,32 +33,24 @@ module fms_diag_yaml_object_mod
...
+
+!> @brief type to hold the sub region information about a file
+type sub_region_type
+     character (len=:), allocatable :: grid_type !< Flag indicating the type of region,
+                                                 !! acceptable values are "latlon" and "index"
+     real :: lat_lon_sub_region (NUM_SUB_REGION_ARRAY)  !< Array that stores the grid point bounds for the sub region
+                                                        !! to use if grid_type is set to "latlon"
+                                                        !! [dim1_begin, dim1_end, dim2_begin, dim2_end,
+                                                        !!  dim3_begin, dim3_end, dim4_begin, dim4_end]
+     integer :: index_sub_region (NUM_SUB_REGION_ARRAY) !< Array that stores the index bounds for the sub region to
```
thomas-robinson, line 47: It seems like a waste of memory to have these both on the stack if only one is going to be used.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -67,38 +59,38 @@ module fms_diag_yaml_object_mod
      character (len=:), allocatable :: file_timeunit !< The unit of time
      character (len=:), allocatable :: file_unlimdim !< The name of the unlimited dimension
      logical :: file_write
-     character (len=:), allocatable :: string_file_write !< false if the user doesn’t want the file to be 
+     character (len=:), allocatable :: string_file_write !< false if the user doesn’t want the file to be
```
thomas-robinson, line 62: shouldn't this be a logical?  (I realize I probably did this)

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -67,38 +59,38 @@ module fms_diag_yaml_object_mod
...
+                                      !! frequency for creating new files.
                                       !! NOTE: The file_duration_units field must also be present if
                                       !! this field is present.
     character (len=:), allocatable :: file_duration_units !< The file duration units
-    character (len=:), dimension(:), allocatable :: file_varlist !< An array of variable names
+    character (len=MAX_STR_LEN), dimension(:), allocatable :: file_varlist !< An array of variable names
                                                              !! within a file
-    character (len=:), dimension(:,:), allocatable :: file_global_meta !< Array of key(dim=1) 
-                                                        !! and values(dim=2) to be added as global 
+    character (len=MAX_STR_LEN), dimension(:,:), allocatable :: file_global_meta !< Array of key(dim=1)
```
thomas-robinson, line 81: I think you should add a comment about why `len=MAX_STR_LEN` and not `:` for someone else who will have to deal with this in the future.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -137,10 +114,9 @@ module fms_diag_yaml_object_mod
      character (len=:), allocatable :: var_outname !< Name of the variable as written to the file
      character (len=:), allocatable :: var_longname !< Overwrites the long name of the variable
      character (len=:), allocatable :: var_units !< Overwrites the units
-     character (len=:), dimension (:), allocatable :: var_attributes !< Attributes to overwrite or
+     character (len=MAX_STR_LEN), dimension (:, :), allocatable :: var_attributes !< Attributes to overwrite or
```
thomas-robinson, line 117: I think you should add a comment about why `len=MAX_STR_LEN` and not `:` for someone else who will have to deal with this in the future.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -231,12 +174,12 @@ pure function get_file_realm(diag_files_obj) result (res)
  character (:), allocatable :: res !< What is returned
   res = diag_files_obj%file_realm
 end function get_file_realm
-!> \brief Inquiry for diag_files_obj%file_region
-pure function get_file_region (diag_files_obj) result (res)
+!> \brief Inquiry for diag_files_obj%file_subregion
```
thomas-robinson, line 177: Update based on style guide.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -231,12 +174,12 @@ pure function get_file_realm(diag_files_obj) result (res)
...
+pure function get_file_sub_region (diag_files_obj) result (res)
  class (diag_yaml_files_type), intent(in) :: diag_files_obj !< The object being inquiried
- real :: res (NUM_REGION_ARRAY) !< What is returned
-  res = diag_files_obj%file_region
-end function get_file_region
+ type(sub_region_type) :: res !< What is returned
+  res = diag_files_obj%file_sub_region
+end function get_file_sub_region
 !> \brief Inquiry for diag_files_obj%file_new_file_freq
```
thomas-robinson, line 183: Update based on style guide.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -252,7 +195,7 @@ end function get_file_new_file_freq_units
 !> \brief Inquiry for diag_files_obj%file_start_time
```
thomas-robinson, line 195: Update based on style guide.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -283,28 +226,7 @@ end function get_file_global_meta
...
- diag_var_obj%var_write= .true.
- if (diag_var_obj%string_var_write(1:1)=="f" .or. &
-     diag_var_obj%string_var_write(1:1)=="F") &
-        diag_var_obj%var_write= .false.
- deallocate (diag_var_obj%string_var_write)
-end subroutine copy_variable_struct_to_object
+
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 !!!!!!! YAML VAR INQUIRIES !!!!!!!
 !> \brief Inquiry for diag_yaml_files_var_obj%var_fname
```
thomas-robinson, line 232: Update based on style guide.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -364,10 +286,23 @@ end function get_var_write
 !> \brief Inquiry for diag_yaml_files_var_obj%var_attributes
```
thomas-robinson, line 286: Update based on style guide.

thomas-robinson (changes requested):

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -33,32 +33,24 @@ module fms_diag_yaml_object_mod
...
-                                      !! frequency for creating new files.  
-                                      !! NOTE: The file_duration_units field must also be present if
-                                      !! this field is present.
-    character (kind=c_char) :: file_duration_units (3)!< The file duration units
-end type diag_yaml_files_struct
+integer, parameter :: NUM_SUB_REGION_ARRAY = 8
+integer, parameter :: MAX_STR_LEN = 255
+
+!> @brief type to hold the sub region information about a file
+type sub_region_type
```
thomas-robinson, line 40: This should be `subRegion_type` or `subRegion_t`

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -33,32 +33,24 @@ module fms_diag_yaml_object_mod
...
+                                                        !!  dim3_begin, dim3_end, dim4_begin, dim4_end]
+     integer :: index_sub_region (NUM_SUB_REGION_ARRAY) !< Array that stores the index bounds for the sub region to
+                                                        !! to use if grid_type is set to "index"
+                                                        !! [dim1_begin, dim1_end, dim2_begin, dim2_end,
+                                                        !!  dim3_begin, dim3_end, dim4_begin, dim4_end]
+     integer :: tile !< Tile number of the sub region, required if using the "index" grid type
+
+end type sub_region_type
 
 type diag_yaml_files_type
```
thomas-robinson, line 55: This should be `diagYamlFiles_type` or `diagYamlFiles_t`

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -109,21 +101,6 @@ module fms_diag_yaml_object_mod
...
-     character (kind=c_char) :: var_module(20) !< The module that th variable is in
-     character (kind=c_char) :: var_skind(8) !< The type/kind of the variable
-     character (kind=c_char) :: var_write(5) !< false if the user doesn’t want the variable to be
-                                             !! written to the file (default: true).
-     character (kind=c_char) :: var_outname(20) !< Name of the variable as written to the file
-     character (kind=c_char) :: var_longname(100) !< Overwrites the long name of the variable
-     character (kind=c_char) :: var_units(10) !< Overwrites the units
-end type diag_yaml_files_var_struct
-
 type diag_yaml_files_var_type
```
thomas-robinson, line 104: This should be `diagYamlFilesVar_type` or `diagYamlFilesVar_t`

thomas-robinson (changes requested):
The type names need to follow the style guide https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#derived-types

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  integer :: nkeys         !< Number of key/value pairs of attributes
+  integer :: j             !< For do loops
+
+  integer, allocatable :: key_ids(:) !< Id of each attribute key/value pair
+
+  field%var_write = .true.
+  call diag_get_value_from_key(file_id, var_id, "var_name", field%var_varname)
+  call diag_get_value_from_key(file_id, var_id, "reduction", field%var_reduction)
+  call diag_get_value_from_key(file_id, var_id, "module", field%var_module)
+  call diag_get_value_from_key(file_id, var_id, "kind", field%var_skind)
```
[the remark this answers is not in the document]
uramirez8707: the diag_table equivalent, `pack` was not optional.

diag_manager/fms_diag_object.F90
```diff
@@ -196,11 +194,11 @@ subroutine fms_register_diag_field_obj &
...
-  if (is_field_type_null(dobj%diag_field)) then
-     dobj%diag_id = diag_not_found
-     dobj%vartype = diag_null
-     return
-  endif
+!  if (is_field_type_null(dobj%diag_field)) then
+!     dobj%diag_id = diag_not_found
+!     dobj%vartype = diag_null
+!     return
+!  endif
```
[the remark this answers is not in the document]
uramirez8707: I am not sure how this going to work, until we finish the register diag field subroutine.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  integer, allocatable :: key_ids(:) !< Id of the gloabl atttributes key/value pairs
+
+  call diag_get_value_from_key(diag_yaml_id, file_id, "file_name", fileobj%file_fname)
+  call diag_get_value_from_key(diag_yaml_id, file_id, "freq_units", fileobj%file_frequnit)
+  call get_value_from_key(diag_yaml_id, file_id, "freq", fileobj%file_freq)
+  call diag_get_value_from_key(diag_yaml_id, file_id, "unlimdim", fileobj%file_unlimdim)
+  call diag_get_value_from_key(diag_yaml_id, file_id, "time_units", fileobj%file_timeunit)
+  call diag_get_value_from_key(diag_yaml_id, file_id, "write_file", fileobj%string_file_write, is_optional=.true.)
+  if (fileobj%string_file_write .eq. "false") fileobj%file_write = .false.
+  !< Modeling realm?
```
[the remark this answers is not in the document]
uramirez8707: I am not sure what `realm` is? Is that not the same as `module`?

diag_manager/fms_diag_yaml_object.F90
uramirez8707: It is read in a string and used to set the logical `file_write` bellow.

diag_manager/fms_diag_yaml_object.F90
uramirez8707: It seems weird to have type names be camel case and variable names be snake case. We should stick with one ...

diag_manager/fms_diag_object.F90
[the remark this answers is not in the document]
thomas-robinson: OK, leave it for now.  You should comment that you are unsure how this will work for now and you might use this later

diag_manager/fms_diag_yaml.F90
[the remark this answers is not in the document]
thomas-robinson: `realm` is an idea from @ceblanton that defines which modeling realm (ie. atm, ocn, ice, lnd) the variable is from.  It is not the same as module.

diag_manager/fms_diag_yaml.F90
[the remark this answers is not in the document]
thomas-robinson: OK, going to resolve this then.

diag_manager/fms_diag_yaml_object.F90
thomas-robinson: Got it

diag_manager/fms_diag_yaml_object.F90
thomas-robinson: ::sigh::

diag_manager/fms_diag_yaml.F90
uramirez8707: https://github.com/NOAA-GFDL/FMS/pull/866/commits/c56ebeedcec7cb0eaa022248af71d77305585a80

thomas-robinson: This is much better!

diag_manager/fms_diag_yaml_object.F90
uramirez8707: You need to have both. There is no real way around it. It can be allocatable but still ...

diag_manager/fms_diag_yaml.F90
```diff
@@ -1,62 +1,328 @@
...
+  integer :: natt          !< Number of attributes in variable
+  integer :: var_att_id(1) !< Id of the variable attribute block
+  integer :: nkeys         !< Number of key/value pairs of attributes
+  integer :: j             !< For do loops
+
+  integer, allocatable :: key_ids(:) !< Id of each attribute key/value pair
+
+  field%var_write = .true.
+  call diag_get_value_from_key(file_id, var_id, "var_name", field%var_varname)
+  call diag_get_value_from_key(file_id, var_id, "reduction", field%var_reduction)
```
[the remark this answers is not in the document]
uramirez8707: @thomas-robinson I think the error checking + tests should be a separate PR. This PR is already huge and adding this will make it even bigger.

diag_manager/fms_diag_yaml.F90
thomas-robinson: For posterity, it was decided to use the `diag_table.yaml` name by the FMS group.

diag_manager/fms_diag_yaml_object.F90
thomas-robinson: Yes, you need both of them. But they should be `allocatable`, and only one should be allocated. Then you shouldn't need the `grid_type` variable because only one will be allocated (unless the string is used downstream for some reason).  
With this set up where they are both on the stack, you are just wasting memory.

diag_manager/fms_diag_yaml_object.F90
```diff
@@ -152,222 +128,206 @@ module fms_diag_yaml_object_mod
...
   res = diag_var_obj%var_write
 end function get_var_write
-!> \brief Inquiry for diag_yaml_files_var_obj%var_attributes
+!> @brief Inquiry for diag_yaml_files_var_obj%var_attributes
+!! @return var_attributes of a diag_yaml_files_var_obj
 pure function get_var_attributes(diag_var_obj) result (res)
- class (diag_yaml_files_var_type), intent(in) :: diag_var_obj !< The object being inquiried
- character (len=:), allocatable :: res (:) !< What is returned
+ class (diagYamlFilesVar_type), intent(in) :: diag_var_obj !< The object being inquiried
+ character (len=MAX_STR_LEN), allocatable :: res (:,:) !< What is returned
```
thomas-robinson, line 314: I think you should add a comment about why len=MAX_STR_LEN and not : for someone else who will have to deal with this in the future.

thomas-robinson (changes requested):
3 things, and they're all in diag_manager/fms_diag_yaml_object.F90
1. in `type subRegion_type` the arrays should be allocatable because this set up is wasting memory.  You may be able to remove `grid_type` if you are only allocating one of them.  Alternatively, you could also use two types that extend `type subRegion_type`, They would be named something like `type subRegionIndicies_type` and `type subRegionLatLon_type`, and each would have the correct variable defined on the stack like how you have it now.
2. Line 81 should have a comment about why the character length is set and not allocatable.  In the future, someone will look at this and change it to a `:` without any context for why it's set up like this.
3. Same as 2 but line 117.

thomas-robinson (approved):

PR #885, opened 2022-01-11
Dm get fns

### LINKED ISSUE #884: Adding `get` functions to the diag objects
All of the new diag_manager objects need to have `get_` functions to get the values in the object.
```fortran
type diag_type
     variable
 contains
     procedure get_variable
end type
```

thomas-robinson:
**Description**
- Uses a class(*) variable for missing_value and data_RANGE
- Fills in CMOR_MISSING_VALUE for missing_value if none is given
- Adds  rountines for fmsDiagObject_type

Fixes #884 

**How Has This Been Tested?**
`make` with intel compilers

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -36,7 +36,7 @@ type fmsDiagObject_type
      class(FmsNetcdfFile_t), dimension (:), pointer   :: fileob => NULL()  !< A pointer to all of the
                                                                            !! file objects for this variable
      character(len=:), allocatable, dimension(:)      :: metadata          !< metadata for the variable
-     logical, private                                 :: static            !< true is this is a static var
+     logical, allocatable, private                    :: static            !< true if this is a static var
      logical, allocatable, private                    :: registered        !< true when registered
      logical, allocatable, private                    :: mask_variant      !< If there is a mask variant
      logical, allocatable, private                    :: local             !< If the output is local
@@ -81,12 +81,14 @@ type fmsDiagObject_type
      procedure :: copy => copy_diag_obj
      procedure :: register => fms_register_diag_field_obj !! Merely initialize fields.
      procedure :: setID => set_diag_id
-     procedure :: is_registered => diag_ob_registered
      procedure :: set_type => set_vartype
      procedure :: vartype_inq => what_is_vartype
 ! Check functions
      procedure :: is_static => diag_obj_is_static
+     procedure :: is_registered => diag_ob_registered
      procedure :: is_registeredB => diag_obj_is_registered
+     procedure :: is_mask_variant => get_mask_variant
+     procedure :: is_local => get_local
 ! Get functions
      procedure :: get_diag_id => fms_diag_get_id
      procedure :: get_metadata
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -425,7 +425,7 @@ end function diag_obj_is_static
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 !! Get functions
 
-!> Gets metedata
+!> @brief Gets metedata
 !! @return metadata string array, or a single space if metadata is not allocated
 function get_metadata (obj) &
 result(rslt)
@@ -439,7 +439,7 @@ result(rslt)
        rslt = " "
      endif
 end function get_metadata
-!> Gets static
+!> @brief Gets static
 !! @return true if the variable is static
 function get_static (obj) &
 result(rslt)
@@ -447,7 +447,7 @@ result(rslt)
      logical :: rslt 
      rslt = obj%static
 end function get_static
-!> Gets regisetered
+!> @brief Gets regisetered
 !! @return true if the object is registered
 function get_registered (obj) &
 result(rslt)
@@ -455,7 +455,7 @@ result(rslt)
      logical :: rslt 
      rslt = obj%registered
 end function get_registered
-!> Gets mask variant
+!> @brief Gets mask variant
 !! @return true if there is a mask variant
 function get_mask_variant (obj) &
 result(rslt)
@@ -463,7 +463,7 @@ result(rslt)
      logical :: rslt 
      rslt = obj%mask_variant
 end function get_mask_variant
-!> Gets local
+!> @brief Gets local
 !! @return true if the variable is a local variable 
 function get_local (obj) &
 result(rslt)
@@ -471,7 +471,7 @@ result(rslt)
      logical :: rslt 
      rslt = obj%local
 end function get_local
-!> Gets initial time 
+!> @brief Gets initial time 
 !! @return the initial time
 !! TODO
 !function get_init_time (obj) &
@@ -480,7 +480,7 @@ end function get_local
 !     TYPE(time_type) :: rslt 
 !
 !end function get_init_time
-!> Gets vartype 
+!> @brief Gets vartype 
 !! @return  The integer related to the variable type
 function get_vartype (obj) &
 result(rslt)
@@ -488,7 +488,7 @@ result(rslt)
      integer :: rslt 
      rslt = obj%vartype
 end function get_vartype
-!> Gets varname
+!> @brief Gets varname
 !! @return the variable name
 function get_varname (obj) &
 result(rslt)
@@ -496,7 +496,7 @@ result(rslt)
      character(len=:), allocatable :: rslt 
      rslt = obj%varname
 end function get_varname
-!> Gets longname
+!> @brief Gets longname
 !! @return the variable long name or a single string if there is no long name
 function get_longname (obj) &
 result(rslt)
@@ -508,7 +508,7 @@ result(rslt)
        rslt = " "
      endif
 end function get_longname
-!> Gets standname
+!> @brief Gets standname
 !! @return the standard name
 function get_standname (obj) &
 result(rslt)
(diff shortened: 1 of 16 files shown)


diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +421,298 @@ function diag_obj_is_static (obj) result (rslt)
...
-
-function diag_obj_get_varname(obj) result (rslt)
-    class(fmsDiagObject_type), intent(in) :: obj
-    character(len=len(obj%varname)) :: rslt
-    rslt = obj%varname
-end function diag_obj_get_varname
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!! Get functions
+
+!> Gets metedata
```
uramirez8707, line 427: These need a `@brief `

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +421,298 @@ function diag_obj_is_static (obj) result (rslt)
...
+!! @return metadata string array, or a single space if metadata is not allocated
+function get_metadata (obj) result(rslt)
```
uramirez8707, line 429: I **think** `result(rslt)` should be its own line ???
https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#functions

uramirez8707 (changes requested):

diag_manager/fms_diag_object.F90
thomas-robinson: Added

diag_manager/fms_diag_object.F90
thomas-robinson: The style guide means that the result variable is declared on it's own line, not that a continuation is used.  I did it anyway too be consistent with your functions. I'm going to update the style guide too.

uramirez8707 (approved):

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+       allocate(character(len=(len(obj%metadata(1)))) :: rslt (size(obj%metadata)) )
+       rslt = obj%metadata
+     else
+       allocate(character(len=1) :: rslt(1:1))
+       rslt = " "
+     endif
+end function get_metadata
+!> @brief Gets static
+!! @return true if the variable is static
+function get_static (obj) &
```
ngs333, line 444: What you have works, but we should discuss if its a good idea to  follow a convention that adds 
a function ( doesn;t get rig of the get_satic, but adds another ) named ``` is_static``` 
It is used like 
``` 
if (obj%is_static() ) then
```
In the end, waht you ahve passes the litmus test that these two statements  be
 equivalent (if  static had been public)
```
 s = obj%static
s = obj%get_static()
```
(this is because staic is another intrisic (logical) like int, real etc.

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+result(rslt)
+     class (fmsDiagObject_type), intent(in) :: obj !< diag object
+     logical :: rslt 
+     rslt = obj%static
+end function get_static
+!> @brief Gets regisetered
+!! @return true if the object is registered
+function get_registered (obj) &
```
ngs333, line 452: See get_static comment. Do we want a user to have ``` if (obj%is_registered() ``` ?

diag_manager/fms_diag_object.F90
```diff
@@ -35,7 +35,7 @@ module fms_diag_object_mod
      integer, allocatable, private                    :: diag_id           !< unique id for varable
      class(FmsNetcdfFile_t), dimension (:), pointer   :: fileob => NULL()  !< A pointer to all of the
                                                                            !! file objects for this variable
-     character(len=:), allocatable, dimension(:)      :: metadata          !< metedata for the variable
+     character(len=:), allocatable, dimension(:)      :: metadata          !< metadata for the variable
      logical, private                                 :: static            !< true is this is a static var
      logical, allocatable, private                    :: registered        !< true when registered
```
ngs333, line 40: I think its less complex and more efficient if the intrinsic primitives (like logical, integer, real) are not allocatable
and have a default set in the type definiition. E.g. like this
``` 
logical, private    ::  registered => .false.
```
it would help to add functions like ``` logical is_registered()```  (see comments below). Probably,  type fms_dia_object
may nedd one or more initialize functions that the constructor calls.

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+result(rslt)
+     class (fmsDiagObject_type), intent(in) :: obj !< diag object
+     integer :: rslt 
+     rslt = obj%vartype
+end function get_vartype
+!> @brief Gets varname
+!! @return the variable name
+function get_varname (obj) &
+result(rslt)
+     class (fmsDiagObject_type), intent(in) :: obj !< diag object
```
ngs333, line 495: I guess in this case the getter litterally returns the member field (seems to pass the litmus test). But I will read (or as) what exatly does the assignemnt (```=```) do.  Hope we dont have to look at assemberr  :-)

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+     character(len=:), allocatable :: rslt 
+     rslt = obj%varname
+end function get_varname
+!> @brief Gets longname
+!! @return the variable long name or a single string if there is no long name
+function get_longname (obj) &
+result(rslt)
```
ngs333, line 502: And in this case we dont follow the litmus test with the ```else = " " ``` statement. Too confusing to not follow one rule for such getters. BTW, One possibility is again with the default values in the type definition. Where
fields like longName is initialized (with =>) to "" or " " or some class or saystem default.

diag_manager/fms_diag_object.F90
```diff
@@ -91,12 +84,36 @@ module fms_diag_object_mod
...
+     procedure :: get_longname
+     procedure :: get_standname
+     procedure :: get_units
+     procedure :: get_modname
+     procedure :: get_realm
+     procedure :: get_err_msg
+     procedure :: get_interp_method
+     procedure :: get_frequency
+     procedure :: get_output_units
+     procedure :: get_t
```
ngs333, line 108: Wat is "t" ? Is this a good name? Are there comments somewhere?

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+     character(len=:), allocatable :: rslt 
+     if (allocated(obj%interp_method)) then
+       rslt = obj%interp_method
+     else
+       rslt = " "
+     endif
+end function get_interp_method
+!> @brief Gets frequency
+!! @return the  frequency or DIAG_NULL if obj%frequency is not allocated
+function get_frequency (obj) &
```
ngs333, line 585: Does not return the frequency. It returns a copy.

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
+!> @brief Gets t
+!! @return t 
+function get_t (obj) &
+result(rslt)
+     class (fmsDiagObject_type), intent(in) :: obj !< diag object
+     integer :: rslt 
+     if (allocated(obj%t)) then
+       rslt = obj%t
+     else
+       rslt = -999
```
ngs333, line 620: Similar litmus test comments.

ngs333 (changes requested):
@thomas-robinson 
I added several comments but you may first want to read the longer one for 
function "get_metadata". If you want we can discuss these in a meeting, etc.

diag_manager/fms_diag_object.F90
```diff
@@ -396,17 +422,323 @@ function diag_obj_is_static (obj) result (rslt)
...
-function diag_obj_get_varname(obj) result (rslt)
-    class(fmsDiagObject_type), intent(in) :: obj
-    character(len=len(obj%varname)) :: rslt
-    rslt = obj%varname
-end function diag_obj_get_varname
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!! Get functions
+
+!> @brief Gets metedata
+!! @return metadata string array, or a single space if metadata is not allocated
```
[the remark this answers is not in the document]
thomas-robinson: As a group, we decided to return a copy, and all of the comments of the `get_` functions have been updated to reflect that we are returning a copy.  

We can't return an unallocated.  Returning an unallocated will at best cause a crash and at worst not crash and do some sort of compiler undefined behavior.

We can't initialize the strings because we would have to limit their size if we did.  These have to be allocatable.

The `has_` functions are coming in a future separate PR to not make this PR too large.

diag_manager/fms_diag_object.F90
thomas-robinson: Yes, we WANT the user to check if something is allocated or registered.  What happens if they don't?

diag_manager/fms_diag_object.F90
thomas-robinson: I added a varaible called `diag_null_sring` that is a parameter and has the value " ".  This can be used in if statments
```fortran
if (obj%get_longname() .ne. diag_null_string) longname = obj%get_longname()
```

ngs333 (approved):

PR #887, opened 2022-01-13
Moves fms_diag_yaml_obj to fms_diag_yaml

uramirez8707:
**Description**
Moves fms_diag_yaml_obj to fms_diag_yaml
Sets the variables of diag_files and diag_fields types to private
Uses DIAG_NULL to initialize optional values. 

Fixes # (issue)

**How Has This Been Tested?**
`make passes` with intel and gnu

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_yaml.F90
```diff
@@ -507,6 +608,248 @@ pure function is_valid_time_units(time_units) &
     is_valid = .false.
   end select
 end function is_valid_time_units
+
+!!!!!!! YAML FILE INQUIRIES !!!!!!!
+!> @brief Inquiry for diag_files_obj%file_fname
+!! @return file_fname of a diag_yaml_file obj
```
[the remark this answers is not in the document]
ngs333: @thomas-robinson 
If all , or almost all getters,  return copies, then I think just easiest to say in a one sentence paragraph in the intro class description something like:
 "All getter functions (functions named get_x(), for memeber field named x) return copies of the member variables unless explicitly noted." 
Maybe we can make guidelines like this one?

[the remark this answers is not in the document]
thomas-robinson: OK, all of these functions return a copy, so I think it's safe to just put it in the documentation of the derived type.  @uramirez8707 can you add that to the documentation?

thomas-robinson (changes requested):
Update the type documentation to reflect that all get functions/routines return a copy of the variable being requested.

diag_manager/fms_diag_yaml.F90
```diff
@@ -42,9 +40,112 @@ module fms_diag_yaml_mod
...
+
+!> @brief type to hold the diag_file information
+type diagYamlFiles_type
+  character (len=:), private, allocatable :: file_fname !< file name
+  character (len=:), private, allocatable :: file_frequnit !< the frequency unit
+  integer, private    :: file_freq !< the frequency of data
+  character (len=:), private, allocatable :: file_timeunit !< The unit of time
+  character (len=:), private, allocatable :: file_unlimdim !< The name of the unlimited dimension
+  logical, private :: file_write !< false if the user doesn't want to the file to be created
+  character (len=:), private, allocatable :: string_file_write !< false if the user doesn’t want the file to be
```
ngs333, line 74: Is a string version of file_write really needed? If so, then every place file_write is set, string_file_write needs also be set.
Might it be safer or easier to just use a funcion like : get_file_write_str() if (file_write .eq. .true) return 'true' else return 'false' ?

diag_manager/fms_diag_yaml.F90
```diff
@@ -42,9 +40,112 @@ module fms_diag_yaml_mod
...
+  integer, private :: file_new_file_freq !< Frequency for closing the existing file
+  character (len=:), private, allocatable :: file_new_file_freq_units !< Time units for creating a new file.
+                                                                      !! Required if “new_file_freq” used
+  character (len=:), private, allocatable :: file_start_time !< Time to start the file for the first time. Requires
+                                                             !! “new_file_freq”
+  integer, private :: file_duration !< How long the file should receive data after start time
+                                    !! in “file_duration_units”.  This optional field can only
+                                    !! be used if the start_time field is present.  If this field
+                                    !! is absent, then the file duration will be equal to the
+                                    !! frequency for creating new files.
```
ngs333, line 87: What does this really mean: "If this field is absent, then the <file duration> will be equal to the frequency for creating new files".  In particular, what does <file duration> mean. Is it a different field than file_duration. Seems like somewhere (maybe not necessarily in this module) there should be another function to encasulate the functionality. Like maybe caculate_actual_file_duration( diagYmlFile_type.file_duration, ... )  ?

ngs333 (approved):
Approving, but with comments and questions.

diag_manager/fms_diag_yaml.F90
uramirez8707: If the file_duration/file_duration_units is present (e.g 6 days) , diag_manager will write a new file every 6th day from the start time with data at whatever the new_file_frequency is. (e.g hourly data or whatever)

If the file_duration/file_duration_units is not present, diag_manager will write a new file at whatever the `new_file_frequency` is. 

This will probably change, but we are not there yet.

ngs333: I think that if a comments  switches from a description about the current class to something  that happens in  another class, then the comments should make that switch obvious. Perhaps change the comment to "If this field
is absent, then the diag_manager will use a file duration that will be equal to the frequency for creating new files." ?

thomas-robinson (approved):

Issue #878, opened 2022-01-05
crash in test_diag_object_container.F90


thomas-robinson:
**Describe the bug**
The test in test_diag_object_container.F90 give me a segmentation fault.  The traceback looks like this:
```
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source             
test_diag_object_  000000000043700A  Unknown               Unknown  Unknown
libpthread-2.28.s  000015506FD63D50  Unknown               Unknown  Unknown
libFMS.so.10.0.0   00001550749460AA  Unknown               Unknown  Unknown
libFMS.so.10.0.0   000015507494764C  Unknown               Unknown  Unknown
libFMS.so.10.0.0   000015507494237B  Unknown               Unknown  Unknown
test_diag_object_  0000000000411A5E  MAIN__                    102  test_diag_object_container.F90
test_diag_object_  000000000040E5A2  Unknown               Unknown  Unknown
libc-2.28.so       000015506F5AFCA3  __libc_start_main     Unknown  Unknown
test_diag_object_  000000000040E4AE  Unknown               Unknown  Unknown
FAIL test_diag_manager2.sh (exit status: 174)
```
I tried to set my stacksize to unlimited, but that didn't solve the issue

**To Reproduce**
On the GFDL lscamd50-d
```bash
cd $HOME
git clone git@github.com:NOAA-GFDL/FMS.git && cd FMS && git checkout dmUpdate
mkdir $HOME/build && cd $HOME/build
module load oneapi compiler-rt/2021.2.0 compiler/2021.2.0 mpi/2021.2.0 hdf5/1.12.0 libyaml/0.2.5 netcdf/4.8.0
../FMS/configure FC=mpiifort CC=mpiicc CPPFLAGS="`nc-config --cflags` `nc-config --libs` `nc-config --fflags` `nc-config --flibs` -O0 -traceback -g " LDFLAGS="`nc-config --cflags` `nc-config --libs` `nc-config --fflags` `nc-config --flibs`"
make check
```

**Expected behavior**
This should pass

**System Environment**
GFDL lscamd50-d

**Additional context**

thomas-robinson:
I also tried with these modules loaded and hit the crash
```
1) intel/2020_up3   2) impi/2020_up3   3) netcdf/4.8.0   4) hdf5/1.12.0
```

PR #888, opened 2022-01-14
diag_manager ontainer and linked list -test core dump fixes and more

ngs333:
**Description**
This PR fixes a core dump onserved when running the unit test for fms_diag_dlinked_list
type.  The head and tail sentinal nodes were changed to be pointers (like all other nodes) intead of
allocatable, this resulted on (on Intel compilers) for the constructor to behave as expected and not
deallocate the sentinal modes when it goes out of scope. The destructor was also slighly improved.
Intead of a default constructor, clients shoul use the new initialize() function (which is called once
in the diag manager module)

Additionally, numerous comments and minor miscellanous changes were made thought the code and 
unit tests.

Fixes # (878)
[crash in test_diag_object_container.F90 #878](https://github.com/NOAA-GFDL/FMS/issues/878)

**How Has This Been Tested?**
The unite tests have been run sucessfully and without failure may times with Intel 2021 compiler builds, and also with gnu 9.3 compier builds.

**Checklist:**
- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] I have made corresponding changes to the documentation
- [ X] My changes generate no new warnings
- [X ] Any dependent changes have been merged and published in downstream modules
- [X ] New check tests, if applicable, are included
- [ X] `make distcheck` passes


diag_manager/fms_diag_dlinked_list.F90
```diff
@@ -41,282 +41,297 @@
...
-    class(*), pointer :: data => null()        !< The data pointed to by the node.
-    type(FmsDlListNode_t), pointer :: next => null() !< A pointer to the previous node.
-    type(FmsDlListNode_t), pointer :: prev => null() !< A pointer to the next node.
-  end type FmsDlListNode_t
+   USE fms_mod, ONLY: error_mesg, FATAL, WARNING, NOTE
+   implicit none
+   !> The doubly-linked list node type.
+   type, public:: FmsDlListNode_t
+      private
+      class(*), pointer :: data => null()        !< The data pointed to by the node.
```
thomas-robinson, line 49: [You can't name this variable `data`](https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#general-1).  `DATA` is a reserved word in Fortran.

diag_manager/fms_diag_dlinked_list.F90
```diff
@@ -41,282 +41,297 @@
...
-    litr%current => fnd
-    litr%end => tnd
-  end function literator_constructor
+   !> @brief The list iterator constructor.
+   !! @return Returns a newly allocated list iterator.
+   function literator_constructor ( fnd, tnd ) result (litr)
+      type (FmsDlListNode_t), pointer  :: fnd
+      !< The sentinal (non-data) "first node" of the iterator will be fnd
+      type (FmsDlListNode_t), pointer  :: tnd
+      !< The sentinal (non-data) "last node" of the iterator will be tnd.
```
thomas-robinson, line 206: Will this style of commenting on the following line generate the documentation correctly?  I think these comments need to be on the same line as the variable, and you can use `!!` on the next line if there isnt enough space.

thomas-robinson: It looks like this is in a lot of places.

diag_manager/fms_diag_object_container.F90
```diff
@@ -56,13 +56,15 @@ MODULE fms_diag_object_container_mod
    !!
    type, public:: FmsDiagObjectContainer_t
    private
-      TYPE (FmsDlList_t), ALLOCATABLE :: the_linked_list !< This version based on the FDS linked_list.
+      TYPE (FmsDlList_t), pointer :: the_linked_list => null() !< This version based on the FDS linked_list.
```
thomas-robinson, line 59: This change makes me worried you are leaking memory

diag_manager/fms_diag_object_container.F90
```diff
@@ -202,10 +202,20 @@ end function diag_obj_iterator_constructor
    function diag_object_container_constructor () result (doc)
       type(FmsDiagObjectContainer_t), allocatable :: doc !< The resultant container.
       allocate(doc)
-      doc%the_linked_list = FmsDlList_t()
-      !! print * , "In DOC constructor"
+      allocate(doc%the_linked_list)
```
thomas-robinson, line 205: Does the destructor get called every time?  I think you were getting a seg fault because you weren't deallocating this variable.  If that's the case and you are allocating a pointer that's been allocated, then you are leaking memory.

thomas-robinson (changes requested):

diag_manager/fms_diag_dlinked_list.F90
ngs333: Data used as a type or type qualifier has a different meaning than used here. For a member variable name, its OK. The program works with both Intel and gcc compilers. Also see "Modern Fortran Explained, Incorporating Fortran 2018", Metcalf, pg 317 vs pg 154.

diag_manager/fms_diag_object_container.F90
ngs333: No memory leaks in this class. The unit test was checked with Inspector. The class has a procedure decorated with "final" which de-allocates what the class is responsible for.

diag_manager/fms_diag_dlinked_list.F90
thomas-robinson: OK, I guess I phrased this incorrectly because, yes, Fortran *allows* you to name variables after reserved words.  However, our [FMS style guide does not allow it](https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#general-1), and it specifically mentions `data` as a variable name that you shouldn't use in our code.  If you has a goto statement or a common black, I would ask you to remove them as well.

diag_manager/fms_diag_object_container.F90
thomas-robinson: So then why was it changed from `allocatable` to `pointer`?

diag_manager/fms_diag_object_container.F90
thomas-robinson: Can this be changed to 
```fortran
doc%the_linked_list => null
```
Are you setting this equal to something, or using it as a pointer?

ngs333: Before and now linked list class (and in the container class) , have a function decorated with the “final” keyword.  When class instance goes out of scope, the final decorated function cleans up any links (nodes) that remain.

Before,  linked list class had an OO-stype constructor (function with exactly the same name as the type) used in the unit test like this :

ll = FmsLlList_t()

What I coul see as I stepped though it in the debugger for Intel compiled versions, was that as soon as the RHS above finished executing, deallocation of anything that was “allocatable” in the linked list was automatically happening.  So for one instance created by constructor related functions, members head and tail were manually allocated, automatically deallocated when the instance went out of scope, and finally, with the “=” above, memberwise copied to ll.

In this version I got rid of the FmsLlList_t() constructor function. The final keywork is important, and the worked differently Intel vs gcc, so not a great loss. Now must explicitly do :
ll = allocate(FmsllList_t)
call ll%initialize()

For having the type FmsLlList_t() clean up after itself, users can just have its instance be allocatable.
In this case the final keyword decoration on function “destructor()” will make sure function 
clear() (or clear_all()) is called and all pointers data will be deallocated. If the user makes the class instance a pointer, then the user is responsible for calling function clear() when desired or before end of program execution.

diag_manager/fms_diag_dlinked_list.F90
ngs333: OK. Changed!

diag_manager/fms_diag_object_container.F90
ngs333: This one wasn't a completely necessary change, but for consistency sake: all links (nodes) between head and tail nodes were pointers anyways, so I made those two also be pointers.

diag_manager/fms_diag_object_container.F90
ngs333: Concerning your second question, I think yes. Do you mean change lines 204  - 205 from 
```
allocate(doc)
allocate(doc%the_linked_list)
```
to 
```
doc=>null()
allocate(doc)
allocate(doc%the_linked_list)
```

ngs333: or
```
doc=>null()
allocate(doc)
doc%the_linked_list => null()
allocate(doc%the_linked_list)
```

diag_manager/fms_diag_dlinked_list.F90
ngs333: I think its working. See attached image made with 
``` ../configure --enable-docs && ``
Screenshot from 2022-02-01 12-42-22
`

PR #901, opened 2022-02-07
diag_yaml update

### LINKED ISSUE #900: diag_yaml bugs
**Describe the bug**
1. `check_file_freq` crashes if the freq read in from a diag_table.yaml is less than 1 claiming that is not valid. However a frequency of 0 and -1 are allowed. 
https://github.com/NOAA-GFDL/FMS/blob/2c218d0aabd5340bde8f157d15ade6efd128e163/diag_manager/fms_diag_yaml.F90#L381-L382
https://github.com/NOAA-GFDL/FMS/blob/2c218d0aabd5340bde8f157d15ade6efd128e163/diag_manager/diag_table.F90#L79-L86

2. `check_field_reduction` tries to use specific indices to determine the reduction method. 
https://github.com/NOAA-GFDL/FMS/blob/2c218d0aabd5340bde8f157d15ade6efd128e163/diag_manager/fms_diag_yaml.F90#L471
However if your reduction method is something like "none" it will cause the crash:
``` 
forrtl: severe (408): fort: (4): Variable FIELD has substring ending point 7 which is greater than the variable length of 4
```

**To Reproduce**
Try to run diag_manager_init with a frequency of -1 or 0 and the code will crash. 
Try to run diag_manager_init with a reduction_method of "none" and the code will crash (in debug mode)

**Expected behavior**
-1 and 0 need to be accepted as valid frequencies. 
The reduction_method needs to be determined correctly. 

**System Environment**
This happens in any system. 

**Additional context**
N/A

uramirez8707:
**Description**
Fix issue checking if frequency is valid.
Fix reading the diurnal and pow reduction

Fixes #900 

**How Has This Been Tested?**
`make check` passes

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_yaml.F90
+++ b/diag_manager/fms_diag_yaml.F90
@@ -485,7 +485,7 @@ subroutine check_file_freq(fileobj)
   type(diagYamlFiles_type), intent(inout) :: fileobj      !< diagYamlFiles_type obj to check
 
   if (fileobj%file_freq < -1 ) &
-    call mpp_error(FATAL, "freq must be greater than -1. &
+    call mpp_error(FATAL, "freq must be greater than or equal to -1. &
       &Check you entry for"//trim(fileobj%file_fname))
   if(.not. is_valid_time_units(fileobj%file_frequnit)) &
     call mpp_error(FATAL, trim(fileobj%file_frequnit)//" is not a valid file_frequnit. &
--- a/diag_manager/fms_diag_yaml.F90
+++ b/diag_manager/fms_diag_yaml.F90
@@ -484,7 +484,7 @@ end function
 subroutine check_file_freq(fileobj)
   type(diagYamlFiles_type), intent(inout) :: fileobj      !< diagYamlFiles_type obj to check
 
-  if (fileobj%file_freq < -1 ) &
+  if (.not. (fileobj%file_freq >= -1) ) &
     call mpp_error(FATAL, "freq must be greater than or equal to -1. &
       &Check you entry for"//trim(fileobj%file_fname))
   if(.not. is_valid_time_units(fileobj%file_frequnit)) &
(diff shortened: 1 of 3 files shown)


diag_manager/fms_diag_yaml.F90
```diff
@@ -484,8 +484,8 @@ function get_total_num_vars(diag_yaml_id, diag_file_ids) &
 subroutine check_file_freq(fileobj)
   type(diagYamlFiles_type), intent(inout) :: fileobj      !< diagYamlFiles_type obj to check
 
-  if (fileobj%file_freq < 1 ) &
-    call mpp_error(FATAL, "freq must be greater than 0. &
+  if (fileobj%file_freq < -1 ) &
+    call mpp_error(FATAL, "freq must be greater than -1. &
```
thomas-robinson, line 488: This checks if `fileobj%file_freq` is less than -1.  If it IS -1, then this will be false.  Your error message says that it must be greater than -1.
Either this should be
```fortran
if (.not.(fileobj%file_freq > -1)) 
```
or your error message needs to be changed.

thomas-robinson (changes requested):

diag_manager/fms_diag_yaml.F90
uramirez8707: I changed it to: 
https://github.com/NOAA-GFDL/FMS/blob/011145d0ae844d01b7f1a474d8420ba3cd803137/diag_manager/fms_diag_yaml.F90#L487-L488

If the file_freq is not greater than on equal to -1, it should fail because freq must be greater than or equal to -1. 
It is easier to read. 

-1 is a valid freq: 
https://github.com/NOAA-GFDL/FMS/blob/2c218d0aabd5340bde8f157d15ade6efd128e163/diag_manager/diag_table.F90#L79-L86

thomas-robinson (approved):

PR #911, opened 2022-02-15
String utils module

uramirez8707:
**Description**
Adds a string_utils module that contains the following public routines and added unit tests for each:
-  fms_array_to_pointer : to convert an array of strings to c_pointers
-  fms_pointer_to_array : to convert an array of c_pointers back to an array of strings
-  fms_sort_this: sorts an array of c_pointers in alphabetical order
-  fms_find_my_string: uses a binary search to search for a string in an array of sorted c_pointers

These routines will be implemented in the diag_manager rewrite (i.e when searching the diag_yaml for a field)

To do (in a separate pr):
- Move other string-related routines to the string_utils_mods (i.e fms_c2f_string, string, str_copy, etc ...)

**How Has This Been Tested?**
`make check` passes with intel20, gnu 9.3.0, and intel18 on skylake

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/string_utils/fms_string_utils.F90
+++ b/string_utils/fms_string_utils.F90
@@ -41,8 +41,8 @@ module fms_string_utils_mod
 !> @}
 
   interface
-  !> @brief Sorts an array of pointers (my pointer) of size (p_size) in 
-  !! alphabetical order. 
+  !> @brief Sorts an array of pointers (my pointer) of size (p_size) in
+  !! alphabetical order.
   subroutine fms_sort_this(my_pointer, p_size, indices) bind(c)
     use iso_c_binding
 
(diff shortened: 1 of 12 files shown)


thomas-robinson:
```
To do (in a separate pr):

Move other string-related routines to the string_utils_mods (i.e fms_c2f_string, string, str_copy, etc ...) which are currently scattered and duplicated all over FMS
```

This should be done.  You should use a `USE` statement in fms_mod, and make these routines public through there.  Then fms_mod can depend on this instead of the other way around.

string_utils/fms_string_utils.F90
```diff
@@ -0,0 +1,148 @@
...
+!> @brief Routines to use for string manipulation
+
+!> @file
+!> @brief File for @ref fms_string_utils_mod
+
+!> @addtogroup fms_string_utils_mod
+!> @{
+module fms_string_utils_mod
+  use, intrinsic :: iso_c_binding
+  use fms_mod, only: fms_c2f_string
```
thomas-robinson, line 31: Can you move `fms_c2f_string` here, then have fms_mod `use fms_string_utils_mod, only:fms_c2f_string`.  I think you want the dependency going the other way instead...

rem1776 (approved):

PR #912, opened 2022-02-17
chore: update libFMS module with new routines

rem1776:
**Description**
Just adds a the past few updates to the libFMS.F90 file so that they're accessible through the module. If anything else added was public and not included here please let me know

**How Has This Been Tested?**
intel/impi 2020 on amd

**Checklist:**
- [x] `make distcheck` passes


bensonr (commented):
Are any or all of these intended for more than internal FMS use?

rem1776:
> Are any or all of these intended for more than internal FMS use?

I know the mpp_shift_nest_domains is for external use and some of the string routines are as well I believe. Not too sure about the parser routines, but they're public in their module so they'd be accessible either way.

bensonr:
@rem1776  - My comment was specific to the yaml and string routines/functions.  It's public in our modules because we need to use them throughout FMS itself, but that doesn't mean we necessarily want to expose them as externals from FMS by default.

rem1776:
@bensonr Yeah I get what you mean, I think we would want these included here in this case. I spoke to Uriel since he wrote most of the routines and he thinks they should be accessible externally for models.

bensonr (approved):

PR #800, opened 2021-08-13
Test script updates and input tests

rem1776:
**Description**
Was gonna make this off of the branch from #654 , but figured it might be better to make a separate PR since the comparison was huge with the updates to FMS.

- takes Seth's updates from #654 and converts each test script to use the new harness
- Adds environment variables for verbose test output when set (TEST_VERBOSE) 
- Adjusts logic so SKIP_TESTS can be used with two digit test numbers and ranges
- Adds in/fixes some tests that were in PRs or skipped previously
- Adds a configure flag, `--enable-test-input=/path`, that enables tests in FMS that require input files. The input directory and structure it needs to be in is on both dev boxes at /home/fms_test_input

Fixes #209 
Fixes #322 
Fixes #482 
Fixes #484 

**How Has This Been Tested?**
Tested on amd and skylake with intel and gnu compilers

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #933, opened 2022-03-09
Moving empty and non-existent data table handling inside data_override

### LINKED ISSUE #934: Move empty and non-existent data_table handling to data_override
The decision of whether to call data_override_init is currently handled within the various FMScoupler coupler_main.F90 routines (simple, full, SHiELD). Since some setups don't have a data_table or a grid_spec.nc, the current implementation is the only way to prevent a "grid_spec.nc not found" crash in data_override. While the current setup is adequate as a stop-gap measure, the handling of the data_table file's existence and contents should be done within data_override itself.

Remove the checks from FMSCoupler regarding data_override_init and just call the subroutines there. Within data_override, check to see if the file exists (within the read_data subroutine for the ASCII data_table, within the yaml_parser for reading the data_table.yaml). If the file does not exist, or its contents are empty, exit out of the data_override_init routine before reading in the grid_spec.nc to avoid crashes.

Keeping it as is.

**Additional context**
Since this involved parser changes for any yaml inputs, and sub-module that uses the yaml parser will likewise be affected when yaml parsing is implemeted in the future (diag_manager, field_manager, etc.).

GFDL-Eric:
**Description**
Instead of having these checks within the various couplers, we are moving the handling of empty and non-existent data_table and data_table.yaml files within data_override. Note that this involved hardcoding the file_id returned by open_and_parse_file for yaml files in order for that to be acted upon within data_override.

Fixes #934 

**How Has This Been Tested?**
Compiles on skylake. This is mostly runtime changes and they have not been tested.

**Checklist:**
- [x] `make distcheck` passes


bensonr:
I've tested the if-test for diag_table existence in two FV3 solo tests (no grid_spec and an empty data_table) and it works as expected.

GFDL-Eric:
I'm pretty sure this is already based on the 2022.01-beta2 tag, which is the latest commit? Unless another PR made a change?

GFDL-Eric:
Now I get it... I rolled back the tap-driver.sh stuff to what's currently in main.

rem1776:
@GFDL-Eric That file used to be auto-generated by autoreconf until #800 added it to the repo with some changes, so that probably messed up the merge. I see you fixed it now though

rem1776 (approved):

data_override/data_override.F90
```diff
@@ -367,7 +371,14 @@ subroutine read_table(data_table)
 
 !  Read coupler_table
     open(newunit=iunit, file='data_table', action='READ', iostat=io_status)
-    if(io_status/=0) call mpp_error(FATAL, 'data_override_mod: Error in opening file data_table')
+    if(io_status/=0) then
+      if(io_status==29) then
```
thomas-robinson, line 375: Is this compiler or system dependent?

GFDL-Eric: ugh, you're right, I thought I was being clever here and using a "standard". I guess I can just use INQUIRE instead?

GFDL-Eric: I've now pushed a change that should also obviate the need to go through the reading table portion if the file doesn't exist... let me know if anything looks like it won't work with that.

data_override/data_override.F90
```diff
@@ -365,20 +365,21 @@ subroutine read_table(data_table)
...
 !  Read coupler_table
-    open(newunit=iunit, file='data_table', action='READ', iostat=io_status)
-    if(io_status/=0) then
-      if(io_status==29) then
-        call mpp_error(NOTE, 'data_override_mod: File data_table does not exist.')
-      else
-        call mpp_error(FATAL, 'data_override_mod: Error in opening file data_table.')
-      endif
-    endif
+    inquire(file='data_table', EXIST=table_exists)
```
thomas-robinson, line 374: will this work for a yaml also?

data_override/data_override.F90
```diff
@@ -361,13 +365,21 @@ subroutine read_table(data_table)
...
+    if (.not. table_exists) then
+      call mpp_error(NOTE, 'data_override_mod: File data_table does not exist.')
+      table_size = 0
+      return
+    end if
+
     open(newunit=iunit, file='data_table', action='READ', iostat=io_status)
-    if(io_status/=0) call mpp_error(FATAL, 'data_override_mod: Error in opening file data_table')
+    if(io_status/=0) call mpp_error(FATAL, 'data_override_mod: Error in opening file data_table.')
```
thomas-robinson, line 382: Should this be in the 
```fortran 
if (.not. table_exists) then
```
block?

data_override/data_override.F90
GFDL-Eric: right now that check is in the yaml_parser, because the yaml_parser has to return an integer file_id (I've hardcoded that to 999 in the yaml_parser). Data_override then checks for a 999 file_id return from the yaml_parser. Is the suggestion here to handle the absence of the data_table.yaml directly in data_override (i.e. not even call the parser)?

data_override/data_override.F90
GFDL-Eric: this is for errors that may arise when the table doesn't exist but there's a problem reading the file. Hopefully, the return in the if block would end the subroutine if the file is missing and it wouldn't even get to this point (unless return only works for functions?)

data_override/data_override.F90
thomas-robinson: If the parser can handle it, then let the parser do it's job.

rem1776 (approved):

bensonr:
@GFDL-Eric - Does that mean subsequent calls to data_override will also return?

GFDL-Eric:
not as it's currently written... that's gonna be a problem!

GFDL-Eric:
hmm, actually, subsequent data_override_init calls _should_ return because table_size should already be an established module variable after the first data_override_init call. As for calls to data_override itself with an empty table, assuming a do i=1,0 just gets skipped, those subroutines should also return. Also, nobody should be calling data_override (2D or 3D) with an empty data table in the first place.

bensonr:
@GFDL-Eric - I'm clear on the behavior of data_override_init.  It's the behavior of **_data_override_** called directly without any conditional or trigger around it.  I've looked through an ESM codebase and only see unprotected instances in FMSCoupler and MOM6 (coupler functionality that seems to have been moved within) that could be a problem if an absent data_table isn't handled the same way as an empty data_table.

GFDL-Eric:
@bensonr For the data_override calls (both 2D and 3D), the first section in both subroutines attempts to find the index within the table. Here are the lines from 2D:

```
  if(PRESENT(override)) override = .false.
  index1 = -1
  do i = 1, table_size
     if( trim(gridname) /= trim(data_table(i)%gridname)) cycle
     if( trim(fieldname) /= trim(data_table(i)%fieldname_code)) cycle
     index1 = i                               ! field found
     exit
  enddo
  if(index1 .eq. -1) return  ! NO override was performed
```
data_table values are set to the default values on the first run of data_override_init, regardless of empty or non-existent data_table. Similarly, table_size is set to 0 on that first init. This should result in index1 remaining at -1 and the return being triggered on the last line, provided data_override_init has been called once (module_is_initialized = .true.). If it hasn't, it appears that 2D will fail, but 3D will not, because 3D has:

```
if(.not.module_is_initialized) &
       call mpp_error(FATAL,'Error: need to call data_override_init first')
```
We can add this protection to 2D to be safe, although it would only be applicable in cases where data_override_init was never called. Should we just go ahead and add it?

bensonr:
@GFDL-Eric - sounds like we are good as is.  I'll approve it here shortly.

test_fms/parser/test_yaml_parser.sh
```diff
@@ -26,7 +26,7 @@
 . ../test-lib.sh
 
 if [ ! -z $parser_skip ]; then
-  SKIP_TESTS='test_yaml_parser.[1-22]'
+  SKIP_TESTS='test_yaml_parser.[1-21]'
```
bensonr, line 29: Is the change from 22 to 21 expected?

GFDL-Eric: There was a test to check if the parser would fail with a non-existent yaml file... but with the changes it shouldn't fail. I guess we could modify the test to check for the desired behavior (file_id = 999) instead of just removing it.

bensonr: Have been assured this is good

bensonr (approved):

PR #938, opened 2022-03-16
fms_find_unique

uramirez8707:
**Description**
Adds a function to string_utils_mod to determine the number of unique strings in a sorted array and adds a test for it. This will be used in the diag_manager update to determine the number of unique diag_fields in a diag_yaml. 

**How Has This Been Tested?**
CI, including the added unit test. 

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/string_utils/fms_string_utils_binding.c
+++ b/string_utils/fms_string_utils_binding.c
@@ -162,3 +162,20 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
   strcpy(string_p, string);
   return string_p;
 }
+
+// Finds the number of unique strings in an array
+int fms_find_unique(char** arr, int *n)
+{
+  int i;
+  int nfind;
+
+  nfind=1;
+  //printf("n is %i", *n);
+  for(i=1; i<*n; i++){
+    //printf("Comparing %s and %s \n",arr[i], arr[i-1]);
+    if (strcmp(arr[i], arr[i-1]) != 0){ nfind = nfind + 1;}
+  }
+
+  //printf("nfind=%i",nfind);
+  return nfind;
+}
--- a/string_utils/fms_string_utils_binding.c
+++ b/string_utils/fms_string_utils_binding.c
@@ -163,19 +163,24 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
   return string_p;
 }
 
-// Finds the number of unique strings in an array
+/*!
+ * @brief Finds the number of unique strings in an array
+ * @param[in]  arr Array of strings
+ * @param[in]  n   Size of the array
+ * @return Number of unique strings in an array
+ */
 int fms_find_unique(char** arr, int *n)
 {
-  int i;
-  int nfind;
+  int i; // For loops
+  int nfind; // Number of unique strings in an array
+  int * ids = calloc(*n, sizeof(int)); // Array of integers initialized to 0
+
+  fms_sort_this(arr, n, ids);
 
   nfind=1;
-  //printf("n is %i", *n);
   for(i=1; i<*n; i++){
-    //printf("Comparing %s and %s \n",arr[i], arr[i-1]);
     if (strcmp(arr[i], arr[i-1]) != 0){ nfind = nfind + 1;}
   }
 
-  //printf("nfind=%i",nfind);
   return nfind;
 }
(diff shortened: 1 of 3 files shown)


string_utils/fms_string_utils_binding.c
```diff
@@ -162,3 +162,20 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
...
+  nfind=1;
+  //printf("n is %i", *n);
+  for(i=1; i<*n; i++){
+    //printf("Comparing %s and %s \n",arr[i], arr[i-1]);
+    if (strcmp(arr[i], arr[i-1]) != 0){ nfind = nfind + 1;}
+  }
+
+  //printf("nfind=%i",nfind);
+  return nfind;
+}
```
bensonr, line 181: Unless I'm misreading the logic, this only works for a sorted array.   To do this without a pre-sort, you'd need a nested loop structure.  You can test this by moving the fms_find_unique in your unit test to be line 59 prior to the sort.  If you want to stick with a single loop, you'd need add an fms_sort_this prior to the search-loop.

string_utils/fms_string_utils_binding.c
```diff
@@ -162,3 +162,20 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
...
+  int nfind;
+
+  nfind=1;
+  //printf("n is %i", *n);
+  for(i=1; i<*n; i++){
+    //printf("Comparing %s and %s \n",arr[i], arr[i-1]);
+    if (strcmp(arr[i], arr[i-1]) != 0){ nfind = nfind + 1;}
+  }
+
+  //printf("nfind=%i",nfind);
```
thomas-robinson, line 179: The unused `printf`s should be cleaned up.

thomas-robinson (changes requested):

string_utils/fms_string_utils_binding.c
```diff
@@ -162,3 +162,20 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
   strcpy(string_p, string);
   return string_p;
 }
+
+// Finds the number of unique strings in an array
+int fms_find_unique(char** arr, int *n)
```
thomas-robinson, line 167: These C function variables should be documented https://www.doxygen.nl/manual/commands.html#cmdparam

string_utils/fms_string_utils_binding.c
uramirez8707: This was only intended to work with a sorted array:
https://github.com/NOAA-GFDL/FMS/blob/3db45b1d78d4787fa47e52d8f121a4fdd4621d68/string_utils/fms_string_utils.F90#L71-L81

But the documentation should be more clear.

bensonr: You could add a call to the fms_sort_this function to ensure a sort is done prior to the comparison search.  That would remove the requirement to pass in only sorted arrays.

string_utils/fms_string_utils_binding.c
uramirez8707: Fixed

string_utils/fms_string_utils_binding.c
uramirez8707: I added the call

thomas-robinson (approved):

string_utils/fms_string_utils_binding.c
thomas-robinson: Is there a quick way to test if it's sorted?  Is the sorting of a sorted array fast enough that it doesn't matter that you sort it again?

bensonr (approved):

string_utils/fms_string_utils_binding.c
```diff
@@ -162,3 +162,25 @@ char* fms_find_my_string_binding(char** arr, int *n, char *find_me, int *np)
...
+ * @param[in]  n   Size of the array
+ * @return Number of unique strings in an array
+ */
+int fms_find_unique(char** arr, int *n)
+{
+  int i; // For loops
+  int nfind; // Number of unique strings in an array
+  int * ids = calloc(*n, sizeof(int)); // Array of integers initialized to 0
+
+  fms_sort_this(arr, n, ids);
```
ngs333, line 178: @uramirez8707 
The fms_sort_this function uses qsort function from stdlib.h. I was able to look at the code and web comments and it seems to me that it uses the known/published tricks to avoid the worst (O(N^2)) cases. But I am not 100% sure. To be closer to sure, one can check if he array is already sorted or reverse is sorted. This is fast (O(N)) in compariosn to qsort. One can also test and determine the performance as functin of size of input, with test data in sorted order, reverse sorted order, and random order.

PR #857, opened 2021-11-05
Fms mixedmode

MinsukJi-NOAA:
**Description**
This PR includes code changes to allow a single FMS library to handle running 32-bit FV3 and 64-bit MOM6 for the coupled UFS weather model. The approach taken to minimize the code changes is to assume FMS is compiled in 64 bit, and to only modify the interfaces to FV3.

**How Has This Been Tested?**
- Code changes pass the FMS CI tests (GNU) on GitHub Actions.
- UFS weather model regression tests have been run on Hera using Intel compiler with expected results
  - 64 bit tests all reproduce the baseline
  - 32 bit tests complete, but do not reproduce the baseline

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes: `make -k check` resulted in no FAIL or ERROR on Hera platform with intel and impi 2020.2

PR #962, opened 2022-04-15
Fms mixedmode update

MinsukJi-NOAA:
**Description**
This PR includes code changes to allow a single FMS library to handle running 32-bit FV3 and 64-bit MOM6 for the coupled UFS weather model. The approach taken to minimize the code changes is to assume FMS is compiled in 64 bit, and to only modify the interfaces to FV3.

**How Has This Been Tested?**
* Code changes pass the FMS CI tests (GNU) on GitHub Actions.
* UFS weather model regression tests have been successfully done using code changes in the previous PR #857
  * UFS weather model regression tests will be performed using the updated code changes in this PR

**Checklist:**
- [ ] `make distcheck` passes


thomas-robinson:
@ngs333 (and to some extent @uramirez8707 ), this is going to conflict with your development, especially in send_data.  Do you think:
1. These updates can coexist with your updates? *or*
2. These updates will be replaced by your updates?
3. @ngs333 you deleted a lot of what this code updates.  Do you foresee adding that code back in to maintain/support old behavior, or do you think what you've done will supersede these updates?

thomas-robinson:
There are two merge conflicts with the current [dmUpdate](https://github.com/NOAA-GFDL/FMS/tree/67f97ce943a6131bbf17e501974a809159a1f6ef) specifically in [diag_manager.F90](https://github.com/NOAA-GFDL/FMS/blob/67f97ce943a6131bbf17e501974a809159a1f6ef/diag_manager/diag_manager.F90)
```
git remote add gfdl https://github.com/NOAA-GFDL/FMS.git
git remote add emc https://github.com/NOAA-EMC/FMS.git
git fetch gfdl
git fetch emc
git checkout gfdl/dmUpdate
git merge emc/fms_mixedmode
```
There will be more conflicts when the send_data work is merged in
https://github.com/NOAA-GFDL/FMS/pull/979

thomas-robinson:
@MinsukJi-NOAA we should be able to resolve these conflicts because they seem relatively straight forward.  The "old" and "modern" routines are going to co-exist for some time.  We will move forward with our development based off of your updates.

thomas-robinson:
@rem1776 We will merge in main and then merge this PR.  We will tag this `2022.03-alpha1` and run testing.

thomas-robinson:
@MinsukJi-NOAA We had a few more updates, but they should be minor.  The check for `range` being size 2 was moved and there were some typos.  Otherwise I think we are good

thomas-robinson (approved):
The changes to the diag manager are now in line with our diag manager development, and I think we can merge this.

bensonr (changes requested):
@binli2337 @MinsukJi-NOAA @junwang-noaa - as part of the 2022.02 FMS release, constants handling was rewritten to isolate the individual constants groups by laboratory/use.  Please look at the [new structure](https://github.com/NOAA-GFDL/FMS/tree/main/constants) and propose a way to introduce the desired constant behavior/handling.

MinsukJi-NOAA:
> @binli2337 @MinsukJi-NOAA @junwang-noaa - as part of the 2022.02 FMS release, constants handling was rewritten to isolate the individual constants groups by laboratory/use. Please look at the [new structure](https://github.com/NOAA-GFDL/FMS/tree/main/constants) and propose a way to introduce the desired constant behavior/handling.

@bensonr Could you point out the hash where these changes to constants were made?

bensonr:
@MinsukJi-NOAA - [16003bb](https://github.com/NOAA-GFDL/FMS/commit/16003bb31ec930bb0e468eabb8dcc0333f149ff8)

junwang-noaa:
@bensonr I want to double check with you, my understanding is that in GFDL applications FMS will be compiled along with model code with corresponding compile options. So the mixed mode changes for constantsR4 have no impact on GFDL/GOES applications. 

The constantR4 will only be added for the prebuilt FMS library in UFS. When running in mixed mode, dycore compiled with 32BIT=Y can still use the r4 constants. Will GFDL applications run in mixed mode coupled applications? What are the compile options for FMS you are going to use in this mixed mode runs?

bensonr:
@junwang-noaa - the bottom line is it doesn't conform to the current programming and macros for constants within FMS.  Being that FMS is a community library, it is not in our best interest to have conflicting approaches which could cause confusion and future issues.  In my opinion, leaving all the constants in full precision and casting them to the appropriate precision in the various components would be the correct solution. 

GEOS does have the desire (and I believe the capability) of running mixed-mode.  GFDL's SHiELD model has always run with mixed-mode and the climate researchers have expressed interest in exploring the capability, but have concerns about conservation that need to be satisfied.

junwang-noaa:
@bensonr Thanks for the explanation. Bin will update the constantR4 to follow the same structure. But in general when mixed model FMS is used in the mixed model coupled runs (fms is compiled in R8, dycore runs in 32BIT/mom6 runs on 64BIT), we will need the hard coded R4 constantsR4 module used in dycore. Constants module can be used in all other configurations (fms compiled in r4 (r8) and dycore runs in r4 (r8), or fms compiled in r8 and mom6/dycore run in r8).

PR #966, opened 2022-04-25
Diag axis obj

uramirez8707:
**Description**
This add a diag_axis object to be used in the modern diag_manager. 
There are still a lot of todos. 

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_axis_object.F90
+++ b/diag_manager/fms_diag_axis_object.F90
@@ -1,0 +1,288 @@
+!***********************************************************************
+!*                   GNU Lesser General Public License
+!*
+!* This file is part of the GFDL Flexible Modeling System (FMS).
+!*
+!* FMS is free software: you can redistribute it and/or modify it under
+!* the terms of the GNU Lesser General Public License as published by
+!* the Free Software Foundation, either version 3 of the License, or (at
+!* your option) any later version.
+!*
+!* FMS is distributed in the hope that it will be useful, but WITHOUT
+!* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+!* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+!* for more details.
+!*
+!* You should have received a copy of the GNU Lesser General Public
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
+!***********************************************************************
+
+!> @defgroup fms_diag_axis_object_mod fms_diag_axis_object_mod
+!> @ingroup diag_manager
+!! @brief fms_diag_axis_object_mod stores the diag_axis_object
+
+!> @file
+!> @brief File for @ref diag_axis_object_mod
+
+!> @addtogroup fms_diag_axis_object_mod
+!> @{
+module fms_diag_axis_object_mod
+  use mpp_domains_mod, only:  domain1d, domain2d, domainUG, mpp_get_compute_domain, CENTER, &
+                            & mpp_get_compute_domain
+  use platform_mod,    only:  r8_kind, r4_kind
+  use diag_data_mod,   only:  diag_atttype
+  use mpp_mod,         only:  FATAL, mpp_error
+  implicit none
+
+  PRIVATE
+
+  public :: diag_axis_t, diag_axis_init, set_subaxis
+  !> @}
+
+  !> @brief Type to hold the domain info for an axis
+  !! This type was created to avoid having to send in "Domain", "Domain2", "DomainUG" as arguments into subroutines
+  !! and instead only 1 class(fms_domain_t) argument can be send
+  !> @ingroup diag_axis_object_mod
+  type fms_domain_t
+    contains
+      procedure :: set => set_axis_domain
+      procedure :: length => get_length
+  end type fms_domain_t
+
+  !> @brief Type to hold the 1d domain
+
+  type, extends(fms_domain_t) :: fms_domain1d_t
+     type(domain1d) :: Domain
+  end type
+
+  !> @brief Type to hold the 2d domain
+  type, extends(fms_domain_t) :: fms_domain2d_t
+    type(domain2d) :: Domain2
+  end type
+
+  !> @brief Type to hold the unstructured domain
+  type, extends(fms_domain_t) :: fms_domainUG_t
+    type(domainUG) :: DomainUG
+  end type
+
+  !> @brief Type to hold the subaxis
+  !> @ingroup diag_axis_object_mod
+  TYPE subaxis_t
+    CHARACTER(len=:), ALLOCATABLE :: subaxis_name   !< Name of the subaxis
+    INTEGER                       :: starting_index !< Starting index of the subaxis relative to the parent axis
+    INTEGER                       :: ending_index   !< Ending index of the subaxis relative to the parent axis
+    class(*)        , ALLOCATABLE :: bounds         !< Bounds of the subaxis (lat/lon or indices)
+    contains
+      procedure :: exists => check_if_subaxis_exists
+  END TYPE subaxis_t
+
+  !> @brief Type to hold the diagnostic axis description.
+  !> @ingroup diag_axis_object_mod
+  TYPE diag_axis_t
+     CHARACTER(len=:),   ALLOCATABLE, private :: axis_name       !< Name of the axis
+     CHARACTER(len=:),   ALLOCATABLE, private :: units           !< Units of the axis
+     CHARACTER(len=:),   ALLOCATABLE, private :: long_name       !< Long_name attribute of the axis
+     CHARACTER(len=1)               , private :: cart_name       !< Cartesian name "X", "Y", "Z", "T", "U", "N"
+     CLASS(*),           ALLOCATABLE, private :: axis_data(:)    !< Data of the axis
+     !< TO DO this can be a dlinked to avoid having limits
+     type(subaxis_t)                , private :: subaxis(3)      !< Array of subaxis
+     integer                        , private :: nsubaxis        !< Number of subaxis
+     class(fms_domain_t),ALLOCATABLE, private :: axis_domain     !< Domain
+     INTEGER                        , private :: length          !< Global axis length
+     INTEGER                        , private :: direction       !< Direction of the axis 0, 1, -1
+     INTEGER                        , private :: edges           !< Axis ID for the previously defined "edges axis"
+     CHARACTER(len=128)             , private :: aux             !< Auxiliary name, can only be <TT>geolon_t</TT>
+                                                                 !! or <TT>geolat_t</TT>
+     CHARACTER(len=128)             , private :: req             !< Required field names.
+     INTEGER                        , private :: tile_count      !< The number of tiles
+     TYPE(diag_atttype),allocatable , private :: attributes(:)   !< Array to hold user definable attributes
+     INTEGER                        , private :: num_attributes  !< Number of defined attibutes
+     INTEGER                        , private :: domain_position !< The position in the doman (NORTH or EAST or CENTER)
+
+     contains
+
+     PROCEDURE :: register => diag_axis_init
+     PROCEDURE :: axis_length => get_axis_length
+     PROCEDURE :: set_subaxis
+
+     ! TO DO:
+     ! PROCEDURE :: write_axis_metadata
+     ! PROCEDURE :: write_axis_data
+     ! PROCEDURE :: get_fileobj_type_needed (use the domain to figure out what fms2 fileobj to use)
+     ! Get/has/is subroutines as needed
+  END TYPE diag_axis_t
+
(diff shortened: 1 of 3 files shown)


diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+
+  !!!!!!!!!!!!!!!!! DIAG AXIS PROCEDURES !!!!!!!!!!!!!!!!!
+  !> @brief Initialize the axis
+  subroutine diag_axis_init(obj, name, axis_data, units, cart_name, long_name, direction,&
+  & set_name, edges, Domain, Domain2, DomainU, aux, req, tile_count, domain_position )
+    class(diag_axis_t)                       :: obj             !< Diag_axis obj
+    CHARACTER(len=*),   INTENT(in)           :: name            !< Name of the axis
+    class(*),           INTENT(in)           :: axis_data(:)    !< Array of coordinate values
+    CHARACTER(len=*),   INTENT(in)           :: units           !< Units for the axis
+    CHARACTER(len=*),   INTENT(in)           :: cart_name       !< Cartesian axis ("X", "Y", "Z", "T", "U", "N")
```
thomas-robinson, line 126: The `len` for `cart_name` should always be 1.  Is there any reason to not have it be 1?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+
+  !> @addtogroup fms_diag_yaml_mod
+  !> @{
+  contains
+
+  !!!!!!!!!!!!!!!!! DIAG AXIS PROCEDURES !!!!!!!!!!!!!!!!!
+  !> @brief Initialize the axis
+  subroutine diag_axis_init(obj, name, axis_data, units, cart_name, long_name, direction,&
+  & set_name, edges, Domain, Domain2, DomainU, aux, req, tile_count, domain_position )
+    class(diag_axis_t)                       :: obj             !< Diag_axis obj
```
thomas-robinson, line 122: Should this be `intent(out)`?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+    if (present(edges)) obj%edges = edges
+
+    if (present(aux)) obj%aux = diag_copy_string(aux)
+    if (present(req)) obj%req = diag_copy_string(req)
+
+    obj%nsubaxis = 0
+  end subroutine diag_axis_init
+
+  !> @brief Get the length of the axis
+  !> @return axis length
```
thomas-robinson, line 193: Imagine it like this:
```fortran
integer :: axis_length !< axis length
```
Would this be acceptable?  I mean... I guess it would be correct, but it seems a little too short.  Change it or don't, I don't care either way.  This is just nitpicky

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+  function get_axis_length(obj) &
+  result (axis_length)
+    class(diag_axis_t), intent(inout) :: obj !< diag_axis obj
+    integer                           :: axis_length
+
+    axis_length = obj%length
+
+    !< If the axis is domain decomposed axis_length will be set to the length for the current PE:
+    if (allocated(obj%axis_domain)) axis_length = obj%axis_domain%length(obj%cart_name, obj%domain_position)
```
thomas-robinson, line 202: Should this be an if/else statement?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+
+  end function
+
+  !> @brief Set the subaxis of the axis obj
+  subroutine set_subaxis(obj, bounds)
```
thomas-robinson, line 207: A routine like this already exist.  Are you planning on copying it over/modifying it for the purpose of filling in this object?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+  !!!!!!!!!!!!!!!!!! SUB AXIS PROCEDURES !!!!!!!!!!!!!!!!!
+  !> @brief Check if a subaxis was already defined
+  !> @return Flag indicating if a subaxis is already defined
+  function check_if_subaxis_exists(obj,bounds) &
+  result(exists)
+    class(subaxis_t), INTENT(INOUT) :: obj       !< diag_axis obj
+    class(*),         INTENT(IN)    :: bounds(:) !< bounds of the subaxis
+    logical                         :: exists
+
+    !< TO DO: compare bounds
```
thomas-robinson, line 231: These are some big TODOs.  I think that maybe a list of the TODOs should go into an issue for this module.  There seems to be a lot of work that you have planned.

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+!> @ingroup diag_manager
+!! @brief fms_diag_axis_object_mod stores the diag_axis_object
+
+!> @file
+!> @brief File for @ref diag_axis_object_mod
+
+!> @addtogroup fms_diag_axis_object_mod
+!> @{
+module fms_diag_axis_object_mod
+  use mpp_domains_mod, only:  domain1d, domain2d, domainUG, mpp_get_compute_domain, CENTER, &
```
thomas-robinson, line 30: What about `NORTH` and `EAST`?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+  PRIVATE
+
+  public :: diag_axis_t, diag_axis_init, set_subaxis
+  !> @}
+
+  !> @brief Type to hold the domain info for an axis
+  !! This type was created to avoid having to send in "Domain", "Domain2", "DomainUG" as arguments into subroutines
+  !! and instead only 1 class(fms_domain_t) argument can be send
+  !> @ingroup diag_axis_object_mod
+  type fms_domain_t
```
thomas-robinson, line 46: this is a diag manager type, so it should be named `fmsDiagDomain_t` or `diagDomain_t`.  Using this name makes it seem like it comes from `fms_mod`.    The same is true for the types that extend this one.

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+    exists = .false.
+  end function
+
+  !> @brief Get the length of a 2D domain
+  !> @return Length of the 2D domain
+  function get_length(obj, cart_axis, domain_position) &
+  result (length)
+    class(fms_domain_t), INTENT(INOUT) :: obj       !< diag_axis obj
+    character(len=*),    INTENT(IN)    :: cart_axis !< cart_axis of the axis
+    integer,             INTENT(IN)    :: domain_position !< Domain position (CENTER, NORTH, EAST)
```
thomas-robinson, line 241: If you are inquiring about the domain, shouldn't you already know the `cart_axis` in object you are calling it for?  I guess what I'm saying is shouldn't this function be called using the axis object instead of the domain object?  Then you wouldn't need any input arguments other than the object.  Is there a use case for this function where an axis object wouldn't/couldn't be used?

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+      obj%Domain = Domain
+    type is (fms_domain2d_t)
+      obj%Domain2 = Domain2
+    type is (fms_domainUG_t)
+      obj%DomainUG = DomainU
+    end select
+  end subroutine set_axis_domain
+
+  !!!!!!!!!!!!!!!!! OTHER FUNCTIONS/SUBROUTINES !!!!!!!!!!!!!!!!!
+  !> @brief Copy a allocatable string
```
thomas-robinson, line 272: Copy *an* allocatable string

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+  !> TO DO Move this somewhere else?
+  function diag_copy_string(string_in) &
+  result(string_out)
+    character(len=*),             INTENT(IN) :: string_in !< String to copy
+    character(len=:), ALLOCATABLE            :: string_out
+
+    allocate(character(len=len_trim(string_in)) :: string_out)
+    string_out = trim(string_in)
+
+  end function
```
thomas-robinson, line 283: 1. Why do you need this at all?  Why not just allocate and copy the string in line?
2. There is a whole module of string routines external to diag manager where this would fit better.

thomas-robinson (changes requested):
In addition to the comments made in the review:
1. Your type names don't follow the [capitalWordFormat](https://github.com/NOAA-GFDL/FMS/blob/main/CODE_STYLE.md#derived-types) and need to be fixed.
2. You have a lot of big TODOs in this one.  I think that you should make an issue that has the TODOs in it for easier tracking.

diag_manager/fms_diag_axis_object.F90
```diff
@@ -0,0 +1,287 @@
...
+  use mpp_mod,         only:  FATAL, mpp_error
+  implicit none
+
+  PRIVATE
+
+  public :: diag_axis_t, diag_axis_init, set_subaxis
+  !> @}
+
+  !> @brief Type to hold the domain info for an axis
+  !! This type was created to avoid having to send in "Domain", "Domain2", "DomainUG" as arguments into subroutines
```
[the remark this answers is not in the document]
uramirez8707: The diag_obj is going to use the diag_axis to figure out what domain the field is in.

diag_manager/fms_diag_axis_object.F90
uramirez8707: `axis_length` is already defined in L199 if it is not.

diag_manager/fms_diag_axis_object.F90
uramirez8707: They are not being used... It might be used later, but not right now. 
Center is just used to initialize the %domain_position variable:
https://github.com/NOAA-GFDL/FMS/blob/f8610606f52991c97010d5bb79395af6e3dbad25/diag_manager/fms_diag_axis_object.F90#L173

diag_manager/fms_diag_axis_object.F90
uramirez8707: less is more

diag_manager/fms_diag_axis_object.F90
uramirez8707: yes

diag_manager/fms_diag_axis_object.F90
uramirez8707: it is cleaner and less error prone to do:
```F90
string_out = diag_copy_string(string_in)
```

than

```F90
    allocate(character(len=len_trim(string_in)) :: string_out)
    string_out = trim(string_in)
```

especially when it is done multiple times. 

Moving this function is already in the TO DO, it has to be done through main first ...

uramirez8707:
@thomas-robinson 
I think i addressed all your comments in https://github.com/NOAA-GFDL/FMS/pull/966/commits/eb29e721c954b6ec694c714804cea79660eb14cc or by replying to them. 

I also opened issue #976

diag_manager/fms_diag_axis_object.F90
thomas-robinson: Yeah, so if `(allocated(obj%axis_domain))` you are setting `axis_length` twice instead of once.
```fortran
if (.not.allocated(obj%axis_domain)) then
    axis_length = obj%length
else
    axis_length = obj%axis_domain%length(obj%cart_name, obj%domain_position)
endif
```
Where or not it evaluates to true, the code will still check the `if` statement, so you are just adding an extra instruction.

diag_manager/fms_diag_axis_object.F90
thomas-robinson: They will definitely be used, but you can add them when needed.

diag_manager/fms_diag_axis_object.F90
thomas-robinson: smh

diag_manager/fms_diag_axis_object.F90
thomas-robinson:
```fortran
string_out = trim(string_in)
```
You shouldn't need to allocate the string length, the assignment should handle that for you.  Then you don't need a function (which should be named `fms_string_copy`) or an allocate statement.  
https://fortranwiki.org/fortran/files/character_handling_in_Fortran.html

thomas-robinson (commented):
I resolved most of the conversations (some begrudgingly).  There are still two outstanding though that should be addressed.

thomas-robinson (approved):

PR #982, opened 2022-06-01
diag_axis io

uramirez8707:
**Description**
Adds procedure to the axis_object that write the axis metadata and the axis data. 
Adds a subroutine to determine the fms2_io fileobj to use

Fixes # (issue)

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_axis_object.F90
+++ b/diag_manager/fms_diag_axis_object.F90
@@ -31,7 +31,8 @@ module fms_diag_axis_object_mod
   use mpp_domains_mod, only:  domain1d, domain2d, domainUG, mpp_get_compute_domain, CENTER, &
                             & mpp_get_compute_domain, NORTH, EAST
   use platform_mod,    only:  r8_kind, r4_kind
-  use diag_data_mod,   only:  diag_atttype, max_axes, NO_DOMAIN, TWO_D_DOMAIN, UG_DOMAIN
+  use diag_data_mod,   only:  diag_atttype, max_axes, NO_DOMAIN, TWO_D_DOMAIN, UG_DOMAIN, &
+                              direction_down, direction_up
   use mpp_mod,         only:  FATAL, mpp_error, uppercase
   use fms2_io_mod,     only:  FmsNetcdfFile_t, FmsNetcdfDomainFile_t, FmsNetcdfUnstructuredDomainFile_t, &
                             & register_axis, register_field, register_variable_attribute, write_data
@@ -270,7 +271,7 @@ module fms_diag_axis_object_mod
     select case (obj%direction)
     case (direction_up)
       call register_variable_attribute(fileobj, axis_name, "positive", "up", str_len=2)
-    case (direction_down))
+    case (direction_down)
       call register_variable_attribute(fileobj, axis_name, "positive", "down", str_len=4)
     end select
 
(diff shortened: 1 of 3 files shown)


diag_manager/fms_diag_axis_object.F90
```diff
@@ -384,6 +477,39 @@ subroutine check_if_valid_edges(edges)
        call mpp_error(FATAL, "diag_axit_init: The edge axis has not been defined. "&
                              "Call diag_axis_init for the edge axis first")
   end subroutine check_if_valid_edges
+
+  !> @brief Loop through a variable's axis_id to determine and set the fms2_io filobj type to use
+  !!        and return the domain
+  subroutine determine_the_fileobj_type(axis_id, fileobj_type, domain, var_name)
```
thomas-robinson, line 483: Where is this being called?  

Why is `domain` a pointer?  I'm not 100% sure it's thread safe because of the pointer (I think it is thread safe, but I am just not totally positive).  

The description says "to determine and **set** the fms2_io filobj type to use" but it uses `axis_obj(j)%fileobj_type` which is already set.  So nothing is being set here.

This doesn't return a fileobj type; it returns an integer.  If it didn't have the domain pointer as an `intent (out)` argument, this could be a pure integer function instead of a subroutine.

thomas-robinson (changes requested):

diag_manager/fms_diag_axis_object.F90
uramirez8707: > Where is this being called?
This is going to be called in `register_diag_field`, the field has an array of axis ids, this function is going to return the domain and the "fileobj_type" (the type of fms2_io fileobj the variable needs). This then has to be passed on to the diag_files. so we can know what fms2_io obj to open and we have the domain to open the file with. 
I think I should rename "fileobj_type" to "type_of_domain"

> Why is domain a pointer? I'm not 100% sure it's thread safe because of the pointer (I think it is thread safe, but I am just not totally positive).
This is going to be called the init, so there is no openmp(?). If it is not a pointer, then it has to be a copy ...

> The description says "to determine and set the fms2_io filobj type to use" but it uses axis_obj(j)%fileobj_type which is already set. So nothing is being set here.
what if it says "to determine and return the domain and domain_type to use"

thomas-robinson: OK, so what happens to this domain pointer?  Is the domain pointer you pass in part of the diag_object?  I think if you had the actual call to this routine somewhere (even commented out for now), it would be more helpful for me to understand it.

thomas-robinson (approved):

diag_manager/fms_diag_axis_object.F90
```diff
@@ -384,6 +484,38 @@ subroutine check_if_valid_edges(edges)
        call mpp_error(FATAL, "diag_axit_init: The edge axis has not been defined. "&
                              "Call diag_axis_init for the edge axis first")
   end subroutine check_if_valid_edges
+
+  !> @brief Loop through a variable's axis_id to determine and return the domain type and domain to use
```
thomas-robinson, line 488: I still think this description is lacking.  The \brief describes what this subroutine does, but there is no explanation for it's use.  I think with the next update, you should have a \description instead of \brief that explains why you use this.  In the larger plan, this routine is necessary.  If you're just looking at this module, you will scratch your head and wonder why this exists.

thomas-robinson (approved):

PR #988, opened 2022-06-27
find_diag_field,  get_num_unique_fields fix

uramirez8707:
**Description**
`find_diag_field` only searches for the variable_name, it needs to search for the module name too. It is possible to have duplicate variable names that come from different modules:

For example, in AM4 the diag_field `area` is present twice with two different modules and two different grids!
https://github.com/NOAA-GFDL/AM4/blob/7fe212ab2f942f808c51f701c809c336cf5176f6/run/diag_table#L20
https://github.com/NOAA-GFDL/AM4/blob/7fe212ab2f942f808c51f701c809c336cf5176f6/run/diag_table#L402

`get_num_unique_fields` also needs to count these as two unique diag_fields.

This PR:
- Instead of saving just the variable name, it saves `varname:module_name` to the `variable_list`
- Adds module_name as an argument to `find_diag_field` and it now searches for `varname:module_name` to be consistent with the `variable_list`
- Because the `variable_list` has both the variable name and the module name it will be able to get the correct number of unique fields 

Fixes #
**How Has This Been Tested?**
CI including updated tests

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_manager.F90
+++ b/diag_manager/diag_manager.F90
@@ -479,7 +479,7 @@ end function register_diag_field_array
 #ifdef use_yaml
     integer, allocatable :: diag_field_indices(:) !< indices where the field was found
 
-    diag_field_indices = find_diag_field(field_name)
+    diag_field_indices = find_diag_field(field_name, module_name)
     if (diag_field_indices(1) .eq. diag_null) then
       !< The field was not found in the table, so return diag_null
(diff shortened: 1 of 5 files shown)


thomas-robinson (approved):
It would be cleaner to use a derived type that just has module and file name, but this *works*. If you used a derived type, you wouldn't have a **:** to worry about.

PR #998, opened 2022-07-07
Modern_diag remaining init work

uramirez8707:
**Description**
- Implements the modern version of `diag_field_add_attribute` and adds `attributes` and `num_attributes` to the diag_obj
- Implements the modern version of `get_diag_field_id`, `get_axis_length` and `get_domain2d`
- Fixes bugs that were found when running AM4.0

Fixes # (issue)

**How Has This Been Tested?**
CI
AM4.0 runs for 0 days with use_modern_diag=.true. and use_modern_diag=.false. (-Duse_yaml)

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_axis_object.F90
```diff
@@ -556,16 +557,28 @@ subroutine get_domain_and_domain_type(axis_id, domain_type, domain, var_name)
...
+               domain => axis_obj(j)%axis_domain
           endif
         else
           call mpp_error(FATAL, "The variable:"//trim(var_name)//" has axis that are not in the same domain")
         endif
       endif
     enddo
   end subroutine get_domain_and_domain_type
+
+  function fms_get_axis_length(axis_id)&
```
thomas-robinson, line 572: This function needs documentation

diag_manager/fms_diag_object.F90
```diff
@@ -116,7 +119,7 @@ module fms_diag_object_mod
      procedure :: has_data_RANGE
 ! Get functions
      procedure :: get_diag_id => fms_diag_get_id
-     procedure :: get_metadata
+     procedure :: get_attributes
```
thomas-robinson, line 122: why rename this?  seems like extra work.

diag_manager/fms_diag_object.F90
```diff
@@ -1131,4 +1131,43 @@ FUNCTION get_diag_obj_from_id ( id ) result (obj_ptr)
...
+!> @brief Determines the diag_obj id corresponding to a module name and field_name
+!> @return diag_obj id
+INTEGER FUNCTION fms_get_diag_field_id(module_name, field_name)
+  CHARACTER(len=*), INTENT(in) :: module_name !< Module name that registered the variable
+  CHARACTER(len=*), INTENT(in) :: field_name !< Variable name
+
+  integer :: i !< For do loops
+
+  fms_get_diag_field_id = DIAG_FIELD_NOT_FOUND
+  do i = 1, registered_variables
```
thomas-robinson, line 1164: Would a `where` be better here?  I feel like this is very inefficient but I can't think of anything else that might be better.

diag_manager/fms_diag_object.F90
```diff
@@ -1131,4 +1131,43 @@ FUNCTION get_diag_obj_from_id ( id ) result (obj_ptr)
...
+    call mpp_error(FATAL, "diag_field_add_attribute: Number of attributes exceeds max_field_attributes for field:"&
+                           //trim(obj%varname)//".  Increase diag_manager_nml:max_field_attributes.")
+
+  call obj%attributes(obj%num_attributes)%add(att_name, att_value)
+  nullify(obj)
+end subroutine fms_diag_field_add_attribute
+
+!> @brief Determines the diag_obj id corresponding to a module name and field_name
+!> @return diag_obj id
+INTEGER FUNCTION fms_get_diag_field_id(module_name, field_name)
```
thomas-robinson, line 1157: Is this a pure function?

thomas-robinson (changes requested):

diag_manager/fms_diag_object.F90
uramirez8707: I think so

diag_manager/fms_diag_object.F90
uramirez8707: just to be consistent with the axis_obj and with the `diag_field_add_attribute` function

diag_manager/fms_diag_object.F90
uramirez8707: I don't think there is any way around it. 

Maybe like this. There is no reason to compare the modname if the varname is not the same? 
```F90
if ((diag_objs(i)%get_varname() .eq. trim(field_name)) then
  if ( diag_objs(i)%get_modname() .eq. trim(module_name)) then
    fms_get_diag_field_id = i
    return
  endif
endif
```

diag_manager/diag_manager.F90
```diff
@@ -4220,7 +4232,11 @@ SUBROUTINE diag_field_add_attribute_scalar_i(diag_field_id, att_name, att_value)
     CHARACTER(len=*), INTENT(in) :: att_name !< new attribute name
     INTEGER, INTENT(in) :: att_value !< new attribute value
 
-    CALL diag_field_add_attribute_i1d(diag_field_id, att_name, (/ att_value /))
+    if (use_modern_diag) then
+      call fms_diag_field_add_attribute(diag_field_id, att_name, (/att_value /))
```
thomas-robinson, line 4236: Is this a wrapper on an object procedure?  Can you just call the object procedure directly?  there are a few of these...

diag_manager/fms_diag_axis_object.F90
```diff
@@ -556,16 +557,30 @@ subroutine get_domain_and_domain_type(axis_id, domain_type, domain, var_name)
...
         else
           call mpp_error(FATAL, "The variable:"//trim(var_name)//" has axis that are not in the same domain")
         endif
       endif
     enddo
   end subroutine get_domain_and_domain_type
+
+  !> @brief Gets the length of the axis based on the axis_id
+  !> @return Axis_length
+  function fms_get_axis_length(axis_id)&
```
thomas-robinson, line 574: Why not make this an object function?  
axis_length = axis_obj(axis_id)%get_axis_length

diag_manager/fms_diag_file_object.F90
```diff
@@ -462,7 +462,7 @@ subroutine add_axes(obj, axis_ids)
   do i = 1, size(axis_ids)
     do j = 1, obj%number_of_axis
       !> Check if the axis already exists, if it does leave this do loop
```
thomas-robinson, line 464: This comment needs to be updated to the change

thomas-robinson (changes requested):
There's a comment that needs to be updated.  

There are a few functions that could be object procedures, but they are not.

diag_manager/diag_manager.F90
uramirez8707: fms_diag_field_add_attribute is a wrapper basically to avoid having to duplicate this code for each case:

https://github.com/NOAA-GFDL/FMS/blob/09a396e4cc62f19a7db9a9d1521513f13eb5302e/diag_manager/fms_diag_object.F90#L1143-L1149

diag_manager/fms_diag_axis_object.F90
uramirez8707: The model is going to call `get_axis_length`. 

`get_axis_length` currently lives in diag_axis_mod, so I added `fms_get_axis_length` as a wrapper to the object function to avoid having axis_obj code in the old diag_axis_mod.

thomas-robinson (approved):

PR #965, opened 2022-04-21
update CI images and testing options

rem1776:
**Description**
Updates the CI to use images built off the e4s spack images. This will allow us more flexibility with dependencies than the older image since it used apt package manager installs. The new image uses gcc 9.3.0, netcdf 4.8.0, libyaml 0.2.5, and mpich 3.4.2

Also changes the CI workflow to run the older input file tests, test the serial compilation and increase the amount of output for debugging.

This will need someone with admin privileges to remove the gnu CI requirement, and then re-add the new jobs.

**How Has This Been Tested?**
CI

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #1001, opened 2022-07-11
Field manager updates

### LINKED ISSUE #1000: Field Manager yaml support
There is a need to have the field manager be able to parse a yaml file as part of the transition to yaml.

An update of field_manager.F90, and the possible addition of an fm_yaml.F90, to tie it all together.

Sticking with our rollyourown format.

**Additional context**

GFDL-Eric:
**Description**
Updates the field manager to use yaml. Includes a new file (fm_yaml.F90) that creates a nested object list, which is then utilized by field_manager to populate the current objects. Should not cause any backwards compatibility issues. Also involves substantial code cleanup within field_manager, including the removal of unused functions and reducing the verbosity within the code.

Fixes #1000 

**How Has This Been Tested?**
Tested on Skylake with Intel19.

**Checklist:**
- [x] `make distcheck` passes


rem1776:
@GFDL-Eric This just needs a small makefile change and then it'll pass the gnu CI once #965 is merged. In `test_fms/field_manager/Makefile.am` it'll need the `*.yaml` extension added to the CLEANFILES list.

field_manager/field_manager.F90
```diff
@@ -5058,47 +3617,30 @@ function fm_copy_list(list_name, suffix, create ) &
...
-endif  !}
-
-!
+  endif
+endif
 !        Find the list
-!
-if (success) then  !{
-  method(:) = ' '
-  control(:) = ' '
```
rem1776, line 5099: should these two strings be initialized in the if below?

rem1776 (changes requested):

field_manager/field_manager.F90
GFDL-Eric: I'm not sure what to do here. In the part you referenced, method and control have a fixed dimension of MAX_FIELD_METHODS. However, in theory, we'd want these to be allocatable according to whatever the actual length of control and method are from the result of fm_find_methods. I think, although I'm not sure, that the output of the fm_find_methods does the allocation, so we don't have to worry about it going unallocated. it would then just be a question of if we think not having a "default" value for them would be dangerous. I can't add those lines in as is because they aren't allocated until the fm_find_methods routine is called. Any alternatives you can think of or should we just leave it?

bensonr (commented):
I don't know what has happened with baselining of your field_try_again branch, but these constants4/ files that are being added as new should not be there.

field_manager/field_manager.F90
rem1776: Yeah i see what you mean. I don't see an allocation anywhere but i think since it's intent(out) and written to in `find_methods` it works out. Maybe it should be allocated and initialized in find_methods before it's written to, but that function is recursive so that might be a bad idea and it seems to be working fine. I guess we should leave it as is.

thomas-robinson (changes requested):
Please add some more documentation

thomas-robinson (approved):

PR #1019, opened 2022-08-15
Dm buffer object

rem1776:
**Description**
adds the file for the buffers with a few of the initial changes to the other files.

Fixes # (issue)

**How Has This Been Tested?**

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
+!> @author Ryan Mulhall
```
thomas-robinson, line 1: You need a license

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    ! TODO deferred routines, will require some interfaces 
+    !procedure(allocate_buffer), deferred :: allocate_buffer
+    !procedure, deferred :: get_buffer 
+    !procedure, deferred :: initialize_buffer
+
+end type fmsDiagBuffer_class
+
+!> holds pointer to allocated buffer0-5d objects
+type :: fmsDiagBufferContainer_type
+    class(fmsDiagBuffer_class), pointer :: buffer_obj
```
thomas-robinson, line 54: I think this should be allocatable not a pointer.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+!
+contains
+
+
+!!--------module routines
+
+logical function fms_diag_buffer_init(buffobjs)
+    class(fmsDiagBufferContainer_type), allocatable, intent(out) :: buffobjs(:)
+    allocate(buffobjs(DEFAULT_BUFFER_LIST_SIZE))
+    fms_diag_buffer_init = allocated(buffobjs)
```
thomas-robinson, line 162: You should set the IDs to diag_not_registered or something like that

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    rslt%buffer_obj => buffobj
+end function 
+
+!!--------generic routines for any fmsDiagBuffer_class objects 
+
+!> Remaps 0-5d data buffer from the given object onto a 5d array and sets it in the type
+subroutine remap_buffer(buffobj)
+    class(fmsDiagBuffer_class), intent(inout) :: buffobj
+    integer                                   :: buff_bounds(5)
+
```
thomas-robinson, line 180: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+logical function fms_diag_buffer_init(buffobjs)
+    class(fmsDiagBufferContainer_type), allocatable, intent(out) :: buffobjs(:)
+    allocate(buffobjs(DEFAULT_BUFFER_LIST_SIZE))
+    fms_diag_buffer_init = allocated(buffobjs)
+end function
+
+!> creates a container type with the given buffer object
+function fms_diag_buffer_create_container(buffobj) &
+result(rslt)
+    class(fmsDiagBuffer_class), allocatable, target :: buffobj
```
thomas-robinson, line 168: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+
+!!--------module routines
+
+logical function fms_diag_buffer_init(buffobjs)
+    class(fmsDiagBufferContainer_type), allocatable, intent(out) :: buffobjs(:)
+    allocate(buffobjs(DEFAULT_BUFFER_LIST_SIZE))
+    fms_diag_buffer_init = allocated(buffobjs)
+end function
+
+!> creates a container type with the given buffer object
```
thomas-robinson, line 165: result needs to be included in the description

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+                    buffobj%remap_buffer = buff
+            end select
+        class default
+            call mpp_error( FATAL, 'remap_buffer_pointer: invalid buffer type for remapping')
+    end select
+   
+end subroutine 
+
+!> @brief Gets the remapped buffer pointer
+!! Will do the remapping if not set already
```
thomas-robinson, line 303: Result needs to be documented

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+function get_remapped_buffer_pointer (obj) &
+result(rslt)
+    class(fmsDiagBuffer_class), target, intent(inout) :: obj
+    class(fmsDiagBuffer_class), pointer       :: objptr
```
thomas-robinson, line 307: needs to be documented

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+                     FATAL)
+         end select
+       else
+         call mpp_error ("get_data_RANGE", &
+                 "The data_RANGE value is not allocated", FATAL)
+       endif
+end function get_data_RANGE
+
+subroutine flush_buffer(buffobj)
+    class(fmsDiagBuffer_class), intent(inout) :: buffobj
```
thomas-robinson, line 402: subroutine needs documentation and so does the buffobj

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    num_buffers(5) = num_buffers(5) + 1
+    rslt = num_buffers(5)
+    buffobj%buffer_id = num_buffers(5)
+end function
+
+!> @brief Gets buffer data from buffer0d type
+!! @return copy of the buffer data
+function get_buffer_0d (obj) &
+result(rslt)
+    class (buffer0d), intent(in) :: obj
```
thomas-robinson, line 690: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    rslt = num_buffers(0)
+    buffobj%buffer_id = num_buffers(0)
+
+end function
+
+!> allocates a 1D buffer to given mold type
+integer function allocate_buffer_1d(buffobj, mold, size) &
+result(rslt)
+    class(buffer1d), intent(inout), target :: buffobj !< scalar buffer object
+    class(*),intent(in) :: mold !< allocates to the type of mold
```
thomas-robinson, line 477: You shouldn't use mold as a variable name

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    integer, intent(in) :: size !< dimension bounds
```
thomas-robinson, line 478: you shouldn't use size as a variable name.  size() is an intrinsic function.  I use `isize` personally.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        rslt = obj%buffer
+    else
+        call mpp_error(FATAL, 'get_buffer_2d: buffer not allocated')
+    endif
+end function
+!> @brief Gets buffer data from buffer3d type
+!! @return copy of the buffer data
+function get_buffer_3d (obj) &
+result(rslt)
+    class (buffer3d), intent(in) :: obj
```
thomas-robinson, line 726: need documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        rslt = obj%buffer
+    else
+        call mpp_error(FATAL, 'get_buffer_3d: buffer not allocated')
+    endif
+end function
+!> @brief Gets buffer data from buffer4d type
+!! @return copy of the buffer data
+function get_buffer_4d (obj) &
+result(rslt)
+    class (buffer4d), intent(in) :: obj
```
thomas-robinson, line 738: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        rslt = obj%buffer
+    else
+        call mpp_error(FATAL, 'get_buffer_4d: buffer not allocated')
+    endif
+end function
+!> @brief Gets buffer data from buffer5d type
+!! @return copy of the buffer data
+function get_buffer_5d (obj) &
+result(rslt)
+    class (buffer5d), intent(in) :: obj
```
thomas-robinson, line 750: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    else
+        call mpp_error(FATAL, 'get_buffer_5d: buffer not allocated')
+    endif
+end function
+
+!> @brief Initializes a buffer to a given fill value
+!! TODO default to 0?
+subroutine initialize_buffer_0d (buffobj, fillval)
+    class(buffer0d), intent(inout) :: buffobj !< any buffer object
+    class(*) :: fillval
```
thomas-robinson, line 763: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    class default
+        call mpp_error(FATAL, 'initialize buffer_3d: buffer allocated to invalid data type, this shouldnt happen')
+    end select
+
+end subroutine
+
+!> @brief Initializes a buffer to a given fill value
+subroutine initialize_buffer_4d (buffobj, fillval)
+    class(buffer4d), intent(inout) :: buffobj !< any buffer object
+    class(*) :: fillval
```
thomas-robinson, line 934: needs documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    class default
+        call mpp_error(FATAL, 'initialize buffer_4d: buffer allocated to invalid data type, this shouldnt happen')
+    end select
+
+end subroutine
+
+!> @brief Initializes a buffer to a given fill value
+subroutine initialize_buffer_5d (buffobj, fillval)
+    class(buffer5d), intent(inout) :: buffobj !< any buffer object
+    class(*) :: fillval
```
thomas-robinson, line 977: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    end select
+
+end subroutine
+
+!> @brief Add values to 0d buffer
+!! this will just call the init routine since there's only one value
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_0d(buffobj, input_data)
+    class(buffer0d), intent(inout) :: buffobj
+    class(*)            :: input_data
```
thomas-robinson, line 1022: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+    if( .not. allocated(buffobj%buffer)) call mpp_error (FATAL, 'add_to_buffer_1d: buffer not yet allocated')
+    call buffobj%initialize_buffer(input_data)
+end subroutine
+
+!> @brief Add values to 1d buffer
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_1d(buffobj, input_data)
+    class(buffer1d), intent(inout) :: buffobj
+    class(*)                       :: input_data(:)
```
thomas-robinson, line 1031: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        end select
+    end select
+    if( type_error ) call mpp_error (FATAL,'add_to_buffer_1d: mismatch between allocated buffer and input data types')
+end subroutine
+
+!> @brief Add values to 2d buffer
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_2d(buffobj, input_data)
+    class(buffer2d), intent(inout) :: buffobj
+    class(*)                       :: input_data(:,:)
```
thomas-robinson, line 1075: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        end select
+    end select
+    if( type_error ) call mpp_error (FATAL,'add_to_buffer_1d: mismatch between allocated buffer and input data types')
+end subroutine
+
+!> @brief Add values to 3d buffer
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_3d(buffobj, input_data)
+    class(buffer3d), intent(inout) :: buffobj
+    class(*)                       :: input_data(:,:,:)
```
thomas-robinson, line 1123: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        end select
+    end select
+    if( type_error ) call mpp_error (FATAL,'add_to_buffer_1d: mismatch between allocated buffer and input data types')
+end subroutine
+
+!> @brief Add values to 4d buffer
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_4d(buffobj, input_data)
+    class(buffer4d), intent(inout) :: buffobj
+    class(*)                       :: input_data(:,:,:,:)
```
thomas-robinson, line 1173: document

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1271 @@
...
+        end select
+    end select
+    if( type_error ) call mpp_error (FATAL,'add_to_buffer_4d: mismatch between allocated buffer and input data types')
+end subroutine
+
+!> @brief Add values to 5d buffer
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_5d(buffobj, input_data)
+    class(buffer5d), intent(inout) :: buffobj
+    class(*)                       :: input_data(:,:,:,:,:)
```
thomas-robinson, line 1223: document

thomas-robinson (changes requested):
First pass:
1.  You shouldn't use `mold` or `size` as a variable name.  mold is used in some intrinsics, and size is a function.  
2. You need to add some documentation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1097 @@
...
+          allocate (real(kind=r8_kind) :: this%remap_buffer(1,1,1,1,1))
+          this%remap_buffer = buff 
+      end select
+    type is (buffer1d)
+      if (.not. allocated(this%buffer)) call mpp_error(FATAL, "remap_buffer: buffer data not yet allocated")
+      if( DEBUG) print *, '1d buffer'
+      buff_bounds(1) = SIZE(this%buffer, 1)
+      select type (buff => this%buffer)
+        type is(integer(i4_kind))
+          allocate (integer(kind=i4_kind) :: this%remap_buffer(buff_bounds(1),1,1,1,1))
```
uramirez8707, line 244: I think remap_buffer should be a pointer.
Then all you will have to do is
```F90
this%remap_buffer(1:size(buff,1), 1:1, 1:1, 1:1, 1:1) => buff(1:size(buff,1))
```
you don't need the allocate and if you save the size of the buffer in the object you don't need the second select type because this works with class(*)

I don't think this will work for the buffer0d though. unless we do like this%buffer(1)

However either way this is not thread safe (@thomas-robinson?) since this%buffer will be shared by multiple threads, so the two threads will try to allocate/point remap_buffer. To make it thread safe(r), you will have to use the is_in, ie_in
https://github.com/NOAA-GFDL/FMS/blob/dmUpdate/diag_manager/diag_manager.F90#L1586 
```F90
this%remap_buffer(is:ie, 1:1, 1:1, 1:1, 1:1) => buff(is:ie)
```
That way each thread gets only the section of the variable that they need. 

or this will need to be inside an `!$OMP CRITICAL`

diag_manager/fms_diag_object.F90
```diff
@@ -73,14 +89,20 @@ subroutine fms_diag_object_init (obj,diag_subset_output)
...
-  obj%registered_variables = 0
-  obj%initialized = .true.
+  obj%buffers_initialized = fms_diag_buffer_init(obj%FMS_diag_buffers_0d, 0) .and. &
+                            fms_diag_buffer_init(obj%FMS_diag_buffers_1d, 1) .and. &
+                            fms_diag_buffer_init(obj%FMS_diag_buffers_2d, 2) .and. &
+                            fms_diag_buffer_init(obj%FMS_diag_buffers_3d, 3) .and. &
+                            fms_diag_buffer_init(obj%FMS_diag_buffers_4d, 4) .and. &
+                            fms_diag_buffer_init(obj%FMS_diag_buffers_5d, 5)
+                             
+ registered_variables = 0
```
uramirez8707, line 104: You need to pull from dmUpdate, this should be obj%registered_variables = 0
and the `registered_variables` should be removed

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1097 @@
...
+!!--------module routines
+
+!> Initializes a list of diag buffers
+!! and sets the buffer_id's to DIAG_NOT_REGISTERED
+logical function fms_diag_buffer_init(buffobjs, buff_dims)
+  type(fmsDiagBufferContainer_type), allocatable, intent(out) :: buffobjs(:) !< an array of buffer container types to allocate
+  integer, intent(in)                                          :: buff_dims !< number of dimensions needed for the buffer data
+  integer :: i
+
+  allocate(buffobjs(DEFAULT_BUFFER_LIST_SIZE))
```
uramirez8707, line 169: This has to be the size of the diag_yaml%diag_fields

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1097 @@
...
+  do i=1, SIZE(buffobjs)
+    buffobjs(i) = fms_diag_buffer_create_container(buff_dims)
+    call buffobjs(i)%diag_buffer_obj%set_buffer_id(DIAG_NOT_REGISTERED)
+  enddo
```
uramirez8707, line 173: This will be done in the register when you know how many dimensions you need

diag_manager/fms_diag_object.F90
```diff
@@ -39,6 +40,13 @@ module fms_diag_object_mod
 !TODO: Remove FMS prefix from variables in this type
   class(fmsDiagFileContainer_type), allocatable :: FMS_diag_files (:) !< array of diag files
   class(fmsDiagField_type), allocatable :: FMS_diag_fields(:) !< Array of diag fields
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_0d(:) !< array of 0 dimensional buffers 
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_1d(:) !< array of 1 dimensional buffers
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_2d(:) !< array of 2 dimensional buffers
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_3d(:) !< array of 3 dimensional buffers
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_4d(:) !< array of 4 dimensional buffers
+  type(fmsDiagBufferContainer_type), allocatable :: FMS_diag_buffers_5d(:) !< array of 5 dimensional buffers
+  integer, private :: registered_buffers(0:5) = 0 !< number of registered buffers, per dimension
```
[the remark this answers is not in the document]
rem1776: Yeah I had it this way originally, I think it was planned to have different lists for each dimension size. It'd be simpler with just the one though, @thomas-robinson @ngs333 Does that sound good to you guys?

diag_manager/fms_diag_buffer.F90
thomas-robinson: `diag_yaml%diag_fields` can be passed in from `fms_diag_object%init`.  That's the number of fields for all files, right?

diag_manager/fms_diag_buffer.F90
thomas-robinson: This isn't something I really thought about yet.  We do need to consider how this will be affected by threading.  I wonder if there's a way to make each pointer private on a thread so they aren't sharing the pointer.

diag_manager/fms_diag_buffer.F90
uramirez8707: Yeah

diag_manager/fms_diag_buffer.F90
thomas-robinson: I think you missed this one

diag_manager/fms_diag_buffer.F90
thomas-robinson: You also didn't document these

thomas-robinson (changes requested):

diag_manager/fms_diag_buffer.F90
thomas-robinson: @uramirez8707 and @rem1776 I *think* the problem is that we are storing the pointer in a module variable.  If we make the pointer-rank remapping local, it *should* be private, but then it's local, so we have to do whatever needs to be done in a local routine.

rem1776: @thomas-robinson  The pointer is in the buffer object type, not the module. I don't know what that would mean for the thread safety though. 

I was looking into the scalar remapping as well and I don't think its possible unless we do some C pointers. I was thinking maybe we could just do a "fake" scalar via a 1 element array.

diag_manager/fms_diag_buffer.F90
```diff
@@ -211,142 +198,62 @@ subroutine set_buffer_id(this, id)
 
 !> Remaps 0-5d data buffer from the given object onto a 5d array and sets it in the type
 subroutine remap_buffer(this)
-  class(fmsDiagBuffer_class), intent(inout) :: this !< any dimension buffer object
-  integer                   :: buff_bounds(5)
+  class(fmsDiagBuffer_class), target, intent(inout) :: this !< any dimension buffer object
```
ngs333, line 201: The buffer updater (the "averaging functions") in send_data_3d assumed the buffers are 4D (3D space + sample). Did we decide to move to 5D, or is this remapping for another purpose?

diag_manager/fms_diag_buffer.F90
```diff
@@ -42,25 +39,24 @@ module fms_diag_buffer_mod
 !> @brief Object that holds buffered data and other diagnostics
 !! Abstract to ensure use through its extensions(buffer0-5d types)
 type, abstract :: fmsDiagBuffer_class
-  class(*), dimension(:,:,:,:,:), allocatable :: remap_buffer !< remapped buffer data
+  class(*), dimension(:,:,:,:,:), pointer     :: remap_buffer !< remapped buffer data
```
ngs333, line 42: The buffer updater (the "averaging functions") in send_data_3d assumed the buffers are 4D (3D space + sample). Did we decide to move to 5D, or is this remapping for another purpose?

ngs333 (commented):
The name add_to_buffer does not seem appropriate for what it does. Would it be better to name it "set_buffer_data",  or "reset_buffer_data", even "initialize_buffer_data" . Also, can the existing name initialize_buffer also be given the same name as above?

rem1776:
I'm getting a GNU seg fault with the get_buffer assignments 1d and greater:
```
  class (buffer1d), intent(in) :: this !< 1D buffer object
  class(*), allocatable :: rslt(:)
  if (allocated(this%buffer)) then
    rslt = this%buffer
```
(fails on the assignment, even though the buffer is allocated and set)

The code is pretty straightforward so i don't know why this is happening, not sure if it's a bug. As a workaround we could allocate copy element by element (not sure how that would affect performace/threading), or return a pointer instead of a copy.

ngs333 (commented):
The legacy type output_filed_type (see diag_data.F90) seems to have variables `buffer`, `counter`, `count_0d`, and `num_elements` "together". Should any of these also go into the buffer class?

thomas-robinson:
> The legacy type output_filed_type (see diag_data.F90) seems to have variables `buffer`, `counter`, `count_0d`, and `num_elements` "together". Should any of these also go into the buffer class?

Probably.  Do they belong with the actual buffer or just in the parent class?

ngs333:
All four are in the same legacy class /type (output_field_type) and they
seem to need to go together. The counter and buffer are very similar and
updated similarly : they have the same rank; counter is updated (by the
math/weighting functions) "usually" when buffer is updated; and updated
using the same index extents ( for any i,j,k, sample index)  that is
applied in the do loops. Count_0d and num_elements are different (different
ranks that buffer) but still updated by the same weighting functions. There
is more documentation in  diag_data.F90.

Overall, right now it seems to me that the buffer class is the most
convenient  place to keep them even though there is no rank remapping of
them. But I suppose there are alternatives.

rem1776:
@thomas-robinson  It's updated now with the gnu workaround for the get_buffer routines. Now get_buffer is a few subroutines with an intent(out) for the buffer copy, i did some kind of type-bound procedure in the objects so it works like an interface but is called on the object with the output as an arg. I added another variable (typestr) to show the allocated type since i figured that would be needed. still need to update for the docs

@ngs333  @uramirez8707 I added the allocations for the extra members and set them to zero. I allocated the counter to the same shape as buffer, counter_0d to the last dimension size (diurnal axis?), and then num_elements is just the amount of dimensions. Let me know if any needs a change

ngs333 (approved):

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1561 @@
...
+  endif
+
+  if(allocated(this%buffer)) call mpp_error(FATAL, "allocate_buffer_2d: buffer already allocated")
+  select type (buff_type)
+    type is (integer(kind=i4_kind))
+      allocate(integer(kind=i4_kind) :: this%buffer(buff_sizes(1), buff_sizes(2)))
+      allocate(integer(kind=i4_kind) :: this%counter(buff_sizes(1), buff_sizes(2)))
+      allocate(integer(kind=i4_kind) :: this%count_0d(n_samples))
+      this%counter = 0
+      this%typestr = 'i4'
```
uramirez8707, line 389: Depending on how this will be used, maybe we should store it as a parameter integer:
https://github.com/NOAA-GFDL/FMS/blob/e8870d4fdcb97388119fa97a2c93a9cdfcc86fc3/diag_manager/diag_data.F90#L84-L87

uramirez8707 (changes requested):
Just one small change and a question.

diag_manager/fms_diag_buffer.F90
rem1776: Yeah that would be good but it might end up being removed anyway. I noticed there's another `vartype`  in the field object so we probably don't need both.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1561 @@
...
+  allocate(buff_out(size(this%buffer,1), size(this%buffer,2), size(this%buffer,3), size(this%buffer,4)))
+  select type (buff=>this%buffer)
+    type is (real(r4_kind))
+      buff_out = buff
+  end select
+end subroutine
+
+!> Gets real(r8_kind) buffer data from a 4d buffer object
+!! called through this%get_buffer(r8_outputdata)
+subroutine get_4d_real8 (this, buff_out)
```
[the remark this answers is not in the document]
rem1776: Yeah the issue was that the class(*) return values would error out after allocations, so i just replaced the class(*) get functions with routines that return specific types.

diag_manager/fms_diag_buffer.F90
```diff
@@ -1363,7 +1363,7 @@ subroutine add_to_buffer_2d(this, input_data)
   class(buffer2d), intent(inout) :: this !< allocated 2d buffer object
   class(*), intent(in)             :: input_data(:,:) !< 2d data array to copy into buffer
   integer            :: n1, n2 !< number of elements per dimension
-  logical            :: type_error
```
thomas-robinson: This needs a comment and should be set on the next line if you want to initialize it to false every time
```fortran
logical            :: type_error !< ADD A COMMENT
type_error = .false.
```

diag_manager/fms_diag_buffer.F90
```diff
@@ -1411,7 +1411,7 @@ subroutine add_to_buffer_3d(this, input_data)
   class(buffer3d), intent(inout) :: this !< allocated 3d buffer object
   class(*), intent(in)             :: input_data(:,:,:)!< 3d data array to copy into buffer
   integer            :: n1, n2, n3 !< number of elements per dimension
-  logical            :: type_error
```
thomas-robinson: see last comment

diag_manager/fms_diag_buffer.F90
```diff
@@ -1461,7 +1461,7 @@ subroutine add_to_buffer_4d(this, input_data)
   class(buffer4d), intent(inout) :: this !< allocated 4d buffer object
   class(*), intent(in)             :: input_data(:,:,:,:) !< 4d data to copy into buffer
   integer            :: n1, n2, n3, n4!< number of elements per dimension
-  logical            :: type_error
```
thomas-robinson: see last comment

diag_manager/fms_diag_buffer.F90
```diff
@@ -1511,7 +1511,7 @@ subroutine add_to_buffer_5d(this, input_data)
   class(buffer5d), intent(inout) :: this !< allocated 5d buffer object
   class(*), intent(in)             :: input_data(:,:,:,:,:) !< 5d data to copy into buffer
   integer            :: n1, n2, n3, n4, n5 !< number of elements per dimension
-  logical            :: type_error
```
thomas-robinson: see last comment

thomas-robinson (changes requested):

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1566 @@
...
+!! called through buffobj%get_buffer(r4_outputdata)
+subroutine get_0d_real4 (this, buff_out)
+  class(buffer0d), intent(in) :: this !< 0d allocated buffer object
+  real(r4_kind), allocatable, intent(out)  :: buff_out !< output of copied buffer data
+  allocate(buff_out)
+  select type (buff=>this%buffer)
+    type is (real(r4_kind))
+      buff_out = buff(1)
+    class default
+      call mpp_error(FATAL, "incorrect output type for allocated buffer")
```
thomas-robinson, line 615: I think you want the variable name and more information for this error.  These general error messages make debugging difficult

thomas-robinson: This applies to all of the error messages for the routines below this as well.  Add the field name as an input to this routine, and I think you should be able to make a more meaningful message

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1566 @@
...
+  !else
+    !call mpp_error(FATAL, 'get_buffer_0d: buffer not allocated')
+  !endif
+!end function get_buffer_0d
+
+!> Gets real(r4_kind) buffer data from a scalar buffer object
+!! called through buffobj%get_buffer(r4_outputdata)
+subroutine get_0d_real4 (this, buff_out)
+  class(buffer0d), intent(in) :: this !< 0d allocated buffer object
+  real(r4_kind), allocatable, intent(out)  :: buff_out !< output of copied buffer data
```
thomas-robinson, line 609: If this was a `class(*)` variable, couldn't you just have a single routine instead of a bunch of them?

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1566 @@
...
+      buff = fillval
+    class default
+      call mpp_error(FATAL, 'initialize_buffer_0d: mismatch between fill and buffer values')
+    end select
+  type is(integer(i4_kind))
+    select type(fillval)
+    type is(integer(i4_kind))
+      buff = fillval
+    class default
+      call mpp_error(FATAL, 'initialize_buffer_0d: mismatch between fill and buffer values')
```
thomas-robinson, line 1083: Pass in the field name and whatever other information to make this a more meaningful message.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1566 @@
...
+
+end subroutine initialize_buffer_5d
+
+!> @brief Add values to 0d buffer
+!! this will just call the init routine since there's only one value
+!! input_data must match allocated type of buffer object
+subroutine add_to_buffer_0d(this, input_data)
+  class(buffer0d), intent(inout) :: this !< allocated scalar buffer object
+  class(*), intent(in)      :: input_data !< data to copy into buffer
+  if( .not. allocated(this%buffer)) call mpp_error (FATAL, 'add_to_buffer_1d: buffer not yet allocated')
```
thomas-robinson, line 1312: Please include more information in the error message such as field name.  You will be able to use it for all of the error messages in these routines.

thomas-robinson (changes requested):
The error messages need a more detailed description of what is going wrong by including the field name and maybe even the file name.

Also please look at the other comments and address them

uramirez8707 (approved):

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+      allocate(real(kind=r4_kind) :: this%buffer(1))
+      allocate(real(kind=r4_kind) :: this%counter(1))
+      allocate(real(kind=r4_kind) :: this%count_0d(n_samples))
+      this%counter = 0
+      this%buffer_type = r4
+    type is (real(kind=r8_kind))
+      allocate(real(kind=r8_kind) :: this%buffer(1))
+      allocate(real(kind=r8_kind) :: this%counter(1))
+      allocate(real(kind=r8_kind) :: this%count_0d(n_samples))
+      this%counter = 0
```
thomas-robinson, line 330: You have an implicit casting here which may cause a warning with some compilers
```fortran
      allocate(real(kind=r4_kind) :: this%counter(1))
      this%counter = 0
```
You have to at a minimum add a `.` at the end of the 0.
```fortran
 this%counter = 0.0
```
You can also use a fancier scientific notation with the kind if you want to be specific. 

This applies to the r4 and r8, and all of the other routines that have this.  Sorry I didn't catch it earlier.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+       call mpp_error("allocate_buffer_0d", &
+           "The buff_type value passed to allocate a buffer is not a r8, r4, i8, or i4" // &
+           "for field:" // field_name_str, &
+           FATAL)
+  end select
+
+  allocate(this%num_elements(n_samples))
+  allocate(this%buffer_dims(1))
+  this%num_elements = 0
+  this%count_0d   = 0
```
thomas-robinson, line 342: How are you allowed to do this if the type is assigned in the allocate statements above?  If this is a real, the `0` will also have to be implicitly converted to a real.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+  this%buffer_dims(4) = buff_sizes(4)
+  this%buffer_dims(5) = buff_sizes(5)
+end subroutine allocate_buffer_5d
+
+!> Get routine for scalar buffers
+!! Sets the buff_out argument to the integer or real value currently stored in the buffer
+subroutine get_0d (this, buff_out, field_name)
+  class(buffer0d_type), intent(in) :: this !< 0d allocated buffer object
+  class(*), allocatable, intent(out)  :: buff_out !< output of copied buffer data
+  character(len=*), optional, intent(in) :: field_name !< optional field name for error output
```
thomas-robinson, line 679: This should be required, but it's ok i guess.

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+  character(len=128)          :: field_name_str
+  if(present(field_name)) then
+    field_name_str = field_name(1:128)
```
thomas-robinson, line 682: What will happen if someone send the field name in trimmed?
```fortran
call buffer%get_0d (buff_out, trim(field_name))
```
Is this going to break because there are 128 characters in the string?  Is it going to fill `field_name_str` with nonsense when optimized?

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+    end select
+  class default
+    call mpp_error(FATAL, 'initialize buffer_5d: buffer allocated to invalid data type, this shouldnt happen')
+  end select
+
+end subroutine initialize_buffer_5d
+
+!> @brief Add values to 0d buffer
+!! this will just call the init routine since there's only one value
+!! input_data must match allocated type of buffer object
```
thomas-robinson, line 1234: Will this print out as one line in the doxygen documentation page?  I think it needs some capitalization and punctuation

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+    class default
+      type_error = .true.
+    end select
+  end select
+  if( type_error ) call mpp_error (FATAL,'add_to_buffer_3d: mismatch between allocated buffer and input data types'//&
+                                         ' for field:'//field_name_str)
+end subroutine add_to_buffer_3d
+
+!> @brief Copy values ( from 1 to size(input_data)) into a 4d buffer object
+!! input_data must match allocated type of buffer object
```
thomas-robinson, line 1424: should these notes about the `input_data` be a `\note` ?

thomas-robinson (changes requested):

diag_manager/fms_diag_buffer.F90
rem1776: I made the field name required so now it just prints it out directly.

diag_manager/fms_diag_buffer.F90
rem1776: Yeah it will I added some capitalization and punctuation to all the routines

diag_manager/fms_diag_buffer.F90
```diff
@@ -0,0 +1,1546 @@
...
+!! not sure which approach would be better
+
+!> allocates scalar buffer data to the given buff_type
+subroutine allocate_buffer_0d(this, buff_type, diurnal_samples, field_name)
+  class(buffer0d_type), intent(inout), target :: this !< scalar buffer object
+  class(*),intent(in) :: buff_type !< allocates to the given type, value does not matter
+  integer, optional :: diurnal_samples !< number of diurnal samples, passed in from diag_yaml
+  integer :: n_samples !< number of diurnal samples, defaults to 1
+  character(len=*), optional, intent(in) :: field_name !< optional field name for error output
+  character(len=128)          :: field_name_str
```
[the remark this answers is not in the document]
rem1776: They should all be documented now

diag_manager/fms_diag_buffer.F90
rem1776: fixed, I added the fancy notation. I moved count_0d as well so it's allocated with the buffer.

diag_manager/fms_diag_buffer.F90
rem1776: I made them required

thomas-robinson (approved):

PR #1037, opened 2022-09-07
Adds zbounds to the diag_field yaml

uramirez8707:
**Description**
Moves the zbound variable from the diag_file_yaml to the diag_field_yaml to allow defining column diagnostics in a per variable basis

Fixes # (issue)

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_yaml.F90
+++ b/diag_manager/fms_diag_yaml.F90
@@ -165,7 +165,7 @@ type diagYamlFilesVar_type
   character (len=:), private, allocatable :: var_outname !< Name of the variable as written to the file
   character (len=:), private, allocatable :: var_longname !< Overwrites the long name of the variable
   character (len=:), private, allocatable :: var_units !< Overwrites the units
-  real(kind=r4_kind), private             :: var_zbounds(2)  !< indices of the z axis limits (zbegin, zend)
+  real(kind=r4_kind), private             :: var_zbounds(2)  !< The z axis limits [vert_min, vert_max]
   integer          , private              :: n_diurnal !< Number of diurnal samples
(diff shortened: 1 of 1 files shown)


uramirez8707:
These changes are needed for the converter: https://github.com/NOAA-GFDL/fms_yaml_tools/pull/10

diag_manager/fms_diag_yaml.F90
```diff
@@ -166,6 +165,7 @@ module fms_diag_yaml_mod
   character (len=:), private, allocatable :: var_outname !< Name of the variable as written to the file
   character (len=:), private, allocatable :: var_longname !< Overwrites the long name of the variable
   character (len=:), private, allocatable :: var_units !< Overwrites the units
+  real(kind=r4_kind), private             :: var_zbounds(2)  !< indices of the z axis limits (zbegin, zend)
```
thomas-robinson, line 168: If these are used as indices, then this variable should be an integer.

thomas-robinson (changes requested):
Your variable `zbounds` says that it's 2 indices, but it's set to real values.  Are these values ever given as pressures or are they just indices?  If they are indices, then the variable should be an integer array.

diag_manager/fms_diag_yaml.F90
uramirez8707: ah yeah, good catch. 
Originally we were going to use just indices, but the values are actually given as pressures so it has to be real. I forgot to update the comment ...

https://github.com/NOAA-GFDL/AM4/blob/main/run/diag_table#L509

PR #1048, opened 2022-10-04
Dump routines for diag objects

rem1776:
**Description**
adds `dump_diag_obj` and `dump_diag_yaml_obj` routines that will dump the set fields of the objects for debugging. Uses helper routines that are private for the field and file obj output. Will also write to a log file if a name is given otherwise goes to stdout

**How Has This Been Tested?**
make check, intel 21

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_field_object.F90
+++ b/diag_manager/fms_diag_field_object.F90
@@ -1018,7 +1018,7 @@ subroutine dump_field_obj (this, unit_num)
     endif
     write(unit_num, *) 'num_attributes:' ,this%num_attributes
     if( allocated(this%attributes)) then
-      do i=1, SIZE(this%attributes)
+      do i=1, this%num_attributes
         if( allocated(this%attributes(i)%att_value)) then
           select type( val => this%attributes(i)%att_value)
             type is (real(r8_kind))
--- a/diag_manager/fms_diag_field_object.F90
+++ b/diag_manager/fms_diag_field_object.F90
@@ -13,7 +13,7 @@ use diag_data_mod,  only: r8, r4, i8, i4, string, null_type_int, NO_DOMAIN
 use diag_data_mod,  only: max_field_attributes, fmsDiagAttribute_type
 use diag_data_mod,  only: diag_null, diag_not_found, diag_not_registered, diag_registered_id, &
                          &DIAG_FIELD_NOT_FOUND
-use mpp_mod, only: fatal, note, warning, mpp_error
+use mpp_mod, only: fatal, note, warning, mpp_error, mpp_pe, mpp_root_pe
 use fms_diag_yaml_mod, only:  diagYamlFilesVar_type, get_diag_fields_entries, get_diag_files_id, &
   & find_diag_field, get_num_unique_fields
 use fms_diag_axis_object_mod, only: diagDomain_t, get_domain_and_domain_type, fmsDiagAxis_type, &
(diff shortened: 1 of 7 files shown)


diag_manager/fms_diag_field_object.F90
```diff
@@ -964,5 +965,78 @@ PURE FUNCTION diag_field_id_from_name(this, module_name, field_name) &
...
+          write(unit_num, *) 'data_RANGE:' ,drange
+       type is(integer(i4_kind))
+          write(unit_num, *) 'data_RANGE:' ,drange
+       type is(integer(i8_kind))
+          write(unit_num, *) 'data_RANGE:' ,drange
+      end select
+    endif
+    write(unit_num, *) 'num_attributes:' ,this%num_attributes
+    if( allocated(this%attributes)) then
+      do i=1, SIZE(this%attributes)
```
uramirez8707, line 1021: I think this should be `this%num_attributes`
this%attributes is allocated to be
https://github.com/NOAA-GFDL/FMS/blob/bd4c4c6ac6d7d0bcedcd92eca09a6033c15c9ee2/diag_manager/fms_diag_field_object.F90#L319

but the field can have less attributes than that

diag_manager/fms_diag_file_object.F90
```diff
@@ -587,6 +588,31 @@ subroutine add_start_time(this, start_time)
...
+
+  if( allocated(this%fileobj)) write( unit_num, *)'fileobj path', this%fileobj%path
+
+  write( unit_num, *)'type_of_domain', this%type_of_domain
+  if( allocated(this%file_metadata_from_model)) write( unit_num, *) 'file_metadata_from_model', &
+                                                                    this%file_metadata_from_model
+  if( allocated(this%field_ids)) write( unit_num, *)'field_ids', this%field_ids
+  if( allocated(this%field_registered)) write( unit_num, *)'field_registered', this%field_registered
+  if( allocated(this%num_registered_fields)) write( unit_num, *)'num_registered_fields', this%num_registered_fields
+  if( allocated(this%axis_ids)) write( unit_num, *)'axis_ids', this%axis_ids
```
uramirez8707, line 611: you do `this%axis_ids(1:this%number_of_axis)`

uramirez8707 (commented):
Thank you for doing this!

thomas-robinson (commented):
THe interfaces for the dump routines are different.  Some take a filename and others are type bound and take a unit number.  I think these should all be the same?  Can the top routine call the get functions of the subobject?

diag_manager/fms_diag_file_object.F90
```diff
@@ -587,6 +588,31 @@ subroutine add_start_time(this, start_time)
 
 end subroutine
 
+!> writes out internal values for fmsDiagFile_type object
+subroutine dump_file_obj(this, unit_num)
```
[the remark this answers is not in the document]
rem1776: Yes it's cause the `dump_file_obj` and `dump_field_obj` routines are just helper methods for `dump_diag_obj`.

It cycles through the field and file lists and writes them with their dump routines, while passing in the unit num so it's all written to the same place.

diag_manager/fms_diag_field_object.F90
rem1776: I'm pissed cause i had that originally lol, thought getting the allocated size would be better

uramirez8707 (approved):

PR #1055, opened 2022-10-13
post release action for version number + fix docs trigger

rem1776:
**Description**
adds an action that will submit a pull request automatically after releases in order to add the -dev suffix to the autotools version and then assign me to review + merge it. 

This might require additional permissions in the github settings (mentioned in the [action page](https://github.com/marketplace/actions/create-pull-request)), but in my testing I was able to do it with the default settings. It always makes a new branch (with a timestamp) so shouldn't affect anything else in the repo.

One caveat is that the auto-generated PR won't run the CI workflows (at least not without additional work + messing with the tokens). Doesn't really matter though since the actual release tag will already be getting tested in the CI whenever this runs.

There is also a small bug fix here for the documentation action, apparently github thought it was a good idea to count the `release` trigger as three different events.

**How Has This Been Tested?**
Here's the  generated PR from my fork:
https://github.com/rem1776/FMS/pull/50
i just made releases on my fork till it worked so it should be the same end result

**Checklist:**
- [x] Any dependent changes have been merged and published in downstream xodules
- [x] `make distcheck` passes

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -8,3 +8,8 @@ jobs:
       run: sed -i '/20[0-9][0-9]\.[0-9][0-9]/ s/]/-dev]/' configure.ac 
     - name: Create pull request
       uses: peter-evans/create-pull-request@v4 
+      with:
+        base: main                   # creates a new branch off of main
+        branch: add-dev-post-release # name of the created branch
+        branch-suffix: timestamp     # add a timestamp to branch name
+        delete-branch: true          # delete afer merge 
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -1,0 +1,10 @@
+on: release 
+jobs:
+  add-dev-to-version:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3 
+    - name: Append version with dev
+      run: sed -i '/20[0-9][0-9]\.[0-9][0-9]/ s/]/-dev]/' configure.ac 
+    - name: Create pull request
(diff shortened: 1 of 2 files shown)


.github/workflows/version.yml
```diff
@@ -0,0 +1,20 @@
...
+    - name: Create pull request
+      uses: peter-evans/create-pull-request@v4 
+      with:
+        base: main                   # creates a new branch off of main
+        branch: add-dev-post-release # name of the created branch
+        branch-suffix: timestamp     # add a timestamp to branch name
+        delete-branch: true          # delete afer merge 
+        title: Append dev to version number post-release
+        body: automated change, adds '-dev' to the version number upon releases
+        reviewers: rem1776
```
thomas-robinson, line 20: Is the `reviewers` required?  Will rem1776 be the only required reviewer?  This could cause a potentially dangerous situation if you're on leave or away at the time of a release and a backup person is required.

rem1776: i think it adds the other reviewers based on codeowners like it does for other prs, but i mean it's always the same changes so i'll just add you and rusty to that list

thomas-robinson: Can you get rid of reviewers entirely?  Then it should just add the 3 of us, right?

laurenchilutti (approved):

.github/workflows/version.yml
rem1776: that works too

laurenchilutti (approved):

thomas-robinson (approved):

PR #1059, opened 2022-10-20
Add support for multiple tabbed section keys in the yaml emitter

rem1776:
**Description**
The current emitter doesn't allow for multiple `level2keys` like below:
```yaml
lvl2key1:
  - key1: val
lvl2key2:
  - key2: val
```
This adds support for it through making the level2key larger and adding a routine(`yaml_out_add_level2keys`) to add them in. Also adds an argument to the yaml write routine to indicate how many top level structs are written per top level key.
It stores an offset and uses a big char array to get around the 2d array+string issues that come up with c/fortran code.

Some of the changes are just me fixing the indents so i could figure out where to add lines.

**How Has This Been Tested?**
tested on amd with intel 21

**Checklist:**
- [x] `make distcheck` passes


parser/fms_yaml_output.F90
```diff
@@ -132,13 +132,14 @@ subroutine write_yaml_from_struct_3 (yamlname, a1size, keys, vals, a2size, key2,
 type (fmsYamlOutValues_type) :: val3(a3size) !< Values corresponding to keys2
 integer (c_int)        :: lvl2keyeach(lvl2_key_parameter) !< amount of key2 'blocks' to print per level2key in keys
 end subroutine write_yaml_from_struct_3
-!> adds a level 2 key to the list.
+!> adds a level 2 key (starts new tabbed section) to the list.
 !! Will print level 2 keys in the order added.
-subroutine yaml_out_add_level2key(key_name, keytype) bind(C, name="add_level2key")
+subroutine yaml_out_add_level2key(key_length, key_name, keytype) bind(C, name="add_level2key")
```
thomas-robinson, line 137: don't the lengths have to be static and the same between fortran and C?

test_fms/parser/test_output_yaml.F90
```diff
@@ -138,11 +138,11 @@ program test_output_yaml
 call fms_f2c_string (v1(1)%val1,"time to eat")
 call fms_f2c_string (k1(1)%key2,"location")
 call fms_f2c_string (v1(1)%val2,"Bridgewater, NJ")
-call yaml_out_add_level2key( "order", k1(1))
+call yaml_out_add_level2key( 5, "order", k1(1))
```
thomas-robinson, line 141: This works.  I think you should make at least one of these use the `len` function and maybe even a `trim`.  That's more likely how it will be used in practice.  I don't think you need to change them all, just one.
```fortran
order="order"
call yaml_out_add_level2key( len(trim(order)), order, k1(1))
```

parser/yaml_output_functions.c
```diff
@@ -30,7 +30,7 @@
 #define KEY_STR_LEN 255
 #define LVL2KEY_SIZE LVL2KEY_NUM*KEY_STR_LEN
 
-#define DEBUG 0
+#define DEBUG 1
```
thomas-robinson, line 33: Is this always going to print a bunch of stuff to the screen?

test_fms/parser/test_output_yaml.F90
rem1776: I'll just add a wrapper for it in fortran, that should make it so that you don't have to worry about the length.

btw theres a len_trim() fortran intrinsic, does both at once

parser/yaml_output_functions.c
rem1776: I had this on for debugging the CI, now that it passed i'll set it back to 0.

parser/fms_yaml_output.F90
rem1776: yeah i added it in to deal with gcc/CI issues. Intel will add the null char on fortran strings passed to c, gcc will not.

thomas-robinson (approved):

parser/fms_yaml_output.F90
thomas-robinson: You should always append the c_null_char to the end of c strings in fortran.
```fotran
"string"//c_null_char
```

thomas-robinson (dismissed):

thomas-robinson (approved):

PR #1080, opened 2022-11-28
diag output yaml routine

rem1776:
**Description**
adds a routine to write a yaml with information on outputted files. Prints all possible fields for files and their variables, leaving them empty if no value was set. Prints during the fms_diag_object_end call. I also needed to add a trim to one of the calls in fms_diag_file_object, otherwise gcc would mess up the filename string and crash.

The time_utils changes were unrelated, i thought i had some issues with it but just needed a reconfigure. I left the changes in though since an intrinsic name being used for a variable can mess with some compilers.
 
**How Has This Been Tested?**
intel, gcc on amd

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_yaml.F90
```diff
@@ -1401,6 +1406,380 @@ subroutine dump_diag_yaml_obj( filename )
...
+          if(DEBUG) print *, 'diag_obj', trim(diag_yaml%diag_fields(varnum_i)%var_varname)
+          if(DEBUG) print *, 'file obj', trim(diag_yaml%diag_fields(varnum_i)%var_varname)
+          if( trim(diag_yaml%diag_fields(varnum_i)%var_varname ) .eq. trim(fileptr%file_varlist(j)) .and. &
+              trim(diag_yaml%diag_fields(varnum_i)%var_fname) .eq. trim(fileptr%file_fname)) then
+            if(DEBUG) print *, 'match'
+            varptr => diag_yaml%diag_fields(varnum_i)
+            exit
+          endif
+        enddo
+        if( .not. associated(varptr)) call mpp_error(FATAL, "diag_yaml_output: var name: "// &
```
thomas-robinson, line 1532: This is just printing a variable name.  A user running into this will have no idea what's gone wrong.

diag_manager/fms_diag_yaml.F90
```diff
@@ -1401,6 +1406,380 @@ subroutine dump_diag_yaml_obj( filename )
...
+  call write_yaml_from_struct_3( 'diag_out.yaml'//c_null_char,  1, keys, vals,          &
+                                 SIZE(diag_yaml%diag_files), keys2, vals2, &
+                                 tier3size, tier3each, keys3, vals3,       &
+                                 (/size(diag_yaml%diag_files), 0, 0, 0, 0, 0, 0, 0/))
+  deallocate( keys, keys2, keys3, vals, vals2, vals3)
+
+end subroutine
+
+!> private function for getting unit string from diag_data parameter values
+character(len=7) function get_diag_unit_string( unit_param )
```
thomas-robinson, line 1742: is this pure?

diag_manager/fms_diag_yaml.F90
```diff
@@ -1401,6 +1406,380 @@ subroutine dump_diag_yaml_obj( filename )
...
+      get_diag_unit_string = 'days'
+    case (DIAG_YEARS)
+      get_diag_unit_string = 'years'
+    case default
+      get_diag_unit_string = ''
+  end select
+end function
+
+!> private function for getting reduction type string from parameter values
+character(len=7) function get_diag_reduction_string( reduction_val ) &
```
thomas-robinson, line 1762: is this pure?

thomas-robinson (changes requested):

diag_manager/fms_diag_yaml.F90
rem1776: it wasn't due to mpp_error call, but i removed that and made this one and the other one pure. I moved the error checking to outside the routine

thomas-robinson:
@rem1776 I'm looking through the PRs and found this one.  Have you worked on it at all?  It looks like there are some merge conflicts now also.

rem1776:
> @rem1776 I'm looking through the PRs and found this one. Have you worked on it at all? It looks like there are some merge conflicts now also.

Yea this is basically finished up the main issue was with the intel CI, it seems to pass when I run it with anything even in the CI's docker image but for some reason fails on the github runner so its pretty hard to figure out what's wrong.

I'll update this and take another look, apparently theres a github action that'll let you ssh into the runner so I'll give that a shot (or maybe a new intel image will fix it).

test_fms/diag_manager/test_diag_manager2.sh
```diff
@@ -804,6 +804,420 @@ _EOF
...
+    tile: -999
+    corner1:
+    corner2:
+    corner3:
+    corner4:
+  global_meta:
+  - {}
+- file_name: file8%4yr%2mo%2dy%2hr%2min
+  freq: 1 1 1 -999 -999 -999 -999 -999 -999 -999 -999 -999
+  freq_units: hours hours hours null null null null null null null null null
```
uramirez8707, line 1119: Why all the nulls and -999?

rem1776: They are just printing what they got initialized to (DIAG_NULL=-999), and then i had the units just write out null. Do you think we should just leave out any defaults? I had debugging in mind but there's probably better ways to do that anyway.

uramirez8707: yeah I think i we should leave out any defaults

rem1776: @uramirez8707 Should be fixed up now

diag_manager/fms_diag_yaml.F90
```diff
@@ -1535,6 +1560,423 @@ subroutine dump_diag_yaml_obj( filename )
...
+    call fms_f2c_string(keys2(i)%key8, 'start_time')
+    call fms_f2c_string(keys2(i)%key9, 'file_duration')
+    call fms_f2c_string(keys2(i)%key10, 'file_duration_units')
+
+    call fms_f2c_string(vals2(i)%val1, fileptr%file_fname)
+    call fms_f2c_string(vals2(i)%val5, fileptr%file_unlimdim)
+    call fms_f2c_string(vals2(i)%val4, get_diag_unit_string((/fileptr%file_timeunit/)))
+    tmpstr1 = ''
+    do k=1, SIZE(fileptr%file_freq)
+        if(fileptr%file_freq(k) .eq. -999) exit
```
uramirez8707, line 1641: rather than hard codding -999, can we use `diag_null`?

thomas-robinson (approved):

PR #1088, opened 2022-12-08
Accept data

thomas-robinson:
**Description**
Creates accept data which will be called by send_data.  Accept data currently buffers the data sent to send_data if the call is in an openmp region.  

This needs @ngs333 to add his work to prepare the field for it's math functions.

Fixes #

**How Has This Been Tested?**

**Checklist:**
- [x] `make distcheck` passes


ngs333:
@thomas-robinson 
I see three .F90 files as part of the PR. I thought you also made a diagram for this?

diag_manager/fms_diag_object.F90
```diff
@@ -425,6 +429,58 @@ FUNCTION fms_diag_axis_init(this, axis_name, axis_data, units, cart_name, long_n
...
+!! is being used, then the processing can be done and stored in the buffer object.  The hope is that
+!! the increase in memory footprint related to buffering can be handled by the shared memory of the
+!! multithreaded case.
+!! \note If some of the diag manager is offloaded in the future, then it should be treated similarly
+!! to the multi-threaded option for processing later
+logical function fms_diag_accept_data (this, diag_field_id, field, time, is_in, js_in, ks_in, &
+                  mask, rmask, ie_in, je_in, ke_in, weight, err_msg)
+  class(fmsDiagObject_type),TARGET,INTENT(inout):: this !< Diaj_obj to fill
+  INTEGER, INTENT(in) :: diag_field_id !< The ID of the input diagnostic field
+  CLASS(*), DIMENSION(:,:,:,:), INTENT(in) :: field !< The data for the input diagnostic
```
ngs333, line 443: I think a better name for "field" here  is "field_data" as the word field is practically synonymous with fmsDiagField_type.

diag_manager/fms_diag_object.F90
```diff
@@ -425,6 +429,58 @@ FUNCTION fms_diag_axis_init(this, axis_name, axis_data, units, cart_name, long_n
...
+  omp_level = omp_get_level()
+#endif
+!If this is true, buffer data
+  main_if: if ((omp_num_threads > 1 .AND. omp_level > 0) .or. buffer_the_data) then
+!> Buffer the data
+    call this%FMS_diag_fields(diag_field_id)%set_data_buffer(field, FMS_diag_object%diag_axis)
+    fms_diag_accept_data = .TRUE.
+    return
+  else
+!!TODO: Loop through fields and do averages/math functions
```
ngs333, line 474: This is a big todo.

diag_manager/fms_diag_object.F90
```diff
@@ -425,6 +429,58 @@ FUNCTION fms_diag_axis_init(this, axis_name, axis_data, units, cart_name, long_n
...
+    fms_diag_accept_data = .TRUE.
+    return
+  end if main_if
+!> Return false if nothing is done
+  fms_diag_accept_data = .FALSE.
+  return
+#endif
+end function fms_diag_accept_data
+!! TODO: This entire routine
```
ngs333, line 483: This is another big TODO

ngs333 (approved):

thomas-robinson:
> @thomas-robinson I see three .F90 files as part of the PR. I thought you also made a diagram for this?

The diagram is in a different PR

diag_manager/fms_diag_object.F90
thomas-robinson: ...With a big if-else

diag_manager/diag_manager.F90
```diff
@@ -1649,13 +1649,51 @@ LOGICAL FUNCTION send_data_3d(diag_field_id, field, time, is_in, js_in, ks_in, &
     SELECT TYPE (field)
     TYPE IS (real(kind=r4_kind))
        field_out = field
+       if (use_modern_diag) then
+         allocate(real(kind=r4_kind) :: field_modern(SIZE(field,1),SIZE(field,2),SIZE(field,3),1))
```
uramirez8707, line 1653: Can we use a pointer here instead?

diag_manager/fms_diag_field_object.F90
```diff
@@ -65,7 +65,12 @@ module fms_diag_field_object_mod
      integer, allocatable, private                    :: area, volume      !< The Area and Volume
      class(*), allocatable, private                   :: missing_value     !< The missing fill value
      class(*), allocatable, private                   :: data_RANGE(:)     !< The range of the variable data
-    contains
+     class(*), allocatable, dimension(:,:,:,:), private :: data_buffer     !< Buffer for field data
```
rem1776, line 68: Are you using this instead of a buffer object, or is this just for the initial call? I think in the original design the buffer id is stored here and then the actual buffers are in an array in the diag object, not sure if that makes more sense.

thomas-robinson: This buffers the data from `accept_data` to be used in `fms_diag_send_complete`.  When the math functions are called, the buffer object buffers are filled in.  There are 2 possibilities:
1. Accept data is called and the math functions are performed in accept data.  In this case, this buffer is not allocated or used.
2. Accept data is called and the only thing that happens is this data buffer is allocated and the data is stored here.  The math functions will be done in fms_diag_send_complete.  This is a big change from how `send_data` in the current diag manager works.  

Why do this at all?  This will 
1. Make accept_data thread safe (which is why there is a check for threading).
2. Allow us to test/use openmp/`do concurrent` for the math part of the diag manager.
3. Provide offloading for the most computationally expensive part of the diag manager in the future.

diag_manager/diag_manager.F90
thomas-robinson: image

thomas-robinson: @uramirez8707 the `field` needs to be contiguous.  Will it always be contiguous?

diag_manager/fms_diag_field_object.F90
```diff
@@ -359,7 +368,115 @@ subroutine set_vartype(objin , var)
...
+  end select
+end subroutine set_data_buffer
+!> Allocates the global data buffer for a given field using a single thread. Returns true when the
+!! buffer is allocated
+logical function allocate_data_buffer(this, input_data, diag_axis)
+  class (fmsDiagField_type), target, intent(inout):: this !< The field object
+  class(*), dimension(:,:,:,:), intent(in) :: input_data !< The input array
+  class(fmsDiagAxisContainer_type),intent(in)   :: diag_axis(:) !< Array of diag_axis
+  integer :: naxes !< The number of axes in the field
+  integer, parameter :: ndims = 4
```
uramirez8707, line 419: what about the diurnal axis? And i think the math functions are going to assume that the diurnal axis is the fourth dimension?

uramirez8707: Actually this is the data is being sent in `send_data`, so the diurnal axis does not matter here.

thomas-robinson:
@rem1776 @uramirez8707 is this ready to merge?

diag_manager/fms_diag_field_object.F90
```diff
@@ -359,7 +368,115 @@ subroutine set_vartype(objin , var)
...
+  if (.not.this%data_buffer_allocated) &
+    call mpp_error ("set_data_buffer", "The data buffer for the field "//trim(this%varname)//" was unable to be "//&
+      "allocated.", FATAL)
+
+!> Buffer a copy of the data
+  select type (input_data)
+    type is (real(kind=r4_kind))
+      select type (db => this%data_buffer)
+        type is (real(kind=r4_kind))
+          db = input_data
```
uramirez8707, line 390: `input_data` contains the data for the 1 thread (only 1 section of the global data)
this%data_buffer is allocated to be the size of the global data

so this is not going to work because the sizes may not be the same for more than 1 thread.

so we need to pass in `is_in, js_in, ks_in, ie_in, je_in, ke_in` from fms_diag_accept_data then do
```F90
db(is_in:ie_in, js_in:je_in, ks_in:ke_in, :) = input_data
```

uramirez8707 (changes requested):

diag_manager/diag_manager.F90
thomas-robinson: ::fingers crossed::

diag_manager/fms_diag_object.F90
```diff
@@ -425,25 +429,101 @@ FUNCTION fms_diag_axis_init(this, axis_name, axis_data, units, cart_name, long_n
...
+        math = .false. !TODO: replace this with real thing
+        calling_math: if (math) then
+          !!TODO: call math functions !!
+        endif calling_math
+        !> Clean up, clean up, everybody everywhere
+        if (associated(diag_field)) nullify(diag_field)
+      enddo field_loop
+      !> Clean up, clean up, everybody do your share
+      if (allocated(file_field_ids)) deallocate(file_field_ids)
+    endif field_outer_if
```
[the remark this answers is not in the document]
thomas-robinson: Sorry.  I think that's something we will have to deal with.

diag_manager/fms_diag_field_object.F90
thomas-robinson: I updated it to look like how you have it.  Do I need to put the indicies on input_data?  I don't think so.

uramirez8707: I don’t think so

uramirez8707 (approved):

PR #1090, opened 2022-12-12
add log_diag_field_info to dmUpdate

rem1776:
**Description**
adds the log_diag_field_info calls for the modern diag's register routines.

**How Has This Been Tested?**
intel and gnu on amd

**Checklist:**
- [x] `make distcheck` passes


diag_manager/diag_manager.F90
```diff
@@ -436,13 +436,20 @@ INTEGER FUNCTION register_diag_field_array(module_name, field_name, axes, init_t
...
+         allow_log = .NOT.do_not_log
+      ELSE
+         allow_log = .TRUE.
+      END IF
       register_diag_field_array = fms_diag_object%fms_register_diag_field_array( &
        & module_name, field_name, axes, init_time, long_name=long_name, &
        & units=units, missing_value=missing_value, var_range=range, mask_variant=mask_variant, &
-       & standard_name=standard_name, verbose=verbose, do_not_log=do_not_log, err_msg=err_msg, &
-       & interp_method=interp_method, tile_count=tile_count, area=area, volume=volume, realm=realm)
+       & standard_name=standard_name, verbose=verbose, do_not_log=.not.(do_diag_field_log.AND.allow_log), &
```
thomas-robinson, line 451: This is adding unneeded complexity.  do_not_log is optional, so if it's not present, then it won't be passed on because it's optional in the next routine (register_diag_field_array) as well.  

You should move this logic to where it's actually used instead of now actually passing something to do_not_log.

diag_manager/diag_manager.F90
```diff
@@ -487,10 +495,16 @@ INTEGER FUNCTION register_static_field(module_name, field_name, axes, long_name,
...
+      IF ( PRESENT(do_not_log) ) THEN
+         allow_log = .NOT.do_not_log
+      ELSE
+         allow_log = .TRUE.
+      END IF
       register_static_field = fms_diag_object%fms_register_static_field(module_name, field_name, axes, &
        & long_name=long_name, units=units, missing_value=missing_value, range=range, mask_variant=mask_variant, &
-       & standard_name=standard_name, dynamic=DYNAMIC, do_not_log=do_not_log, interp_method=interp_method,&
-       & tile_count=tile_count, area=area, volume=volume, realm=realm)
+       & standard_name=standard_name, dynamic=DYNAMIC, do_not_log=.not.(do_diag_field_log.AND.allow_log), &
```
thomas-robinson, line 506: This is the same comment as above.

thomas-robinson (changes requested):
By trying to make the code more easy to understand by renaming the variable, I think you added complexity that requires mental gymnastics and unnecessary code execution further down the call stack.

diag_manager/fms_diag_object.F90
```diff
@@ -673,4 +686,141 @@ subroutine dump_diag_obj( filename )
...
+!!     then a line briefly describing diagnostic field is added to
+!!     the log file.  Normally users should not call this subroutine
+!!     directly, since it is called by register_static_field and
+!!     register_diag_field if do_not_log is not set to .TRUE..  It is
+!!     used, however, in LM3 to avoid excessive logs due to the
+!!     number of fields registered for each of the tile types.  LM3
+!!     code uses a do_not_log parameter in the registration calls,
+!!     and subsequently calls this subroutine to log field information
+!!     under a generic name.
+SUBROUTINE fms_log_field_info(module_name, field_name, axes, long_name, units,&
```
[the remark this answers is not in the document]
rem1776: This routine is only used by the new. Its pretty similar to the old but some info needed to come from the diag object, also it has the changes for the separator and writes the header instead of doing that in the init.

My thinking was that if theres a new register routine in a separate file it would make sense to add a new log routine too. I think I'll change it though, it will simplifiy the do_not_log logic as well if I just make some changes to the original and use that for both.

diag_manager/diag_manager.F90
rem1776: The logic basically prints the log if either the nml flag or do_not_log is set to output , I added it here cause I needed the nml flag and this was the last place where it would be accessible.

I'm gonna change it around anyway so I might take out the do_not_log from the object's register routine if it ends up being useless (which seems likely)

rem1776:
> @rem1776 This kind of fell by the wayside. Is this PR needed anymore?

Yeah we'll still need some changes to get the routine working with the objects, I'll update this.

thomas-robinson (approved):

PR #1100, opened 2023-01-03
Memory leak fix: pointers in interpolator and horiz_interp

### LINKED ISSUE #1099: Allocated Pointers: Potential Memory Leak Issue
In horiz_interp and interpolator, some variables with the type of pointer are being allocated storage space with the use of the allocate statement.  With newer versions of Fortran, we now have access to the allocatable type.  The allocatable type is safer to use as storage will automatically be deallocated at end of a program.  With the pointer type, we can potentially cause memory leaks if we do not deallocate the pointer.

Change any allocated pointers to allocatables.

N/A

**Additional context**
N/A

laurenchilutti:
**Description**
In interpolator and horiz_interp, some pointers are being allocated.  This PR changes the type from pointer to allocatable.  Any use of `if associated()` has now been changed to `if allocated()` to accompany this change.  There are also instances in the code where we were declaring, for example:
```
interface assignment(=)
  module procedure horiz_interp_type_eq
end interface
```
And in that subroutine horiz_interp_type_eq, for example:
```
horiz_interp_out%faci => horiz_interp_in%faci
```

I have changed these pointer assignment operators to an equals, with the  foresight that this may cause issues.  With the old functionality, anytime `horiz_interp_out` is modified, the value of `horiz_interp_in` is also modified.  With my changes this will not be the case. I tested with the land model and did not run into any issues with this but we will test all models with alpha2 testing.

Fixes #1099 

**How Has This Been Tested?**
Compiled with GNU and Intel, ran with the land model and answers reproduce

**Checklist:**
- [x] `make distcheck` passes


laurenchilutti:
@rem1776 Could @ganganoaa be added as a reviewer for this PR? Thanks

horiz_interp/horiz_interp_type.F90
```diff
@@ -189,35 +189,35 @@ subroutine horiz_interp_type_eq(horiz_interp_out, horiz_interp_in)
       call mpp_error(FATAL,'horiz_interp_type_eq: horiz_interp_type variable on right hand side is unassigned')
     endif
 
-    horiz_interp_out%faci            => horiz_interp_in%faci
```
thomas-robinson, line 192: The functionality of this routine is changing.  Can you add doxygen comments to this clearly stating that this routine creates a *copy* of the object.  Also doxygen comments for the variables.

horiz_interp/horiz_interp_type.F90
```diff
@@ -189,35 +189,35 @@ subroutine horiz_interp_type_eq(horiz_interp_out, horiz_interp_in)
...
+    horiz_interp_out%ilon            =  horiz_interp_in%ilon
+    horiz_interp_out%jlat            =  horiz_interp_in%jlat
+    horiz_interp_out%area_src        =  horiz_interp_in%area_src
+    horiz_interp_out%area_dst        =  horiz_interp_in%area_dst
+    horiz_interp_out%wti             =  horiz_interp_in%wti
+    horiz_interp_out%wtj             =  horiz_interp_in%wtj
+    horiz_interp_out%i_lon           =  horiz_interp_in%i_lon
+    horiz_interp_out%j_lat           =  horiz_interp_in%j_lat
+    horiz_interp_out%src_dist        =  horiz_interp_in%src_dist
     horiz_interp_out%found_neighbors => horiz_interp_in%found_neighbors
```
thomas-robinson, line 203: Did you miss this pointer?

thomas-robinson (changes requested):

ganganoaa:
Yes. That was missed.

horiz_interp_out%found_neighbors => horiz_interp_in%found_neighbors


Ganga

horiz_interp/horiz_interp_type.F90
laurenchilutti: I couln't find any use of this pointer in FMS.  Specifically if you grep the repository, you only get the following lines:
https://github.com/NOAA-GFDL/FMS/blob/203c8bf464ff26fe0fe39b1451caedd026bbce55/horiz_interp/horiz_interp_type.F90#L72
https://github.com/NOAA-GFDL/FMS/blob/203c8bf464ff26fe0fe39b1451caedd026bbce55/horiz_interp/horiz_interp_type.F90#L203

So, since it wasn't a pointer being allocated I did not touch that variable at all.

horiz_interp/horiz_interp_type.F90
laurenchilutti: Let me know if these changes are sufficient

horiz_interp/horiz_interp_type.F90
laurenchilutti: Does my reasoning make sense?

ganganoaa:
I think it is safe since it is not allocated any where.

horiz_interp/horiz_interp_type.F90
```diff
@@ -181,43 +181,44 @@ subroutine stats ( dat, low, high, avg, miss, missing_value, mask )
...
+    horiz_interp_out%facj            =  horiz_interp_in%facj
+    horiz_interp_out%ilon            =  horiz_interp_in%ilon
+    horiz_interp_out%jlat            =  horiz_interp_in%jlat
+    horiz_interp_out%area_src        =  horiz_interp_in%area_src
+    horiz_interp_out%area_dst        =  horiz_interp_in%area_dst
+    horiz_interp_out%wti             =  horiz_interp_in%wti
+    horiz_interp_out%wtj             =  horiz_interp_in%wtj
+    horiz_interp_out%i_lon           =  horiz_interp_in%i_lon
+    horiz_interp_out%j_lat           =  horiz_interp_in%j_lat
+    horiz_interp_out%src_dist        =  horiz_interp_in%src_dist
```
ganganoaa, line 203: This and all the above pointers including one below are not allocated but their assignments are different compared to `found_neighbors' case.

horiz_interp/horiz_interp_type.F90
thomas-robinson: It seems odd because the routine is creating a copy except for this one variable.

horiz_interp/horiz_interp_type.F90
thomas-robinson, line 203: I think you need to check if the variable is allocated before you set it.
```fortran
    if (allocated(horiz_interp_in%wtj)) horiz_interp_out%wtj             =  horiz_interp_in%wtj
```

thomas-robinson (changes requested):
I think you need to add `if allocated` statements and change that last pointer to an allocatable unless it's being used somewhere else as a pointer.

horiz_interp/horiz_interp_type.F90
thomas-robinson: Where is that done?

laurenchilutti: I see what you are saying - do you think it would also be necessary to check if a variable is allocated before allocating?

```
if (allocated(variable)) deallocate(variable)
allocate(variable)
```

laurenchilutti: They are allocated in other horiz_interp files:

e.g.
https://github.com/laurenchilutti/FMS/blob/42e061f96343dd8e2df0d0338a372dc8c37b61a9/horiz_interp/horiz_interp_conserve.F90#L168-L171

https://github.com/laurenchilutti/FMS/blob/42e061f96343dd8e2df0d0338a372dc8c37b61a9/horiz_interp/horiz_interp_bilinear.F90#L288-L291

https://github.com/laurenchilutti/FMS/blob/42e061f96343dd8e2df0d0338a372dc8c37b61a9/horiz_interp/horiz_interp_spherical.F90#L217-L220

laurenchilutti: But, found_neighbors is used and allocated nowhere in FMS that I can find

ganganoaa: It is done here:
interface assignment(=)
  module procedure horiz_interp_type_eq
end interface

ganganoaa: Tom was right. It will fail if you try to assign an unallocated variable. I tested it in a simple code.

laurenchilutti: @thomas-robinson I wrote a test program to see if we would get an error if `horiz_interp_type_out` is unallocated, and there is no error and the program runs fine.  I found a section of the Fortran 2003 standard that seems to indicate that If a member of horiz_interp_out is unallocated, it will be allocated with the same shape as horiz_interp_in.

You can see the section of the fortran 2003 standard I am referring to [here](https://wg5-fortran.org/N1601-N1650/N1601.pdf#page=155) in Section 7.4.1.3 Interpretation of intrinsic assignments

ganganoaa:
Lauren, the fortran reference verified my assumption and what I have 
tested recently.

Thanks.

horiz_interp/horiz_interp_type.F90
rem1776: I'd bet its just unused in general, you don't get warnings for unused fields in derived types so it could have never been used and no one noticed. I think we can probably remove it unless we can find some code that actually uses it.

ganganoaa: If we remove this field assignment, it will violate principle of object-to-object copy but if it is never used, we can remove. We can add something like this:

if (associated(in%var)) 
    if (associated(out%var)) deallocate(out%var)
    allocate(out%var(size(in%var))
    out%var = in%var
else
    out%var => NULL()
end if

laurenchilutti: @thomas-robinson @ganganoaa @rem1776 i have not done anything to found neighbors yet.  the options as I see it are:
- removing it seems like something that would require a lot of testing since its possible some model somewhere is using it.  
- changing it from pointer to allocatable change this pointer assignment to equals
- leave it as is
I don't feel like I have enough familiarity with interpolator code to make this decision.  How should I proceed?   Who can I consult on what to do?

horiz_interp/horiz_interp_type.F90
laurenchilutti: @thomas-robinson Can you take a look at my previous comment on this thread and let me know if you agree with my reasoning?

horiz_interp/horiz_interp_type.F90
thomas-robinson: do this:
```fortran
if (allocated(horiz_interp_in%found_neighbors))horiz_interp_out%found_neighbors = horiz_interp_in%found_neighbors
```
and make `found_neighbors` allocatable instead of a pointer.  This is a `=` routine, so it should be making a copy, not providing pointers.  

This is a modernization PR, not a clean up.  No need to remove it here.  We should note it for potential clean up in a future release.

laurenchilutti:
I have updated the `found_neighbors` field to be allocatable.  
Lines 189-191 in hoiz_interp_type.F90 are checking that `horiz_interp_in%I_am_initialized` is set to `.TRUE.` `%I_am_initialized` is set to `.TRUE.` when these fields are allocated, so I will not be adding an `if(allocated(in%field))` before each field assignment in `horiz_interp_type_eq`.
I added `if (allocated(horiz_interp_in%found_neighbors))` before we equate the `found_neighbors` field because this is not allocated in FMS.

rem1776 (approved):

rem1776:
@laurenchilutti It looks like the CI glitched out or something on the last commit. I can see that it ran and passed yesterday [here](https://github.com/NOAA-GFDL/FMS/actions?query=branch%3Apointers) but the status never updated on the actual PR. I tried rerunning the latest jobs but it still didn't update.

I've seen something similar before it might've been a github action outage or something at the time you pushed it, so a new commit could fix it. Maybe you can revert/rebase and then re-push the last commit? Although it might be easier to just make some kind of small change and commit that.

laurenchilutti:
@rem1776 I just added a minor change to a comment in the latest commit to relaunch the CI.

horiz_interp/horiz_interp_type.F90
thomas-robinson: You need to do this for all of the allocatable variables.

thomas-robinson (changes requested):
You need to check that all of the variables are allocated in the assignment routine.  If they aren't allocated, I think that routine will crash.

ganganoaa:
I think the check before the assignments,

if(.not.horiz_interp_in%I_am_initialized) then

callmpp_error(FATAL,'horiz_interp_type_eq: horiz_interp_type variable on 
right hand side is unassigned')

endif

will suffice the requirement. If needed, extra safe-guard can be inserted.

Ganga

PR #1105, opened 2023-01-11
Send math legacy plus review aid dummy pull request

ngs333:
This PR is largely for the field buffer updating math routines to be when selected in 
 a namelist option in the legacy diag manager. Key parts of the presented code is 
intend to  supports the modern diag manager out of the box.
 Functionality that was originally in send_data_3d was refactored and moved
 to their own functions. The added capability of for direct 3D and 4D fields (with
 corresponding 4D and 5D) buffers used in the legacy and modern diag manager, respectively.
 The functions support R4, R8, I4 and I8 fields and buffers (matching in type) , 
  Integer versions of the functions are not allowed in the legacy diag manager and 
  are commented out for this review . 

  Most of the code refactored from send_data_3d is in these files:
        1. fms_diag_fieldbuff_update.F90
        2. include/fms_diag_fieldbuff_update.inc
        3. include/fms_diag_fieldbuff_update.fh

  To allow the math functions to work with both the legacy and the modern diag_manager,
  some refactoring of bounds checking functions was also necessary. This was used also
  as an opportunity to remove repetitive code, and some utilities related to array index bounding boxes were introduced. These functions are largely in diag_util.F90
  
 Some conventions:
 Some of the code has comments that begin with !TODO: .Those that on the same line also have
 text (MDM) , I expect that they be resolved in the modern diag manager.

 Some missed opportunities :
-  Given the existence of the R4 and R8 generated buffer update functions, it should be possible to
 remove the copy of the field array used in send_data_3d that was introduced in the EMC mixed mode merge.
- I was desired to set a pointer to the lowest level (elemental) buffer update math function.
There were issues implementing this (see  this would have removed the innermost if/then/else construct in the buffer update loops. See fms_diag_outfield_type in file fms_daig_outfield.F90.


thomas-robinson:
You still have diag rewrite stuff in your tests:
```
test_diag_update_buffer.F90(86): error #6457: This derived type name has not been declared.   [FMSDIAGFIELD_TYPE]
   TYPE(fmsDiagField_type) :: field
```

ngs333:
Opening new PR.

PR #1163, opened 2023-03-20
GCC Class* bug workaround for character arrays

### LINKED ISSUE #1164: GCC Class* bug with character arrays
**Describe the bug**
When reading and writing string arrays to a netcdf file with fms2io, the GCC 12.1.0 and 12.2.0 compilers can't figure out the len_trim of some members of the string array and write out junk to the end of the string.

**To Reproduce**
do `make check` and see test_mosaic2 fail.

**Expected behavior**
Fortran should be able to decipher a character array within a class* select type correctly.

**System Environment**
tested with GCC 12.1.0 and 12.2.0.

**Additional context**
This is a compiler bug, tracking this as an FMS issue because we have a workaround to implement via a PR.
GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109345

laurenchilutti:
**Description**
When reading and writing string arrays to a netcdf file with fms2io, the GCC 12.1.0 and 12.2.0 compilers can't figure out the `len_trim` of some members of the string array and write out junk to the end of the string. This is implementing a workaround where we moved all of the code for a character array from the select type statement to a separate function and calling that function. Doing this while also initializing our strings to blank strings fixes the issue. 

Fixes #1164 

**How Has This Been Tested?**
Tested with GCC on GFDL's dev box

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/fms2_io/include/array_utils.inc
+++ b/fms2_io/include/array_utils.inc
@@ -23,10 +23,11 @@
 !> @{
 
 !> @brief Allocate arrays using an input array of sizes.
-subroutine allocate_array_i4_kind_1d(buf, sizes)
+subroutine allocate_array_i4_kind_1d(buf, sizes, initialize)
 
   integer(kind=i4_kind), dimension(:), allocatable, intent(inout) :: buf !< Array that will be allocated.
   integer, dimension(1), intent(in) :: sizes !< Array of dimension sizes.
+  logical, intent(in), optional :: initialize !< Optional argument when true will initialize a blank string Not Used
 
   if (allocated(buf)) then
     deallocate(buf)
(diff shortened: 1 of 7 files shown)


fms2_io/include/array_utils.inc
```diff
@@ -23,10 +23,11 @@
 !> @{
 
 !> @brief Allocate arrays using an input array of sizes.
-subroutine allocate_array_i4_kind_1d(buf, sizes)
+subroutine allocate_array_i4_kind_1d(buf, sizes, initialize)
```
rem1776, line 26: Are the changes in this file supposed to be here? I don't see the added arg used anywhere.

fms2_io/include/array_utils_char.inc
```diff
@@ -23,79 +23,202 @@
...
 
   if (allocated(buf)) then
     deallocate(buf)
   endif
   allocate(buf(sizes(1)))
+
+  if (init) then
+    do i = 1, sizes(1)
+      do c = 1, len(buf(i))
+        buf(i)(c:c) = " "
```
rem1776, line 46: Does this workaround work without looping though? Like if you just did buf = " " or even buf(i) = " "

laurenchilutti: This is the only way to ensure every character in the string array is initialized to zero.  I tested both buf = " " and buf(i) = " " and those will not properly initialize all characters in all elements of the string array.

fms2_io/include/array_utils.inc
laurenchilutti: I had originally thought that all modules within the same interface needed the same number of arguments but I was mistaken.  So this will all be removed in my next commit.

rem1776 (approved):

PR #1188, opened 2023-04-10
Modern diag_manager: Fixes related to static and scalar fields

uramirez8707:
**Description**

- Fix so that scalar variables that depend on time can be a function of the unlimited dimension
- Fix to work around the issue that the old diag manager did not have the ability to register scalar static fields
- Fix to reproduce the `get_diag_field_id` behavior in the old diag manager
- Fix to prevent `nv` to be written twice, if it is defined as an axis in the model code
- Fix so that static files are only written once in the beginning of the run

Fixes # (issue)

**How Has This Been Tested?**
This + #1177 + #1176 run with the am4 cmip diag_table

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_file_object.F90
+++ b/diag_manager/fms_diag_file_object.F90
@@ -329,6 +329,9 @@ subroutine set_file_time_ops(this, VarYaml, is_static)
   type (diagYamlFilesVar_type), intent(in)    :: VarYaml   !< The variable's yaml file
   logical,                      intent(in)    :: is_static !< Flag indicating if variable is static
 
+  !< Go away if the file is static
+  if (this%is_static) return
+
   if (this%time_ops) then
     if (is_static) return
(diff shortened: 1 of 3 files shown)


diag_manager/fms_diag_file_object.F90
```diff
@@ -1203,6 +1211,16 @@ subroutine increase_unlimited_dimension(this)
   this%FMS_diag_file%unlimited_dimension = this%FMS_diag_file%unlimited_dimension + 1
 end subroutine increase_unlimited_dimension
 
+!> \brief Get the unlimited dimension that is in the file
+!! \return The unlimited dimension
```
thomas-robinson, line 1215: This is an integer.  Is this an ID?  Is this the current value of the dimension?

thomas-robinson (changes requested):

thomas-robinson (approved):

PR #1191, opened 2023-04-12
Mixed precision random_numbers_mod

J-Lentz:
Initial implementation of mixed precision random_numbers_mod and unit tests.

**Checklist:**
- [x] `make distcheck` passes


test_fms/random_numbers/test_random_numbers.F90
```diff
@@ -0,0 +1,279 @@
...
+use time_manager_mod, only: time_type, get_date, set_date, set_calendar_type, JULIAN
+use random_numbers_mod
+
+implicit none
+
+integer, parameter :: n_moments = 1000 !> Highest-order moment to test
+integer, parameter :: n0_1d = 2500 !> Initial length of 1D random sample vector
+integer, parameter :: n0_2d = 50 !> Initial dimensions of 2D random sample array
+
+integer, parameter :: seeds(0:*) = [0, -5, 3] !> Seed constants
```
mlee03, line 16: You're a Fortran guru, this is the first time I'm seeing something like this lol.  Does this notation work with all compilers?

test_fms/random_numbers/test_random_numbers.F90
```diff
@@ -0,0 +1,279 @@
...
+
+real(k), dimension(n_moments) :: moment_mu !> Expected moment values
+real(k), dimension(n_moments) :: moment_sigma !> Standard deviations of sample moments
+
+type(time_type) :: now !> Current date and time
+
+call fms_init
+
+if (mpp_npes() .ne. 4) then
+  call mpp_error(fatal, "The random_numbers unit test requires four PEs")
```
mlee03, line 27: Should it be FATAL or should it be a NOTE and then exit the program?

test_fms/random_numbers/test_random_numbers.F90
```diff
@@ -0,0 +1,279 @@
...
+    case (3)
+      stream = initializeRandomNumberStream(constructSeed(seeds(1), seeds(2), now))
+      call test_getRandomNumbers_dispatch(stream)
+    case default
+      call mpp_error(fatal, "Unexpected PE: " // string(mpp_pe()))
+  end select
+end subroutine test_getRandomNumbers
+
+! Expression for the expectation value of the i-th raw moment
+#define CALC_MOMENT_(i) (1._k / real(i + 1, k))
```
mlee03, line 112: where did this equation come from?

test_fms/random_numbers/test_random_numbers.F90
```diff
@@ -0,0 +1,279 @@
...
+    do i = 1, n
+      call check_bounds(arr(i, j))
+    enddo
+  enddo
+
+  test_sample_2d = compare_sample_moments(reshape(arr, [size(arr)]))
+end function test_sample_2d
+
+!> Check that the first 1,000 moments of a sample are within one standard
+!> deviation of their expected values
```
mlee03, line 216: why one standard deviation?  do we know what the distribution function looks like?

test_fms/random_numbers/test_random_numbers.F90
J-Lentz: The _m_-th raw moment of a distribution, by definition, is the integral of _x_^_m_ times _f(x)_, where _f(x)_ is the distribution's probability density function. For the uniform distribution, _f(x)_=1 for 0<_x_<1, and _f(x)_=0 outside of that range. So the _m_-th moment is the integral from 0 to 1 of _x_^_m_, which is 1/(_m_ + 1).

test_fms/random_numbers/test_random_numbers.F90
J-Lentz: One standard deviation was an empirical choice, based on the observation that the test passes within a couple seconds. But if we need the test to run faster, we could use a more lenient condition like 1.5 or 2 standard deviations.

I believe the central limit theorem implies that the distribution of sample moments should be approximately Gaussian, at least in the limit as the sample size approaches infinity.

random_numbers/mersennetwister.F90
```diff
@@ -92,6 +92,8 @@
 !> @{
 module MersenneTwister_mod
 ! -------------------------------------------------------------
+  use platform_mod, only: r8_kind
+
```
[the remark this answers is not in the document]
J-Lentz: It could... but the value gets assigned to a 4-byte integer, so I'm pretty confused what the original author had in mind here.

test_fms/random_numbers/test_random_numbers.F90
J-Lentz: What's the problem with using `FATAL`? I believe it should end up with four PEs any time it's run through the script, so this is really just a sanity check that shouldn't fail under normal circumstances.

random_numbers/include/random_numbers.inc
```diff
@@ -16,123 +16,37 @@
...
-    new%theNumbers = new_RandomNumberSequence(seed)
-
-  end function initializeRandomNumberStream_V
-  ! ---------------------------------------------------------
   !> Draws random scalar
-  subroutine getRandomNumber_Scalar(stream, number)
+  subroutine GET_RANDOM_NUMBER_0D_(stream, number)
     type(randomNumberStream), intent(inout) :: stream
-    real,                     intent(  out) :: number
+    real(RN_KIND_),           intent(  out) :: number
```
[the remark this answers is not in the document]
J-Lentz: Changed in fc57559edc068ca6b1407e6fbba70662c047ad20

random_numbers/mersennetwister.F90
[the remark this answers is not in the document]
J-Lentz: I've removed the `_8` in fc57559edc068ca6b1407e6fbba70662c047ad20... but I'm feeling a bit uneasy about it, since someone presumably put it there for a reason that's not clear to me.

test_fms/random_numbers/Makefile.am
```diff
@@ -0,0 +1,52 @@
...
+# Build this test program.
+check_PROGRAMS = \
+    test_random_numbers_r4 \
+    test_random_numbers_r8
+
+# This is the source code for the test.
+test_random_numbers_r4_SOURCES = test_random_numbers.F90
+test_random_numbers_r8_SOURCES = test_random_numbers.F90
+
+test_random_numbers_r4_CPPFLAGS = $(AM_CPPFLAGS) -DKIND_=r4_kind
```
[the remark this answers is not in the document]
J-Lentz: Changed in fc57559edc068ca6b1407e6fbba70662c047ad20

test_fms/random_numbers/test_random_numbers.F90
```diff
@@ -0,0 +1,279 @@
...
+#define CALC_MOMENT_(i) (1._k / real(i + 1, k))
+
+!> Calculate all expected moments and standard deviations of sample moments
+subroutine calc_expected_moments
+  integer :: i !> Moment order
+  integer, parameter :: n = n0_1d !> Sample size for which to calculate moment standard deviations
+
+  do i=1, n_moments
+    moment_mu(i) = CALC_MOMENT_(i)
+    moment_sigma(i) = sqrt((CALC_MOMENT_(2*i) - moment_mu(i)**2) / n)
```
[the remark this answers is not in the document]
J-Lentz: Changed in fc57559edc068ca6b1407e6fbba70662c047ad20

random_numbers/mersennetwister.F90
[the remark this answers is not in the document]
J-Lentz: It seems that the `_8` was a workaround to make it compile with GNU.

I've replaced this constant, as well as several others, with hexadecimal representations in 1d3f3218d60018ba6c91725d330ab15d476faeaf.

test_fms/random_numbers/test_random_numbers.F90
J-Lentz: I've tested it with Intel, GNU, and Cray and it works with all of them. It seems that assumed-size arrays are actually older than the dinosaurs (they're a Fortran 77 feature), while square bracket array constructors are an F2003 feature.

uramirez8707: can you confirm that this works with the nvdia compilers?

uramirez8707 (approved):

test_fms/random_numbers/test_random_numbers.F90
J-Lentz: Changed to a 1-indexed array in b978a99d766f774876429aff0d0d0143e070fcf6.

mcallic2 (approved):

test_fms/random_numbers/test_random_numbers.F90
mlee03: It's been a while since I thought about this stuff.  Why is it a uniform distribution?

J-Lentz: The comments in `random_number_mod` don't seem to explicitly state that the distribution is uniform. But, I think the term "random number generator" usually tends to mean "uniform random number generator", unless some other distribution is explicitly stated. The [0, 1] output range is also a hint that it's intended to output uniform random numbers.

random_numbers/include/random_numbers.inc
```diff
@@ -16,123 +16,37 @@
...
     ! Local variables
     integer :: i
 
     do i = 1, size(numbers)
-      numbers(i) = real(getRandomReal(stream%theNumbers))
+      call GET_RANDOM_NUMBER_0D_(stream, numbers(i))
     end do
-  end subroutine getRandomNumber_1D
-  ! ---------------------------------------------------------
+  end subroutine
```
ganganoaa, line 39: One can randomly distribute random numbers in a linear chain of pockets which may create more randomness for multidimensional arrays (2d, 3d, 4d, ... etc.) down the road .... just a thought !

ganganoaa (commented):
Looks fine to me.

mlee03 (approved):

PR #1200, opened 2023-04-25
Adds real_copy_set() to fms_diag_object_mod

ganganoaa:
**Description**
This update adds a routine _real_copy_set()_ to the _fms_diag_object_mod_ which copies input data to output data or sets the output data to a given real value if the input data is not present. The routine is implemented in the _fms_diag_accept_data()_ to initialize local copy of weight to the user supplied weight (_r4_kind_ or _r8_kind_) if it is present else set to a given value _val_.

Fixes # (issue)

**How Has This Been Tested?**
CI
**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -39,6 +39,7 @@ use fms_diag_output_buffer_mod
 use omp_lib
 #endif
 use mpp_domains_mod, only: domain1d, domain2d, domainUG, null_domain2d
+use platform_mod
 implicit none
 private
 
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -97,6 +97,7 @@ public :: fms_diag_object
 public :: fmsDiagObject_type
 integer, private :: registered_variables !< Number of registered variables
 public :: dump_diag_obj
+public :: real_copy_set
 
 contains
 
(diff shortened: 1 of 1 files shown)


diag_manager/fms_diag_object.F90
```diff
@@ -872,4 +880,27 @@ subroutine dump_diag_obj( filename )
...
+
+  IF ( PRESENT(in_data) ) THEN
+    SELECT TYPE (in_data)
+    TYPE IS (real(kind=r4_kind))
+      out_data = in_data
+    TYPE IS (real(kind=r8_kind))
+      out_data = real(in_data)
+    CLASS DEFAULT
+      CALL mpp_error('fms_diag_object_mod:real_copy_set',&
+        & 'The weight is not one of the supported types of real(kind=4) or real(kind=8)', FATAL)
```
thomas-robinson, line 900: There is not variable `weight` here, but the error message is referring to a weight.

diag_manager/fms_diag_object.F90
```diff
@@ -872,4 +880,27 @@ subroutine dump_diag_obj( filename )
   call mpp_error( FATAL, "You can not use the modern diag manager without compiling with -Duse_yaml")
 #endif
 end subroutine
+
+!> @brief Copies input data to output data with proper type if the input data is present
+!! else sets the output data to a given value val if it is present.
+!! If the value val and the input data are not present, the output data is untouched.
+subroutine real_copy_set(out_data, in_data, val)
```
thomas-robinson, line 887: This routine is more generic and should be in diag_utils I think.  Routines in this module should be type bound procedures.

thomas-robinson (changes requested):
I'm not sure why this is needed.  `fms_diag_accept_data` isn't going to actually use the weight, it will pass it along to another function where weight it used. That's where any `select type` stuff should be done

diag_manager/fms_diag_object.F90
ganganoaa: That was due to copy and paste. I will fix that.

diag_manager/fms_diag_object.F90
ganganoaa: I wanted to replace lines https://github.com/NOAA-GFDL/FMS/blob/bb25a07716dfcf3d097a5c88c31ac0d03a9904e0/diag_manager/diag_manager.F90#L1872-L1884 with this routine.

ganganoaa: I can move the routine to diag_utils_mod.

PR #1202, opened 2023-04-26
mixed mode intel 2022.2.1 allocation bug fix

rem1776:
**Description**
adds a second allocation checks for horiz_interp_Type's deletion routines. This is to fix a bug causing an error on the real types deallocation call, might be system related or just the compiler/mpi.

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes


mlee03:
@rem1776 I tested your fix with AM4 and Intel-classic/2022.2.1 on Gaea and still get the allocation error.  If it's not a compiler bug, I wonder if the allocation error arises because the code is trying to deallocate unallocated components of Interp%horizInterpReals8_type

rem1776:
> @rem1776 I tested your fix with AM4 and Intel-classic/2022.2.1 on Gaea and still get the allocation error. If it's not a compiler bug, I wonder if the allocation error arises because the code is trying to deallocate unallocated components of Interp%horizInterpReals8_type

We could potentially just deallocate the parent type and forget about the individual real fields... in my testing compilers we use do it cleanly no with leaks surprisingly. Don't know if that would fix it though. I'm just wondering if its the compiler or the system.

PR #1208, opened 2023-04-28
Replaces pointers which are allocated with allocatables

### LINKED ISSUE #1099: Allocated Pointers: Potential Memory Leak Issue

ganganoaa:
**Description**
Updates member pointers of _grid_box_type_, _grid_type_, _comm_type_, and _xmap_type_ to allocatables, which were previously allocated, to handle better memory management and to prevent memory leaks. Some local pointers in subroutine _load_xgrid()_ are also updated to allocatables for the same reason.

Fixes #1099 

**How Has This Been Tested?**

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/exchange/xgrid.F90
+++ b/exchange/xgrid.F90
@@ -838,9 +838,14 @@ logical,        intent(in)             :: use_higher_order
         endif
      enddo
 
-     if (allocated(i1_tmp, i2_tmp, j1_tmp, j2_tmp, area_tmp)) deallocate(i1_tmp, i2_tmp, j1_tmp, j2_tmp, area_tmp)
+     if (allocated(i1_tmp)) deallocate(i1_tmp)
+     if (allocated(i2_tmp)) deallocate(i2_tmp)
+     if (allocated(j1_tmp)) deallocate(j1_tmp)
+     if (allocated(j2_tmp)) deallocate(j2_tmp)
+     if (allocated(area_tmp)) deallocate(area_tmp)
      if(use_higher_order) then
-       if (allocated(di_tmp, dj_tmp)) deallocate( di_tmp, dj_tmp)
+       if (allocated(di_tmp)) deallocate(di_tmp)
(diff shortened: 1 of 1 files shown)


exchange/xgrid.F90
```diff
@@ -287,17 +287,17 @@ module xgrid_mod
 !> Type to hold pointers for grid boxes
 !> @ingroup xgrid_mod
 type grid_box_type
```
rem1776, line 289: This type is public and so are it's fields so we're gonna have to check the component's code to make sure these fields aren't used as pointers anywhere else besides internally (unless we want to make them private now but doubt that'll work). 

This applies to xmap_type as well.

exchange/xgrid.F90
```diff
@@ -2257,7 +2254,7 @@ end subroutine set_comm_get1_repro
 
 !#######################################################################
 subroutine set_comm_get1(xmap)
-  type (xmap_type), intent(inout) :: xmap
+  type (xmap_type), intent(inout), target :: xmap
   type (grid_type), pointer, save :: grid1 =>NULL()
```
rem1776, line 2258: guess we can't do much about these saved pointers right? might be less memory to save them as pointers anyway

ganganoaa: Pointers which are used as pointers should not be upgraded.

exchange/xgrid.F90
ganganoaa: Yes. You are right. If component codes are using them, this change certainly breaks something.

PR #1252, opened 2023-06-07
mixed precision time_interp

rem1776:
**Description**
Updates time_interp for mixed precision.

For time_interp, it just calculates fractional weights for different time units so most of the math is done in r8 (via operator overrides) for the time_type. The weight calculations are casted to the requested kind.

For time_interp_external2, it reads in files and interpolates with the calculated weight. Since the `init_external_field` routine does not have real arguments it always reads data in as r8. I tried to get around this by checking whats read in from fms2_io, but it'll only read in whatever type you give it for a buffer so will have to default to r8. The data is casted down for the calculations to match the kind of the weights. I also renamed the`field` module level list to `loaded_fields`, avoids naming conflicts and is a bit more clear.

The other change to note is in load_record (time_interp_external2). There is an optional argument to pass in a allocated horiz_interp_type which tells it to do a horizontal interpolation defined by horiz_interp_new. This causes a mismatch between the types in the two modules, so it casts down for the horiz_interp call if needed and copies the return values back over.

**How Has This Been Tested?**
latest intel on amd box.

**Checklist:**
- [x] `make distcheck` passes


time_interp/include/time_interp.inc
```diff
@@ -653,81 +335,35 @@ character(len=*), intent(out), optional :: err_msg
   if( T>=Timelist(ie) ) then
      ! time is after the end of the portion of the time list within the requested period
      index1 = ie;   index2 = is
-     weight = real( (T-Timelist(ie))//(Period-(Timelist(ie)-Timelist(is))) )
+     weight = real((T-Timelist(ie))//(Period-(Timelist(ie)-Timelist(is))), FMS_TI_KIND_ )
   else if (T<Timelist(is)) then
      ! time is before the beginning of the portion of the time list within the requested period
      index1 = ie;   index2 = is
-     weight = real( 1.0-((Timelist(is)-T)//(Period-(Timelist(ie)-Timelist(is)))) )
+     weight = real( 1.0_kindl -((Timelist(is)-T)//(Period-(Timelist(ie)-Timelist(is)))), FMS_TI_KIND_ )
```
mlee03, line 342: wondering how the math works here, real number - timetype

test_fms/time_interp/test_time_interp.F90
```diff
@@ -30,7 +30,7 @@ program test_time_interp
  type(time_type) :: Time_beg, Time_end, Time(num_Time)
  type(time_type), allocatable, dimension(:) :: Timelist
  integer :: index1, index2, mo, yr, timelist_len, outunit, ntest, nline
- real :: weight
+ real(TI_TEST_KIND_) :: weight
```
mlee03, line 33: It seems like the test is a guide on how to use time_interp.  Is there a way to check that the answers are as expected?

time_interp/time_interp_external2.F90
```diff
@@ -1269,147 +995,124 @@ subroutine realloc_fields(n)
...
-!</DESCRIPTION>
-!
-!<IN NAME="index" TYPE="integer">
-! returned from previous call to init_external_field.
-!</IN>
+end function get_external_field_size
 
-    function get_external_field_missing(index)
+!> return missing value
+function get_external_field_missing(index)
```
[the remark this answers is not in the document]
rem1776: it returns part of the ext_fieldtype that is always read in as an r8

time_interp/include/time_interp.inc
rem1776: The `//` is actually its own operator (as opposed to `/` ) for real division, so by the time it gets to the subtraction it'll be real - real.

mlee03: got it

time_interp/time_interp_external2.F90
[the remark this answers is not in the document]
mlee03: got it

time_interp/time_interp_external2.F90
```diff
@@ -142,22 +142,25 @@ module time_interp_external2_mod
...
+      module procedure time_interp_external_3d_r8
   end interface
 
   !> @addtogroup time_interp_external2_mod
   !> @{
 
   integer :: outunit
 
-  type(ext_fieldtype), save, private, pointer :: field(:) => NULL()
+  type(ext_fieldtype), save, private, pointer :: loaded_fields(:) => NULL()
```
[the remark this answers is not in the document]
mlee03: got it

test_fms/time_interp/test_time_interp.F90
rem1776: I added checks for all but the last, since that one cycles through instead of going to set dates.

rem1776:
I think the coupler build will break with our mixed mode branch until we bring in the main libfms.F90 changes, but this should be ready now. Had to include the horiz_interp fix i made a pr for, otherwise data override's tests would break.

mlee03:
@rem1776, could you compile with -Wconversion-extra with GCC?  There seems to be one or two conversion mismatch

mcallic2 (approved):

test_fms/time_interp/test_time_interp.F90
```diff
@@ -23,19 +23,24 @@ program test_time_interp
...
  type(time_type) :: Time_beg, Time_end, Time(num_Time)
  type(time_type), allocatable, dimension(:) :: Timelist
- integer :: index1, index2, mo, yr, timelist_len, outunit, ntest, nline
- real :: weight
+ integer :: index1, index2, mo, yr, outunit, ntest, nline
+ real(TI_TEST_KIND_) :: weight
+ real(TI_TEST_KIND_) :: ref_weights(num_Time), ref_weights_leap(num_Time)
+ real(TI_TEST_KIND_), parameter :: SMALL = 1.0e-7_kindl ! r4 will fail with 8
+ real(TI_TEST_KIND_), parameter :: midpoint = 0.483870967741935_kindl
+ real(TI_TEST_KIND_), parameter :: day_before_leap_day =  0.964285714285714_kindl
```
mlee03, line 39: where did these numbers come from?

rem1776: Just reference weights from the calculation, its a fraction of time passed over a interval. So for this its the midway point of a month (jan 16). I think i commented on it a bit lower down.

I might be able to replace these with the actual time_type calculation if thats preferred, i guess i just like seeing what the actual value ends up as.

mlee03: could the actual time_type calculation be used and add the numbers as comments?  Is this perhaps the reason you're not getting exactly agreeing answers?

rem1776: After looking into it i don't think i can reproduce the calculation unless i copy in a whole routine (set_modtime). I think this way has some benefits though, like if the time_type arithmentic overrides return the wrong thing a check with the same operations will just be checking wrong answers against wrong answers.

monin_obukhov/include/monin_obukhov.inc
```diff
@@ -123,6 +123,11 @@ do j=1, n3
   enddo
 enddo
 
+do j=1, n3
```
mlee03, line 126: ah it did that for you  too!  the do loops are repeated here

mlee03 (changes requested):

monin_obukhov/include/monin_obukhov.inc
rem1776: i think something wonky might be happening with our branches, might be better to simplify things and just use mixedmode as the starting point going forward

mlee03 (approved):

PR #1278, opened 2023-07-10
Add axis ids to the buffer object

uramirez8707:
**Description**
Adds axis_ids to the output buffers.

Before this update, the axis_ids were added to the field object but that is not enough because the axis_ids will be different depending on whether or not the file is subregional/has reduced k range.

The axis ids in the file object includes the (unique) axis ids of all of the variable in the file so it will be difficult to get the axis ids for the specific variable

There is more clean up that can be done, but this is enough for @ganganoaa to keep moving with his updates

Fixes # (issue)

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes


ganganoaa (commented):
Looks good.

thomas-robinson (approved):

PR #1311, opened 2023-07-31
Write buffer prep

uramirez8707:
**Description**
To prepare for the write_buffer update:
1. The buffer_ids were added to the diag_file object (this is needed to loop through them and write in diag_send_complete)
2. Adds the functionality to get the size of the subaxis (this is needed so that the subregional variables can be allocated to the correct size)
3. Adds the field id and yaml id to the output_buffer object (this is needed so that we can get information about the variable when writing for the variable)

Fixes # (issue)

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_file_object.F90
+++ b/diag_manager/fms_diag_file_object.F90
@@ -88,14 +88,16 @@ type :: fmsDiagFile_type
   integer, allocatable                         :: num_registered_fields !< The number of fields registered
                                                                         !! to the file
   integer, dimension(:), allocatable :: axis_ids !< Array of axis ids in the file
-  integer, dimension(:), allocatable :: buffer_ids !< array of buffer ids associated with the file
   integer :: number_of_axis !< Number of axis in the file
+  integer, dimension(:), allocatable :: buffer_ids !< array of buffer ids associated with the file
+  integer :: number_of_buffers !< Number of buffers that have been added to the file
   logical :: time_ops !< .True. if file contains variables that are time_min, time_max, time_average or time_sum
   integer :: unlim_dimension_level !< The unlimited dimension level currently being written
   logical :: is_static !< .True. if the frequency is -1
 
  contains
   procedure, public :: add_field_and_yaml_id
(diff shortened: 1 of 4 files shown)


diag_manager/fms_diag_file_object.F90
```diff
@@ -287,6 +292,16 @@ subroutine add_field_and_yaml_id (this, new_field_id, yaml_id)
   endif
 end subroutine add_field_and_yaml_id
 
+!> \brief Adds a field and yaml ID to the file
+subroutine add_buffer_ids (this, buffer_id)
```
thomas-robinson, line 296: The name of this routine is a bit misleading.  It only adds one ID, but `ids` is plural in the routine name.

thomas-robinson:
"Adds the field id and yaml id to the output_buffer object (this is needed so that we can get information about the variable when writing for the variable)"
Writing is done on a file level, and the file object should have all of this information.  So why does the buffer object need this?

thomas-robinson:
"Adds the field id and yaml id to the output_buffer object (this is needed so that we can get information about the variable when writing for the variable)"
Writing is done on a file level, and the file object should have all of this information.  So why does the buffer object need this?

uramirez8707:
> "Adds the field id and yaml id to the output_buffer object (this is needed so that we can get information about the variable when writing for the variable)" Writing is done on a file level, and the file object should have all of this information. So why does the buffer object need this?

We are looping through the buffer_ids for the file. The field_ids in the diag file object do not necessary correspond to the buffer_ids (i.e buffer_ids(i) may not be the buffer for field_ids(i)). They were added to the file as the fields were registering/data was sent.

diag_manager/fms_diag_file_object.F90
uramirez8707: https://github.com/NOAA-GFDL/FMS/pull/1311/commits/21f74ed50859c66e613a01d31c00c1169fcd03c8

PR #1312, opened 2023-07-31
Gcc fix: replacing cray pointers with fortran pointers in mpp_scatter and gather

### LINKED ISSUE #1276: unit test failures with gcc + O2 optimization
**Describe the bug**
Some unit test that are passing gnu normally are failing when compiling with prod flags:

test_fms2_io(all)
test_io_with_mask
test_mpp_nesting(all)
horiz_interp(bicubic new)

**To Reproduce**
add  `-O2 -fno-expensive-optimizations` to your compiler flags for gcc

**Expected behavior**
not fail

**System Environment**
amd box
- Compiler(s): gcc 10-12
- MPI type: mpich
- netCDF Version: 4.9.2 (c) + 4.61 (fortran)

**Additional context**

Some (mainly fms2_io) may be related to gcc's handling of cray pointers. The errors are coming from cray pointer assignments using the non-standard `LOC` intrinsic for the gather and scatter routines.

mpp_nesting seems to fail when trying to deallocate a pointer from a derived type (nest_domain_type%nest_level).

The horiz_interp failure is from the mixed precision update, I've seen similar errors trying to access its sub-type's field.

laurenchilutti:
**Description**
This PR fixes 
test_bc_restarts
test_fms2_io(all)
test_io_with_mask

All of these were related to the use of cray pointers in mpp_scatter and mpp_gather.  
https://github.com/NOAA-GFDL/FMS/blob/0efc2e5c57d080d16fb3e076e9edc21a860992fa/mpp/include/mpp_gather.fh#L112-L132
A seg fault occurs at line 126 when we point ot the location of `data`.  `data`will only be allocated on the root pe and all other PEs seg fault.  To fix this, I have replaced cray pointers with fortran pointers and have used pointer rank remapping to point a 3D pointer to a 2D input field.

Also includes a change to compressed_write.F90 where we need to surround calls to `netcdf_write_data` with `if (fileobj%is_root)`.  When running the unit tests with intel's ifort and ifx compilers, multiple test_fms_io tests fail with an error related to buffer variables not being allocated.  We determined that this was because of the FMS 2023.01.01 updates, in which we only allocate these buffer variables on root pe, use mpp_gather to collect a value for this buffer, and then netcdf_write_data to write this data using the buffer.  

Other changes include increasing size of filename to allow for longer filename and fixing some errors in test_compressed_writes.F90 and test_domain_io.F90's `register_field_wrapper` functions.

Fixes #1276 

**How Has This Been Tested?**
Has been tested with GCC 13 with -O2 flags
make check

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_gather.fh
+++ b/mpp/include/mpp_gather.fh
@@ -111,19 +111,22 @@ end subroutine MPP_GATHER_1DV_
 
 subroutine MPP_GATHER_PELIST_2D_(is, ie, js, je, pelist, array_seg, data, is_root_pe, &
                                  ishift, jshift)
-   integer,                           intent(in)    :: is, ie, js, je
-   integer,   dimension(:),           intent(in)    :: pelist
-   MPP_TYPE_, dimension(is:ie,js:je), intent(in)    :: array_seg
-   MPP_TYPE_, dimension(:,:),         intent(inout) :: data
-   logical,                           intent(in)    :: is_root_pe
-   integer,   optional,               intent(in)    :: ishift, jshift
+   integer,                                       intent(in)    :: is, ie, js, je
+   integer,   dimension(:),                       intent(in)    :: pelist
+   MPP_TYPE_, dimension(is:ie,js:je),     target, intent(in)    :: array_seg
+   MPP_TYPE_, dimension(:,:), contiguous, target, intent(inout) :: data
+   logical,                                       intent(in)    :: is_root_pe
+   integer,   optional,                           intent(in)    :: ishift, jshift
 
-   MPP_TYPE_ ::  arr3D(size(array_seg,1),size(array_seg,2),1)
-   MPP_TYPE_ :: data3D(size(     data,1),size(     data,2),1)
-   pointer( aptr,  arr3D )
-   pointer( dptr, data3D )
-   aptr = LOC(array_seg)
-   dptr = LOC(     data)
+   MPP_TYPE_, pointer ::  arr3D(:,:,:)
+   MPP_TYPE_, pointer :: data3D(:,:,:)
+
+   arr3D(1:size(array_seg,1),1:size(array_seg,2),1:1) => array_seg
+   if (is_root_pe) then
+     data3D(1:size(data,1),1:size(data,2),1:1) => data
+   else
+     data3D => null()
+   endif
 
    call mpp_gather(is, ie, js, je, 1, pelist, arr3D, data3D, is_root_pe, &
                    ishift, jshift)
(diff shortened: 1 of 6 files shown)


bensonr:
@laurenchilutti - I haven't looked at the test code source, but it would be good to know if there are standalone gather/scatter tests that call the the routines directly with a data-domain sized array passing only the compute domain portion as an argument.

laurenchilutti:
I have just merged in the main branch and made an additional commit to fix some intel failures we see when using `-check all` flags.  I will update the description of this PR to elaborate.

laurenchilutti:
> @laurenchilutti - I haven't looked at the test code source, but it would be good to know if there are standalone gather/scatter tests that call the the routines directly with a data-domain sized array passing only the compute domain portion as an argument.

We test mpp_scatter and mpp_gather in https://github.com/NOAA-GFDL/FMS/blob/main/test_fms/mpp/test_mpp_gatscat.F90 but you will notice that the fms2io uses of mpp_gather and mpp_scatter are different than in the test.  For one: In fms2io we are only allocating the data arrays on the root pe before calling mpp_gather, but in these tests I see that they allocate the data arrays on all pes, so we would never have caught the cray pointer seg faults.

bensonr (approved):

mpp/include/mpp_gather.fh
```diff
@@ -111,19 +111,22 @@ end subroutine MPP_GATHER_1DV_
...
-   pointer( aptr,  arr3D )
-   pointer( dptr, data3D )
-   aptr = LOC(array_seg)
-   dptr = LOC(     data)
+   integer,                                       intent(in)    :: is, ie, js, je
+   integer,   dimension(:),                       intent(in)    :: pelist
+   MPP_TYPE_, dimension(is:ie,js:je),     target, intent(in)    :: array_seg
+   MPP_TYPE_, dimension(:,:), contiguous, target, intent(inout) :: data
+   logical,                                       intent(in)    :: is_root_pe
+   integer,   optional,                           intent(in)    :: ishift, jshift
```
thomas-robinson, line 119: Why are these lines changed?

laurenchilutti: This is whitepace changes to keep the alignment/formatting.  If you select "hide whitespace" under the Files Changed Tab -> Gear Icon these changes are ignored.

thomas-robinson: If I type `git blame`, the changes won't be ignored...

I'm not saying this NEEDS to be changed, but you will own these lines of code, and it will be unclear why it's lumped together with cray pointer removal.

laurenchilutti: I'm aware, and if I am ever asked about these lines of code, is there a problem with me stating that I changed whitespace?  I don't see a benefit to redoing this just so I don't "own" the code.

rem1776 (approved):

PR #1315, opened 2023-08-01
dmupdate: change type of buffer counter

rem1776:
**Description**
changes the allocation so the counter variable is allocated to a real and adds a remap_counter routine that mirrors remap_buffer.

Alternatively we could just make the counter a real(8) type, right now i just have the counter matching the kind of the buffer regardless of whether its an int or not.

**How Has This Been Tested?**
amd box with oneapi 2023.1

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_output_buffer.F90
```diff
@@ -263,6 +263,63 @@ function remap_buffer(buffobj, field_name, has_diurnal_axis)
 
 end function remap_buffer
 
+!> Remaps the counter
+!> @returns a 5D remappeding of the counter array, with 1:1 for any added dimensions.
```
thomas-robinson, line 267: Typo in remappeding

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -263,6 +263,63 @@ function remap_buffer(buffobj, field_name, has_diurnal_axis)
...
+function remap_counter(buffobj, field_name, has_diurnal_axis)
+  class(fmsDiagOutputBuffer_class), target, intent(inout) :: buffobj !< any dimension buffer object
+  class(*), pointer                                 :: remap_counter(:,:,:,:,:)
+  character(len=*), intent(in)                      :: field_name !< name of field for error output
+  logical, intent(in) :: has_diurnal_axis !< true if the buffer has diurnal axis
+
+  ! get num dimensions from type extension
+  select type (buffobj)
+    type is (outputBuffer0d_type)
+      if (.not. allocated(buffobj%counter)) call mpp_error(FATAL, "remap_counter: counter data not yet allocated" // &
```
thomas-robinson, line 277: can you include the file name in these error messages also?

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -263,6 +263,63 @@ function remap_buffer(buffobj, field_name, has_diurnal_axis)
 
 end function remap_buffer
 
+!> Remaps the counter
+!> @returns a 5D remappeding of the counter array, with 1:1 for any added dimensions.
+function remap_counter(buffobj, field_name, has_diurnal_axis)
+  class(fmsDiagOutputBuffer_class), target, intent(inout) :: buffobj !< any dimension buffer object
+  class(*), pointer                                 :: remap_counter(:,:,:,:,:)
+  character(len=*), intent(in)                      :: field_name !< name of field for error output
```
thomas-robinson, line 271: does the buffer object have this information?

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -263,6 +263,63 @@ function remap_buffer(buffobj, field_name, has_diurnal_axis)
...
+        remap_counter(1:size(buffobj%counter,1), 1:size(buffobj%counter,2), 1:size(buffobj%counter,3), &
+                  1:size(buffobj%counter,4), 1:1) => buffobj%counter(:,:,:,:)
+      end if
+    type is (outputBuffer5d_type)
+      if (.not. allocated(buffobj%counter)) call mpp_error(FATAL, "remap_counter: counter data not yet allocated" // &
+                                                                 "for field:" // field_name)
+      remap_counter(1:size(buffobj%counter,1), 1:size(buffobj%counter,2), 1:size(buffobj%counter,3), &
+                   1:size(buffobj%counter,4), 1:size(buffobj%counter,5)) => buffobj%counter(:,:,:,:,:)
+    class default
+      call mpp_error( FATAL, 'remap_buffer_pointer: invalid buffer type for remapping')
```
thomas-robinson, line 318: This should also have the field and file name

thomas-robinson (changes requested):
Fix the typo at a minimum

thomas-robinson (changes requested):

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -494,17 +551,17 @@ subroutine allocate_buffer_3d(this, buff_type, buff_sizes, field_name, diurnal_s
   select type (buff_type)
     type is (integer(kind=i4_kind))
       allocate(integer(kind=i4_kind) :: this%buffer( buff_sizes(1),buff_sizes(2), buff_sizes(3)))
-      allocate(integer(kind=i4_kind) :: this%counter(buff_sizes(1),buff_sizes(2), buff_sizes(3)))
-      allocate(integer(kind=i4_kind) :: this%count_0d(n_samples))
-      this%counter = 0_i4_kind
-      this%count_0d = 0_i4_kind
+      allocate(real(kind=r4_kind) :: this%counter(buff_sizes(1),buff_sizes(2), buff_sizes(3)))
```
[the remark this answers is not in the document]
thomas-robinson: Right.  That should be documented in the code where they are allocated.

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -43,12 +43,14 @@ module fms_diag_output_buffer_mod
   integer, allocatable, public :: num_elements(:) !< used in time-averaging
   class(*), allocatable, public :: count_0d(:) !< used in time-averaging along with
                                        !! counter which is stored in the child types (bufferNd)
+                                       !! Always allocated to a real type since its used for weights
```
thomas-robinson, line 46: I think this should say "Always allocated to a real type because it is used in a calculation with the variable weights".  I think that's more descriptive

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -213,33 +221,34 @@ end subroutine set_buffer_id
 
 !> Remaps 0-5d data buffer from the given object onto a 5d array pointer.
 !> @returns a 5D remapped buffer, with 1:1 for any added dimensions.
-function remap_buffer(buffobj, field_name, has_diurnal_axis)
+function remap_buffer(buffobj, field_name, file_name, has_diurnal_axis)
   class(fmsDiagOutputBuffer_class), target, intent(inout) :: buffobj !< any dimension buffer object
   class(*), pointer                                 :: remap_buffer(:,:,:,:,:)
   character(len=*), intent(in)                      :: field_name !< name of field for error output
+  character(len=*), intent(in)                      :: file_name !< name of field for error output
```
thomas-robinson, line 228: The comments on all of these in inaccurate.  It says "field" but it should say "file"

diag_manager/fms_diag_output_buffer.F90
```diff
@@ -213,33 +221,34 @@ end subroutine set_buffer_id
...
   logical, intent(in) :: has_diurnal_axis !< true if the buffer has diurnal axis
 
   ! get num dimensions from type extension
   select type (buffobj)
     type is (outputBuffer0d_type)
       if (.not. allocated(buffobj%buffer)) call mpp_error(FATAL, "remap_buffer: buffer data not yet allocated" // &
-                                                                 "for field:" // field_name)
+                                                                 "for field:" // field_name // "in file:" // file_name)
```
thomas-robinson, line 235: I think `field_name` and `file_name` need to be `trim(field_name)` and `trim(file_name)`

thomas-robinson (changes requested):

diag_manager/fms_diag_output_buffer.F90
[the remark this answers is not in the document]
uramirez8707: I don't understand the math here:

When averaging, the buffer is initialized to 0 and each time period the data is multiplied by the weight (which will be 1 if it not passed in the send_data call) and added to the buffer

https://github.com/NOAA-GFDL/FMS/blob/dd53551c20345bd902786ec84a0e37a1820b0df3/diag_manager/diag_manager.F90#L2046-L2048

The counter is initialized to 0 and the weight gets added after every time
https://github.com/NOAA-GFDL/FMS/blob/dd53551c20345bd902786ec84a0e37a1820b0df3/diag_manager/diag_manager.F90#L2050-L2051

When it is time to write, it divides the buffer by the counter and that is the "average" that gets written:
https://github.com/NOAA-GFDL/FMS/blob/dd53551c20345bd902786ec84a0e37a1820b0df3/diag_manager/diag_manager.F90#L3466-L3467

I don't understand why the weight gets added to the counter, shouldn't it just add 1 every time? 

I **think** that the reason why it is an array and not a single value is because when sending in blocks of data only a portion of the buffer gets updated so the counter needs to be tracked in a per grid point basis.

uramirez8707:
I think it might be better to store the counter as an `r8_kind` and convert it to an `r4_kind` when diving by it. 
Currently, we are storing the weight as an r8_kind
https://github.com/NOAA-GFDL/FMS/blob/07ff0679281f4d99423483f33eb7fc4240f5dc88/diag_manager/fms_diag_object.F90#L534-L535

This will help us avoid two more select types!

The weight and the counter as only used when averaging which will not be supported for integers ...

diag_manager/fms_diag_output_buffer.F90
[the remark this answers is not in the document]
rem1776: I think the weight doesn't have to be 1 necessarily, 1.0 is the default but otherwise any real value can get passed in to send data:

https://github.com/NOAA-GFDL/FMS/blob/07ff0679281f4d99423483f33eb7fc4240f5dc88/diag_manager/diag_manager.F90#L1882-L1895

rem1776:
> I think it might be better to store the counter as an `r8_kind` and convert it to an `r4_kind` when diving by it. Currently, we are storing the weight as an r8_kind
> 
> https://github.com/NOAA-GFDL/FMS/blob/07ff0679281f4d99423483f33eb7fc4240f5dc88/diag_manager/fms_diag_object.F90#L534-L535
> 
> This will help us avoid two more select types!
> 
> The weight and the counter as only used when averaging which will not be supported for integers ...

It seems like the weight will be able to be passed in as both r4/r8 kinds. If we were taking the weight argument as r8 initially i think that would make sense, but if we always store as r8 we will have to be doing two conversions if its r4 which probably isn't good.

diag_manager/fms_diag_output_buffer.F90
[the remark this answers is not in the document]
thomas-robinson: @uramirez8707 @rem1776 is this conversation resolved?

uramirez8707 (approved):

thomas-robinson (approved):

PR #1339, opened 2023-08-16
Add target flags for GNU CI

rem1776:
**Description**
adds the gnu mkmf template flags for prod and debug to the ci matrix

**How Has This Been Tested?**
CI, should be working properly. draft until all the tests are passing, needs one or two other fixes.

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/.github/workflows/github_autotools_gnu.yml
+++ b/.github/workflows/github_autotools_gnu.yml
@@ -20,6 +20,7 @@ jobs:
       env:
         TEST_VERBOSE: 1
         DISTCHECK_CONFIGURE_FLAGS: "${{ matrix.conf-flag }} ${{ matrix.input-flag }} ${{ matrix.io-flag }}"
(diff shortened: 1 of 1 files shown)


rem1776:
> Are we using these flags with intel? PROD, REPRO, DEBUG?

No this is just for GNU: https://github.com/NOAA-GFDL/mkmf/blob/61495ce9d197a2a54bbcbb5d879fefabb3cf3c01/templates/ncrc5-gcc.mk#L87-L107
Figured I'd make a separate pr for intel.

I just added one for repro, its blank except for the O2 flag which prod already has so thats why i didn't include it initially. The only difference is prod has the `-fno-expensive-optimizations` flag and repro doesn't and I'm not really sure how much that ends up affecting things, but might as well test both.

rem1776:
will combine this and #1346 with some more CI updates

PR #1352, opened 2023-08-23
merge mixed mode base into mixedmode

rem1776:
**Description**
merges mixedmode_base into mixedmode, brings in the latest main updates .

**Checklist:**
- [x] `make distcheck` passes


axis_utils/include/axis_utils2.inc
```diff
@@ -214,7 +212,7 @@
        endif
 
        lon_strt = lon(1)
-       do i=2,len+1
```
mlee03, line 217: Is something funky going on here?  The mixedmode branch already contains this change and shouldn't show up here..? @J-Lentz, is this correct?

rem1776: I feel like its from updating mixemode_base but not doing a merge commit with main so it views the commits as separate even though its the same changes. If you don't have a merge commit it just thinks its your own changes, when they should match up with the commits on the branch you are merging into. I'll try merging main into mixedmode_base and see if I can get this to go in cleanly.

After this merge I think we should get rid of mixedmode_base. The mixedmode braches should have the same includes and everything as mixedmode_base from merging our prs in, so we shouldn't need the other branch just for main updates. I think it would be cleaner and make more sense to merge main right into mixedmode and start our prs there, I can try it next time we need an update.

mcallic2 (approved):

mlee03 (approved):

PR #1353, opened 2023-08-23
coupler mixed precision

rem1776:
**Description**
Adds mixed precision support for the coupler subdirectory.

Majority of the changes are in coupler_types since it uses a few encapsulated types for 1-3 dimensional fields. The atmos_ocean_fluxes module has changes when needed to interact with the right type but does not take any required real arguments. Ensemble manager does not use reals (or the rest of this subdirectory) at all, so no changes have been made.

I noticed that in the FMScoupler only the `coupler_Nd_bc_type` is used so the r4/r8 split happens within it via the `bc` and `bc_r4` members. This should be less disruptive for other codes using this type since theres nothing being removed for the most part, just the r4 stuff being added and used if needed.

Because the r4/r8 divide happens within the type, most routines that don't take reals just check which type has been associated (bc or bc_r4) and then use whats appropriate.

Fixes # (issue)

**How Has This Been Tested?**

**Checklist:**
- [x] `make distcheck` passes


coupler/atmos_ocean_fluxes.F90
```diff
@@ -118,9 +117,9 @@ function aof_set_coupler_flux(name, flux_type, implementation, atm_tr_index, par
     character(len=*), intent(in)                :: flux_type !< flux_type
     character(len=*), intent(in)                :: implementation !< implementation
     integer, intent(in), optional               :: atm_tr_index !< atm_tr_index
-    real, intent(in), dimension(:), optional    :: param !< param
+    real(r8_kind), intent(in), dimension(:), optional    :: param !< param
     logical, intent(in), dimension(:), optional :: flag !< flag
-    real, intent(in), optional                  :: mol_wt !< mol_wt
+    real(r8_kind), intent(in), optional                  :: mol_wt !< mol_wt
```
mlee03, line 122: Do you think it'd be better if mol_wt was a class(*) here?

coupler/atmos_ocean_fluxes.F90
```diff
@@ -535,172 +546,345 @@ subroutine atmos_ocean_fluxes_init(gas_fluxes, gas_fields_atm, gas_fields_ice, v
...
+          gas_fields_ice%bc_r4(n)%field(m)%long_name =&
+              & fm_util_get_string(trim(fm_util_get_string(trim(flux_list) // 'ice/name', index = m)) // '-long_name')
+          gas_fields_ice%bc_r4(n)%field(m)%long_name = trim(gas_fields_ice%bc_r4(n)%field(m)%long_name) // ' for ' // name
+        enddo
+
+        ! Save the atm_tr_index.
+        gas_fluxes%bc_r4(n)%atm_tr_index = fm_util_get_integer('atm_tr_index', scalar = .true.)
+
+        ! Save the molecular weight.
+        gas_fluxes%bc_r4(n)%mol_wt = fm_util_get_real('mol_wt', scalar = .true.)
```
mlee03, line 814: should there be a real,r4_kind) around fm_util_get_real?  I think fm_util_get_real returns an r8_kind

coupler/atmos_ocean_fluxes.F90
```diff
@@ -510,10 +513,18 @@ subroutine atmos_ocean_fluxes_init(gas_fluxes, gas_fields_atm, gas_fields_ice, v
           write (outunit,*) trim(note_header), ' Processing ', gas_fluxes%num_bcs, ' gas fluxes'
     endif
 
-    ! allocate the arrays
-    allocate (gas_fluxes%bc(gas_fluxes%num_bcs))
-    allocate (gas_fields_atm%bc(gas_fields_atm%num_bcs))
-    allocate (gas_fields_ice%bc(gas_fields_ice%num_bcs))
+    if(present(use_r4_kind)) use_r4_kind_loc = use_r4_kind
```
mlee03, line 516: wondering if this should go on top before the while loop

coupler/atmos_ocean_fluxes.F90
```diff
@@ -424,7 +424,7 @@ end function aof_set_coupler_flux
   !! @throw FATAL, "Num_flags is negative for [name]: [num_flags]"
   !! @throw FATAL, "Problem dumping fluxes tracer tree"
   !! @throw FATAL, "Number of fluxes does not match across the processors: [gas_fluxes%num_bcs] fluxes"
-  subroutine atmos_ocean_fluxes_init(gas_fluxes, gas_fields_atm, gas_fields_ice, verbosity)
+  subroutine atmos_ocean_fluxes_init(gas_fluxes, gas_fields_atm, gas_fields_ice, verbosity, use_r4_kind)
```
mlee03, line 427: do you think there will be a case where the user wants these three fluxes to be set at different precisions?

coupler/atmos_ocean_fluxes.F90
rem1776: Here's where it ends up being used in this routine:

https://github.com/NOAA-GFDL/FMS/blob/5d8e7ffeb25622b4be39ce22c7d5756a1308c7b2/coupler/atmos_ocean_fluxes.F90#L298-L302

But then also again here:

https://github.com/NOAA-GFDL/FMS/blob/5d8e7ffeb25622b4be39ce22c7d5756a1308c7b2/coupler/atmos_ocean_fluxes.F90#L645

I think fm_util_get_real is r8_kind only right cause of the read in strings?

coupler/coupler_types.F90
```diff
@@ -412,6 +562,9 @@ subroutine coupler_type_copy_1d_2d(var_in, var_out, is, ie, js, je,&
       call mpp_error(FATAL, trim(error_header) // ' Number of output fields exceeds zero')
     endif
 
+    if(associated(var_in%bc) .eq. associated(var_in%bc_r4)) &
```
mlee03, line 565: could there be a chance both are unassociated?

coupler/atmos_ocean_fluxes.F90
mlee03: that and because there was no way to tell what precision the user wanted based on the input arguments.  I'm wondering if the users will pass in r4_kind mol_wt here because they're assuming that the r4 version of bc will be populated

coupler/coupler_types.F90
rem1776: yeah definitely, thats why its comparing the logical return values. If they are both associated or both not associated it will error out, this way it'll ensure that the r4/r8 field types used are mutually exclusive.

coupler/atmos_ocean_fluxes.F90
rem1776: I just left the mol_wt and param fields as r8_kind regardless, this way they are always the value returned by the field manager.

coupler/coupler_types.F90
```diff
@@ -67,13 +67,62 @@ module coupler_types_mod
```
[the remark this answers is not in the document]
rem1776: i replaced int64 with i8_kind, int32 wasn't used

coupler/atmos_ocean_fluxes.F90
rem1776: Yeah to be safe I'll switch it over to a class(*), since it's intent(in) should be fine.

coupler/atmos_ocean_fluxes.F90
rem1776: I do not know honestly. It could be added using similar logic later on if we need it.

coupler/atmos_ocean_fluxes.F90
rem1776: Not sure what you mean, I have this if for allocation here and then the while loop below has another if within it for the parts that use reals.

coupler/coupler_types.F90
rem1776: I put in a fix, the check for allocation needed to be after the checks for the optional `as_needed` argument. I also made updates so that it'll only check for one kind allocated if bc%set is true since it indicates initialization.

coupler/coupler_types.F90
```diff
@@ -3112,48 +2783,101 @@ subroutine CT_set_diags_3d(var, diag_name, axes, time)
           & '(coupler_types_set_diags_3d): axes has less than 3 elements')
     endif
 
-    do n = 1, var%num_bcs
-      do m = 1, var%bc(n)%num_fields
-        var%bc(n)%field(m)%id_diag = register_diag_field(diag_name,&
-            & var%bc(n)%field(m)%name, axes(1:3), Time,&
-            & var%bc(n)%field(m)%long_name, var%bc(n)%field(m)%units )
+    if(var%set) then
+      if(associated(var%bc) .eqv. associated(var%bc_r4)) &
```
mlee03, line 2787: i think you missed one lol

rem1776: should be all done now

coupler/coupler_types.F90
```diff
@@ -625,6 +783,12 @@ subroutine CT_spawn_1d_2d(var_in, var, idim, jdim, suffix, as_needed)
     if (.not.var_in%set)&
         & call mpp_error(FATAL, trim(error_header) // ' The parent type has not been initialized.')
 
+    ! check only one kind is used
+    if(var_in%set .and. var_in%num_bcs .gt. 0) then
```
mlee03, line 787: is the var_in%set needed here since it's checked above?  (not that it significantly matters)

mlee03:
@rem1776 could you add @mcallic2  and @J-Lentz as reviewers?

coupler/coupler_types.F90
```diff
@@ -1252,29 +1727,63 @@ subroutine CT_copy_data_2d(var_in, var, halo_size, bc_index, field_index,&
...
-      do m = 1, var%bc(n)%num_fields
-        if (present(field_index)) then
-          if (m /= field_index) cycle
-        endif
-        if ( associated(var%bc(n)%field(m)%values) ) then
-          do j=var%jsc-halo,var%jec+halo
-            do i=var%isc-halo,var%iec+halo
-              var%bc(n)%field(m)%values(i,j) = var_in%bc(n)%field(m)%values(i+i_off,j+j_off)
+    ! num_bcs .lt. 1 -> loop doesn't run but shouldn't error out
+    if (associated(var_in%bc) .or. var_in%num_bcs .lt. 1) then
```
mlee03, line 1736: Would it be clearer if var_in%num_bcs .lt. n1?

mlee03: Also wondering if var_in%num_bcs.lt.1 should be separated into its own if else block for clarity

J-Lentz (approved):

axis_utils/include/axis_utils2.inc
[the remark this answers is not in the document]
rem1776: should be updated, i left that one line deletion in axis utils cause that extra line will mess with the documentation parser.

coupler/coupler_types.F90
rem1776: i think this is a little more clear. Some code within these is outside of loops so needs to run if either part of the conditional is true, otherwise will have to be repeated. could change the logic to use and .and. instead, but that would be more confusing

coupler/coupler_types.F90
rem1776: nope (at least for the spawn routines), took these out with 47ae5f7

mlee03:
YAY! :partying_face:

mlee03 (approved):

PR #1365, opened 2023-09-13
Mixed precision `amip_interp_mod`

J-Lentz:
**Description**
Mixed precision `amip_interp_mod`

**How Has This Been Tested?**
Builds with Intel on the AMD box.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,7 +380,8 @@ foreach(kind ${kinds})
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/interpolator/include>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/coupler/include>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/mosaic2/include>
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/data_override/include>)
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/data_override/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/amip_interp/include>)
 
   target_include_directories(${libTgt} INTERFACE
     $<BUILD_INTERFACE:${moduleDir}>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,7 +323,8 @@ foreach(kind ${kinds})
                                                  tracer_manager/include
                                                  interpolator/include
                                                  coupler/include
-                                                 data_override/include)
+                                                 data_override/include
+                                                 amip_interp/include)
 
   target_compile_definitions(${libTgt}_f PRIVATE "${fms_defs}")
   target_compile_definitions(${libTgt}_f PRIVATE "${${kind}_defs}")
--- a/amip_interp/amip_interp.F90
+++ b/amip_interp/amip_interp.F90
@@ -155,8 +155,7 @@ integer :: i_sst = 1200
 integer :: j_sst = 600
 real(r8_kind), parameter:: big_number = 1.E30_r8_kind
 logical :: forecast_mode = .false.
-real(r8_kind), allocatable, dimension(:,:) ::  sst_ncep,    sst_anom    !< r8 versions of sst_ncep and sst_anom
-real(r4_kind), allocatable, dimension(:,:) ::  sst_ncep_r4, sst_anom_r4 !< r4 versions of sst_ncep and sst_anom
+real(r8_kind), allocatable, dimension(:,:) ::  sst_ncep, sst_anom
 
 public i_sst, j_sst, sst_ncep, sst_anom, forecast_mode, use_ncep_sst
 
--- a/amip_interp/amip_interp.F90
+++ b/amip_interp/amip_interp.F90
@@ -294,7 +294,8 @@ end interface
 type amip_interp_type
    private
    type (horiz_interp_type)              :: Hintrp, Hintrp2 ! add by JHC
-   class(*), dimension(:,:), allocatable :: data1, data2
+   real(r4_kind), dimension(:,:), allocatable :: data1_r4, data2_r4
+   real(r8_kind), dimension(:,:), allocatable :: data1_r8, data2_r8
    type (date_type)                      :: Date1, Date2
    logical                               :: use_climo, use_annual
    logical                               :: I_am_initialized=.false.
@@ -381,8 +382,11 @@ contains
    subroutine amip_interp_del (Interp)
      type (amip_interp_type), intent(inout) :: Interp
 
-     if(allocated(Interp%data1)) deallocate(Interp%data1)
-     if(allocated(Interp%data2)) deallocate(Interp%data2)
+     if(allocated(Interp%data1_r4)) deallocate(Interp%data1_r4)
+     if(allocated(Interp%data1_r8)) deallocate(Interp%data1_r8)
+     if(allocated(Interp%data2_r4)) deallocate(Interp%data2_r4)
+     if(allocated(Interp%data2_r8)) deallocate(Interp%data2_r8)
+
      if(allocated(lon_bnd_r4))   deallocate(lon_bnd_r4)
      if(allocated(lon_bnd_r8))   deallocate(lon_bnd_r8)
      if(allocated(lat_bnd_r4))   deallocate(lat_bnd_r4)
@@ -466,8 +470,10 @@ subroutine amip_interp_type_eq (amip_interp_out, amip_interp_in)
 
     amip_interp_out%Hintrp     =  amip_interp_in%Hintrp
     amip_interp_out%Hintrp2    =  amip_interp_in%Hintrp2 !< missing assignment statement; added by GPP
-    amip_interp_out%data1      =  amip_interp_in%data1
-    amip_interp_out%data2      =  amip_interp_in%data2
+    amip_interp_out%data1_r4   =  amip_interp_in%data1_r4
+    amip_interp_out%data1_r8   =  amip_interp_in%data1_r8
+    amip_interp_out%data2_r4   =  amip_interp_in%data2_r4
+    amip_interp_out%data2_r8   =  amip_interp_in%data2_r8
     amip_interp_out%Date1      =  amip_interp_in%Date1
     amip_interp_out%Date2      =  amip_interp_in%Date2
     amip_interp_out%Date1      =  amip_interp_in%Date1
--- a/amip_interp/amip_interp.F90
+++ b/amip_interp/amip_interp.F90
@@ -147,7 +147,8 @@ private
 !----------------- Public interfaces -----------------------------------
 
 public amip_interp_init, amip_interp_init_r4, amip_interp_init_r8, get_amip_sst, &
-     & get_amip_ice, amip_interp_new, amip_interp_del, amip_interp_type, assignment(=)
+     & get_amip_ice, amip_interp_new, amip_interp_del, amip_interp_type_r4, &
+     & amip_interp_type_r8, assignment(=)
 
 !-----------------------------------------------------------------------
 !----------------- Public Data -----------------------------------
@@ -296,17 +297,28 @@ end interface amip_interp_del
 !----- public data type ------
 
 !> @brief Contains information needed by the interpolation module (exchange_mod) and buffers
-!! data. Real precision is determined by the rkind parameter, which defaults to r8_kind.
+!! data (r4_kind flavor).
 !> @ingroup amip_interp_mod
-type amip_interp_type(rkind)
-   integer, kind            :: rkind = r8_kind ! Real kind for data1 and data2
+type amip_interp_type_r4
    private
    type (horiz_interp_type) :: Hintrp, Hintrp2 ! add by JHC
-   real(rkind), allocatable :: data1(:,:), data2(:,:)
+   real(r4_kind), allocatable :: data1(:,:), data2(:,:)
    type (date_type)         :: Date1, Date2
    logical                  :: use_climo, use_annual
    logical                  :: I_am_initialized=.false.
-end type amip_interp_type
+end type amip_interp_type_r4
+
+!> @brief Contains information needed by the interpolation module (exchange_mod) and buffers
+!! data (r8_kind flavor).
+!> @ingroup amip_interp_mod
+type amip_interp_type_r8
+   private
+   type (horiz_interp_type) :: Hintrp, Hintrp2 ! add by JHC
+   real(r8_kind), allocatable :: data1(:,:), data2(:,:)
+   type (date_type)         :: Date1, Date2
+   logical                  :: use_climo, use_annual
+   logical                  :: I_am_initialized=.false.
+end type amip_interp_type_r8
 
 !> @addtogroup amip_interp_mod
 !> @{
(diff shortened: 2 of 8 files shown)


amip_interp/include/amip_interp_r4.fh
```diff
@@ -0,0 +1,71 @@
+#undef AMIP_INTERP_KIND_
```
mlee03, line 1: FMS_AMIP_INTERP_KIND_? :)

amip_interp/amip_interp.F90
```diff
@@ -265,31 +274,48 @@ module amip_interp_mod
...
 
 !----- public data type ------
 
-!> @brief Contains information needed by the interpolation module (exchange_mod) and buffers data.
+!> @brief Contains information needed by the interpolation module (exchange_mod) and buffers
+!! data. Real precision is determined by the rkind parameter, which defaults to r8_kind.
 !> @ingroup amip_interp_mod
-type amip_interp_type
+type amip_interp_type(rkind)
+   integer, kind            :: rkind = r8_kind ! Real kind for data1 and data2
```
mlee03, line 302: I think we want to avoid using parameterized derived types for now

amip_interp/include/amip_interp.inc
```diff
@@ -16,382 +16,44 @@
...
    type (time_type),         intent(in)    :: Time !< Time to interpolate
-   type (amip_interp_type),  intent(inout) :: Interp !< Holds data for interpolation
-   real,                     intent(out)   ::  sst(:,:) !< Sea surface temperature data
+   type (AMIP_INTERP_TYPE_),  intent(inout) :: Interp !< Holds data for interpolation
+   real(AMIP_INTERP_KIND_),                     intent(out)   ::  sst(:,:) !< Sea surface temperature data
    character(len=*), optional, intent(out) :: err_msg !< Holds error message string if present
 
-   real, dimension(mobs,nobs) :: sice
+   real(AMIP_INTERP_KIND_), dimension(mobs,nobs) :: sice
+   real(AMIP_INTERP_KIND_), allocatable:: temp1(:,:), temp2(:,:)
```
mlee03, line 29: is it safe to move temp1 and temp2 to here?

amip_interp/amip_interp.F90
```diff
@@ -146,18 +146,17 @@ module amip_interp_mod
...
 !-----------------------------------------------------------------------
 !----------------- Public Data -----------------------------------
 integer :: i_sst = 1200
 integer :: j_sst = 600
-real, parameter:: big_number = 1.E30
+real(r8_kind), parameter:: big_number = 1.E30
 logical :: forecast_mode = .false.
-real, allocatable, dimension(:,:) ::  sst_ncep, sst_anom
-
-public i_sst, j_sst, sst_ncep, sst_anom, forecast_mode, use_ncep_sst
```
mlee03, line 160: not sure about changing the public variable names or making some of these private  in mixedmode.  Perhaps these changes could be proposed to main instead.

mlee03 (changes requested):

amip_interp/amip_interp.F90
J-Lentz: I've replaced `amip_interp_type` with `amip_interp_type_r4` and `amip_interp_type_r8` in 323ef7b7da84b0ae73e535c9adf2a20e86a332ee. I don't think there's a way to export the name `amip_interp_type` as an alias for `amip_interp_type_r8`, so minor changes to the model code will be required. Fortunately, I believe `ice_amip` and `ocean_amip` are the only model codes which use this module.

amip_interp/include/amip_interp_r4.fh
J-Lentz: Renamed in a15e58ea3fb062be330ce3e29219d7f1b6f959fa.

amip_interp/include/amip_interp.inc
J-Lentz: I think so, but just to err on the side of caution, I've added the `save` attribute in 3e0db621c989b957ad4506fa4abcfae1e8869b2e.

amip_interp/amip_interp.F90
J-Lentz: One alternative could be to create `sst_ncep` and `sst_anom` pointers as aliases for the r8 versions, but this seems unnecessary if no model code is actually using them.

mlee03: They could be used by non-GFDL models, @bensonr @laurenchilutti?

bensonr:
amip_interp_mod is used in external_sst.F90
<https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/tools/external_sst.F90>
in all branches and GFDL/atmosphere.F90
<https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/driver/GFDL/atmosphere.F90>
for GFDL climate models.

amip_interp/amip_interp.F90
bensonr: @mlee03 @laurenchilutti - those public entities from amip_interp_mod are used in [external_sst.F90](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/tools/external_sst.F90) in all branches and [GFDL/atmosphere.F90](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/driver/GFDL/atmosphere.F90) for GFDL climate models.

amip_interp/include/amip_interp.inc
```diff
@@ -418,7 +80,7 @@ if ( .not.use_daily ) then
    if ( .not. allocated(temp2) ) allocate (temp2(mobs,nobs))
 
    if (use_zonal) then
-      call zonal_sst (Amip_Time, sice, temp1)
+      call ZONAL_SST_ (Amip_Time, sice, temp1)
```
mlee03, line 83: Nice, this seems to be better than declaring bunch of generic interfaces for private procedures

amip_interp/amip_interp.F90
mlee03: @bensonr thank you, @J-Lentz would you be able to change these back to their original names? :)

mcallic2 (approved):

amip_interp/amip_interp.F90
J-Lentz: Sorry about that, I'd gotten the impression that this module is only used by `ice_amip` and `ocean_amip`. d8d70837fea3f26e79da605ff45ca0abcfd34515 restores the public data and adds public pointers to the r8 versions of `sst_ncep` and `sst_anom`.

amip_interp/amip_interp.F90
```diff
@@ -265,31 +280,59 @@ module amip_interp_mod
...
+end interface amip_interp_del
 
 !----- public data type ------
 
-!> @brief Contains information needed by the interpolation module (exchange_mod) and buffers data.
+!> @brief Contains information needed by the interpolation module (exchange_mod) and buffers
+!! data (r4_kind flavor).
 !> @ingroup amip_interp_mod
-type amip_interp_type
+type amip_interp_type_r4
```
mlee03, line 307: Sorry didn't realize amp_interp_type was public.  Is there anyway to mixed precision to amip_interp without changing the name of amip_interp_type?

amip_interp/include/amip_interp.inc
```diff
@@ -452,25 +114,25 @@ if ( .not.use_daily ) then
...
       endif
 
 !-----------------------------------------------------------------------
 
       if (Date2 /= Interp % Date2) then
-          call read_record ('sst', Date2, Udate2, temp2)
+          call READ_RECORD_ ('sst', Date2, Udate2, temp2)
           if ( use_ncep_sst .and. (.not. no_anom_sst) ) then
-               temp2(:,:) = temp2(:,:) + sst_anom(:,:)
+               temp2(:,:) = temp2(:,:) + SST_ANOM_ (:,:)
```
mlee03, line 132: For the case where sst_anom_r4 is used... that means sst_anom_r8 is not used and the pointer sst_anom is pointing to sst_anom_r8, how does the user retrieve the sst_anom_r4 data?

amip_interp/amip_interp.F90
J-Lentz: This could be done by making `data1` and `data2` both `class(*)` and wrapping all code that uses them in `select type` blocks.

CMakeLists.txt
```diff
@@ -323,7 +323,8 @@ foreach(kind ${kinds})
                                                  tracer_manager/include
                                                  interpolator/include
                                                  coupler/include
-                                                 data_override/include)
+                                                 data_override/include
```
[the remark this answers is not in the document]
J-Lentz: Fixed in d1924ee687dc6d4f1e226cb8ac0ac83bd8001169.

amip_interp/amip_interp.F90
```diff
@@ -146,18 +146,18 @@ module amip_interp_mod
...
+public amip_interp_init, amip_interp_init_r4, amip_interp_init_r8, get_amip_sst, &
+     & get_amip_ice, amip_interp_new, amip_interp_del, amip_interp_type_r4, &
+     & amip_interp_type_r8, assignment(=)
 
 !-----------------------------------------------------------------------
 !----------------- Public Data -----------------------------------
 integer :: i_sst = 1200
 integer :: j_sst = 600
-real, parameter:: big_number = 1.E30
+real(r8_kind), parameter:: big_number = 1.E30
```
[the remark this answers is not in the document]
J-Lentz: Fixed in fb78dae60417e134dcdc2c8f2ac3181ab5fed580.

amip_interp/amip_interp.F90
J-Lentz: `amip_interp_type` has been brought back in fb78dae60417e134dcdc2c8f2ac3181ab5fed580.

amip_interp/include/amip_interp.inc
J-Lentz: I think it's not possible to allow both `r4_kind` and `r8_kind` access in a way that's backward-compatible. We could make `sst_anom` and `sst_ncep` both `class(*), allocatable`, but then any user code that accesses them would have to be wrapped in a `select type` block.

amip_interp/amip_interp.F90
```diff
@@ -265,31 +275,39 @@ module amip_interp_mod
...
 type amip_interp_type
    private
-   type (horiz_interp_type) :: Hintrp, Hintrp2 ! add by JHC
-   real, allocatable        ::    data1(:,:),  data2(:,:)
-   type (date_type)         ::    Date1,       Date2
-   logical                  :: use_climo, use_annual
-   logical                  :: I_am_initialized=.false.
-end type
+   type (horiz_interp_type)              :: Hintrp, Hintrp2 ! add by JHC
+   class(*), dimension(:,:), allocatable :: data1, data2
```
mlee03, line 297: I see pointers are being used in the .inc file cause there are class(*)? ...
Since data1 and data2 are private, can it be 
``real(r4_kind), dimension(:,:) :: data1_r4, data2_r4``
      ``real(r8_kind), dimension(:,:) :: data1_r8, data2_r8``

Then in the .inc file, the these arrays can be accessed by Interp%DATA1_KIND_ where DAT1_KIND_ is defined as data1_r4 in the .fh file.  @rem1776  came up with the idea and we've been doing something like that for other modules.

mcallic2 (approved):

amip_interp/amip_interp.F90
```diff
@@ -146,16 +146,17 @@ module amip_interp_mod
...
 !-----------------------------------------------------------------------
 !----------------- Public Data -----------------------------------
 integer :: i_sst = 1200
 integer :: j_sst = 600
-real, parameter:: big_number = 1.E30
+real(r8_kind), parameter:: big_number = 1.E30_r8_kind
 logical :: forecast_mode = .false.
-real, allocatable, dimension(:,:) ::  sst_ncep, sst_anom
+real(r8_kind), allocatable, dimension(:,:) ::  sst_ncep,    sst_anom    !< r8 versions of sst_ncep and sst_anom
+real(r4_kind), allocatable, dimension(:,:) ::  sst_ncep_r4, sst_anom_r4 !< r4 versions of sst_ncep and sst_anom
```
mlee03, line 159: I think it might be best to go with real(r8_kind) sst_ncep, sst_anom and cast it to r4_kind if necessary in the procedures.  Using sst_ncep_r4 and sst_anom_r4 would require way too many changes to the components that is out of scope with this project for now.

J-Lentz: `sst_ncep_r4` and `sst_anom_r4` removed in cf84442e4163db85472c07529130402009c6915c.

amip_interp/include/amip_interp.inc
```diff
@@ -519,8 +180,8 @@ else
 
     mobs_sst = 1440;  nobs_sst = 720
 
-    call set_sst_grid_edges_daily(mobs_sst, nobs_sst)
-    call horiz_interp_new ( Interp%Hintrp2, lon_bnd, lat_bnd, &
+    call SET_SST_GRID_EDGES_DAILY_ (mobs_sst, nobs_sst)
+    call horiz_interp_new ( Interp%Hintrp2, LON_BND_, LAT_BND_, &
```
mlee03, line 184: Curious as to how this work.  Following how lon_bnd and lat_bnd are set in amip_interp_init , they're always going to be in r8_kind.  Wondering what happens if lon_model and lat_model are r4_kind...

J-Lentz: In that case, the module should be initialized with `amip_interp_init_r4` rather than `amip_interp_init` (which is just an alias for `amip_interp_init_r8`), so that `LON_BND_` and `LAT_BND_` refer to the `r4_kind` versions.

mlee03: I think we're trying to minimize the number of changes required on the components side.  It would be nice if the users can just call amip_interp_init instead of making changes to their code to call amip_interp_init_r4

J-Lentz: It's a similar problem to `data_override`, where the initialization routine allocates real data but doesn't take any real arguments. We could take a similar approach here, e.g.,`amip_interp_init(mode=r4_kind)` or `amip_interp_init(mode=r8_kind)`, but I don't think we can avoid user code having to explicitly specify the kind unless we make `tempamip`, `lon_bnd`, and `lat_bnd` all `r8_kind`. Or perhaps make `amip_interp_init` a no-op and have `amip_interp_new` do the actual initialization.

mlee03: Would it bad to make those arrays r8_kind?  It seems like that would make things simpler..

J-Lentz: We could, although I'm not sure what the performance implications (if any) would be.

mcallic2 (approved):

libFMS.F90
```diff
@@ -73,18 +73,13 @@ module fms
...
   use amip_interp_mod, only: fms_amip_interp_init         => amip_interp_init, &
-                             fms_amip_interp_get_amip_sst             => get_amip_sst, &
-                             fms_amip_interp_get_amip_ice             => get_amip_ice, &
+                             fms_amip_interp_get_amip_sst => get_amip_sst, &
+                             fms_amip_interp_get_amip_ice => get_amip_ice, &
                              fms_amip_interp_new          => amip_interp_new, &
                              fms_amip_interp_del          => amip_interp_del, &
-                             FmsAmipInterp_type => amip_interp_type, &
-                             assignment(=), &
-                             fms_amip_interp_i_sst        => i_sst, &
```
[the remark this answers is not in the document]
J-Lentz: They were removed because I made them private in the first iteration of this PR, and then I neglected to add them back in. `libFMS.F90` has been restored to its original state in 58923c69613ad497d58d16391a2b1db8fd43835e.

mlee03 (approved):

rem1776 (approved):

PR #1376, opened 2023-09-22
Add tile number to the filename for subregional files

### LINKED ISSUE #1008: Add tileX to regional diag_manager output
**Describe the bug**
The current diag_manager does not add "tileX" to the filename when using a cube sphere, which leads to trouble when trying to combine the files and regrid them (if the region is in two different tiles)

I think this is a bug that was introduced when fms2_io was implemented in the diag_manager, mpp_io's diag_manager was adding tileX. 

**To Reproduce**
Add a regional diagnostic to the diag_table with a test case that is in a cube_sphere domain. You will see .tileX is not in the filename. See helpdesk ticket #231128 for an example 

**Expected behavior**
Diag_manager should be adding tileX to the filename OR if regions in multiple tiles are not going to be supported there should be a clear error message. 

This is the code to update:
https://github.com/NOAA-GFDL/FMS/blob/1003bebef9f2884e92eb209c85ac7414ef5cd9e8/diag_manager/diag_output.F90#L139-L140

**System Environment**
This happens in any system.

**Additional context**
N/A

uramirez8707:
**Description**
Add tile number to the filename for subregional files

Fixes #1008 

**How Has This Been Tested?**
This was tested with an AM4.2 amip run
and this table:
```
ESM4_amip_D1_am4p2_proto7b_isop_whiteCapsAlbedo_salt_nudge_ch4_2d_adj_vmr_high_terpene_SIS2
1870 1 1 0 0 0

"US_aircraft",1,"hours",1,"days","time"
"dynamics", "H2", "H2", "US_aircraft", "all", .true., "232.50,292.50,25.00,50.00,239.00,1000.00", 2
```

Before this update tile was not added to the filename and now it is. This update is needed in order for the combiner to work correctly.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -111,6 +111,8 @@ CONTAINS
     integer, allocatable, dimension(:) :: current_pelist
     integer :: mype  !< The pe you are on
     character(len=9) :: mype_string !< a string to store the pe
+    character(len=128) :: tmp_filename      !< Temp variable to store the file_name
+
     !---- initialize mpp_io ----
     IF ( .NOT.module_is_initialized ) THEN
        module_is_initialized = .TRUE.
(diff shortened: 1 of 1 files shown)


uramirez8707:
@rem1776 @thomas-robinson can this be included in the 2023.03?
This came up in helpdesk ticket #5023155

diag_manager/diag_output.F90
```diff
@@ -111,6 +111,8 @@ SUBROUTINE diag_output_init (file_name, file_title, file_unit,&
     integer, allocatable, dimension(:) :: current_pelist
     integer :: mype  !< The pe you are on
     character(len=9) :: mype_string !< a string to store the pe
+    character(len=128) :: tmp_filename      !< Temp variable to store the file_name
```
thomas-robinson, line 114: What makes this a temporary variable?  It looks like a file name for a specific type of file.

thomas-robinson (changes requested):

diag_manager/diag_output.F90
thomas-robinson: The PR description does a better job of describing this variable than either its name or the description here.

thomas-robinson (approved):

PR #1407, opened 2023-11-13
fix diag_send_complete loops and add get_file_ids

rem1776:
**Description**
moves the loops around in diag_send_complete so that it goes through the fields, and then the files the field is in.

I've had this as part of the average updates but figured it would better to put it in separately.

Fixes # (issue)

**How Has This Been Tested?**

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -667,38 +667,26 @@ CALL MPP_ERROR(FATAL,"You can not use the modern diag manager without compiling 
     file_ids = diag_field%get_file_ids()
     math = diag_field%get_math_needs_to_be_done()
     ! if doing math loop through each file for given field
-    if (size(file_ids) .ge. 1 .and. math) then
-      file_loop: do ifile = 1, size(file_ids)
-        diag_file => this%FMS_diag_files(ifile)
-        ! if the file is not allocated go away
-        if(.not. allocated(diag_file%FMS_diag_file)) then
-          if(DEBUG_SC) call mpp_error(NOTE, "file id:"//string(ifile)//" not allocated for field:"//diag_field%get_varname())
-          cycle
-        endif
-        ! If the field is not registered go away
-        if (.not. diag_file%FMS_diag_file%is_field_registered(ifield)) then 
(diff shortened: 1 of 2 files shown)


diag_manager/fms_diag_object.F90
```diff
@@ -648,43 +649,58 @@ subroutine fms_diag_send_complete(this, time_step)
...
-        diag_field => this%FMS_diag_fields(file_field_ids(ifield))
-        !> Check if math needs to be done
-        math = diag_field%get_math_needs_to_be_done()
-        calling_math: if (math) then
+        if (.not. diag_file%FMS_diag_file%is_field_registered(ifield)) then 
+          if(DEBUG_SC) call mpp_error(NOTE, "file id:"//string(ifile)//" not registered for field:"//diag_field%get_varname())
+          cycle
+        endif
+        ! Check if buffer alloc'd
+        has_input_buff: if (diag_field%has_input_data_buffer()) then
```
uramirez8707, line 684: I think this if statement can be moved to outside the file loop. 
`allocate_diag_field_output_buffers` and `fms_diag_do_reduction` are looping through the number of buffers for the field. The number of buffers for the field is the same as the number of files the field is in, so i think we will be doing the reductions twice here.

diag_manager/fms_diag_object.F90
```diff
@@ -648,43 +649,58 @@ subroutine fms_diag_send_complete(this, time_step)
...
-        if (associated(diag_field)) nullify(diag_field)
-      enddo field_loop
-      !> Clean up, clean up, everybody do your share
-      if (allocated(file_field_ids)) deallocate(file_field_ids)
-    endif field_outer_if
-  enddo file_loop
-
-  call this%fms_diag_do_io()
+                                                         " -"//trim(error_string)))
+        endif has_input_buff
```
uramirez8707, line 694: Can we add an else here and error out with a message saying that `math_needs_to_be_done` is true, but there is no input buffer?

uramirez8707 (changes requested):

uramirez8707 (approved):

PR #1415, opened 2023-11-20
revert argument changes for xgrid

rem1776:
**Description**
The data argument was changed as part of #1355, but this broke some of the coupler code (and likely other components). This reverts the xgrid changes to arguments.

**How Has This Been Tested?**
make check on amd, ci

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/exchange/xgrid.F90
+++ b/exchange/xgrid.F90
@@ -1457,18 +1457,18 @@ end subroutine get_grid_version2
 
 !#######################################################################
 !> @brief Read the area elements from NetCDF file
-subroutine get_area_elements_fms2_io(fileobj, name, get_area_data)
+subroutine get_area_elements_fms2_io(fileobj, name, data)
   type(FmsNetcdfDomainFile_t), intent(in) :: fileobj
   character(len=*), intent(in)            :: name
-  real(r8_kind), intent(out)              :: get_area_data(:,:)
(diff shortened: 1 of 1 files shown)


thomas-robinson:
Let's keep this open while we assess how much work it will be to update coupler and the components.  I would rather have good code with good coding practice than revert back to the old coding standards.

bensonr (commented):
Was there an issue with testing requiring these to be reverted?  I can see us having to do it for optionals, but why would it affect required arguments?

rem1776:
> Was there an issue with testing requiring these to be reverted? I can see us having to do it for optionals, but why would it affect required arguments?

You can specify the argument name for required arguments as well, its just not as common, so it ends up being the same issue as the optional arguments.

bensonr:
@rem1776 - much of this could become obsolete if the stock integration/calculations are removed as part of the coupler refactor project.  I'll be taking some of that discussion to the appropriate body within GFDL.

rem1776:
> @rem1776 - much of this could become obsolete if the stock integration/calculations are removed as part of the coupler refactor project. I'll be taking some of that discussion to the appropriate body within GFDL.

We made the change in the coupler to fix this so I'll close this one. I think that works out better cause then these calls can just be removed as part of the refactor project.

PR #1416, opened 2023-11-20
ci: cancel runs if newer commit available and update checkout action

rem1776:
**Description**
This adds the same setting Lauren added to the FV3 parallelworks CI. It'll cancel a run in progress if a newer commit is available.

This should help with storage usage and we should have quicker checks with PR's as well since it won't be wasting time on the older commits.

also updates all our checkout actions versions to the newest since the old one was deprecated.

**How Has This Been Tested?**
ci

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/.github/workflows/github_autotools_gnu.yml
+++ b/.github/workflows/github_autotools_gnu.yml
@@ -27,7 +27,7 @@ jobs:
         SKIP_TESTS: "test_yaml_parser.5" # temporary till fixes are in
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
     - name: Prepare GNU autoconf for build
       run: autoreconf -if
     - name: Configure the build
--- a/.github/workflows/github_autotools_gnu.yml
+++ b/.github/workflows/github_autotools_gnu.yml
@@ -4,6 +4,11 @@ name: Build libFMS test with autotools
 
 on: [push, pull_request]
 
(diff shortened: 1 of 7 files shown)


laurenchilutti:
> How tested is this? It's not going to cancel a workflow from a different PR, right?

My understanding is that it will cancel any in process action that matched the conceurrency group. So we name the concurrency group ${{ github.workflow }}-${{ github.ref }} which translates to `<name-of-workflow>-<refs/pr/##>` for a PR.  So this will only cancel in process runs of the same workflow and PR.

I have not tested this on workflow triggered by"push" as in the dycore, we only trigger on PRs

laurenchilutti:
When I did this for the dycore, I tested it in my PR that brought in the workflow yaml changes (the equivalent of this PR for FMS). You can keep making dummy commits to test that it properly cancels the previous commits CI on this PR

rem1776:
> When I did this for the dycore, I tested it in my PR that brought in the workflow yaml changes (the equivalent of this PR for FMS). You can keep making dummy commits to test that it properly cancels the previous commits CI on this PR

Thanks @laurenchilutti , I'll try that now and check the how it works with the pushes on my fork.

rem1776:
@laurenchilutti  @thomas-robinson  looks like its working as intended, will only cancel jobs on the same branch with pushes as well.

thomas-robinson (approved):

laurenchilutti (approved):

PR #1458, opened 2024-02-09
dmUpdate: move current_model_time to the field object

rem1776:
**Description**
moves the current_model_time variable from the main diag object to the individual field objects

I'm not too sure if the check under `all fields should have the same time if writing` should be there.

**How Has This Been Tested?**
latest oneapi on amd

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -62,7 +62,6 @@ private
                                                                        !! one for each variable in the diag_table.yaml
   integer, private :: registered_buffers = 0 !< number of registered buffers, per dimension
   class(fmsDiagAxisContainer_type), allocatable :: diag_axis(:) !< Array of diag_axis
-  !type(time_type)  :: global_model_time !< The model time for the whole object
   integer, private :: registered_variables !< Number of registered variables
   integer, private :: registered_axis !< Number of registered axis
   logical, private :: initialized=.false. !< True if the fmsDiagObject is initialized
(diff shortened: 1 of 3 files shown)


thomas-robinson (approved):

diag_manager/fms_diag_object.F90
```diff
@@ -777,39 +784,54 @@ subroutine fms_diag_do_io(this, is_end_of_run)
...
+        cycle
+      else
+        is_writing = .true.
+      endif
+
+      ! all fields should have the same time if writing
+      if(ibuff .eq. 1 ) &
+        model_time = diag_field%current_model_time
+      if(diag_field%current_model_time .ne. model_time) then
+        call mpp_error(FATAL, "fms_diag_do_io:: each field's model time should be the same for a given file")
```
uramirez8707, line 813: This is not always the case. There are many different time steps:
dt_atmos = 1800, dt_cpld  = 7200, rad_time_step= 10800, sw_rad_time_step = 3600

uramirez8707 (changes requested):

diag_manager/fms_diag_object.F90
rem1776: I'll take that out, but I'm not sure what that means for the `is_writing` if block below since its using the model time to update the file objects:
 https://github.com/rem1776/FMS/blob/9809760ce34b8d1bb6f8b81fc1b9552e49b1c13e/diag_manager/fms_diag_object.F90#L837-L842

I was thinking of adding a loop for each field's current model time (or just each unique model time), just not sure if that would mess with the write routines.

PR #1464, opened 2024-02-23
Modern diag manager: Updates for the mask_variant=.true. case

uramirez8707:
**Description**
1. Makes the `weight_sum` an array instead of a scalar:
     - If the mask changes over time (i.e mask_variant is set to .true. in the register_diag_field call), weight_sum is set to be the size of the compute domain, otherwise it is set to to (1,1,1,1)
     - The weight sum is increased at every grid point only if it is not masked at that grid point
     - If a grid point is masked at every time step, the output buffer will be set to the fill_value
**This is different from old behavior where the output buffer will be set to 0 instead of the fill value!**
2.  Adds a unit to test to test the `mask_variant=.true.` case
3.  Since the `allocate_buffer` interface was changed, the unit tests were also updated to account for this
4.  Updates the logic in `update_buffer_time` so that if send_data has not been called, it still counts as a new model time. This was added because if send_data was called at initialization `new_time` was set to .false. so the counter was off by one

Fixes # (issue)

**How Has This Been Tested?**
CI, included added tests 
1 year long run of `c96L33_am4p0_cmip6Diag`

**Checklist:**
- [x] `make distcheck` passes


test_fms/diag_manager/test_var_masks.F90
```diff
@@ -0,0 +1,90 @@
...
+
+    var1_mask = .True.
+    !< The first point is going to be asked every other hour
+    if (mod(i,2) .eq. 0) var1_mask(1,1) = .False.
+    var1_data = real(i)
+    used = send_data(id_var1, var1_data, Time, mask=var1_mask)
+
+    ! Double send_data calls at the same time. This is similar to what happens with the
+    ! `tntpbl` and `tnhuspbl` diagnostics in the physics
+    ! TODO used = send_data(id_var1, var1_data, Time, mask=var1_mask)
```
rem1776, line 84: Are we gonna leave this in as a todo? just wondering if i should open an issue to expand on this test after it goes in.

rem1776 (approved):

test_fms/diag_manager/test_var_masks.F90
uramirez8707: Good point. This is not going to be done anymore, as it will be fixed in the physics code instead.

thomas-robinson (approved):

PR #1468, opened 2024-02-29
Modern diag manager: Fixes for when buffering the data

uramirez8707:
**Description**
- Adds `send_data_time` to the input_data_buffer object. This will be used when we are buffering the data (i.e running with multiple threads). Without this update the correct time was not used when buffering the data which was causing diagnostic changes when running with and without threads. 
- Moves the code that does the math outside of diag_send_complete to its own function `do_buffer_math`
- Adds a call to `do_buffer_math` in diag_manager_end to finish doing any math that wasn't done before (without this change, there were diagnostic data diffs at the last time step)

Fixes # (issue)

**How Has This Been Tested?**
CI
AM4 1 year long runs.

**Checklist:**
- [x] `make distcheck` passes


diag_manager/fms_diag_field_object.F90
```diff
@@ -406,6 +408,24 @@ subroutine set_vartype(objin , var)
...
+!> @brief Sets the send data time
+subroutine set_send_data_time (this, time)
+  class (fmsDiagField_type) , intent(inout):: this                !< The field object
+  type(time_type),            intent(in)   :: time                !< Current model time
+
+  call this%input_data_buffer%set_send_data_time(time)
+end subroutine set_send_data_time
+
+!> @brief Get the send data time
+!! @result Send data time
```
thomas-robinson, line 420: Could you make the description of what "send data time" is a little more detailed. Something like "The time that was send by the model component to send_data".  You have a better description in `fmsDiagInputBuffer_t`

diag_manager/fms_diag_field_object.F90
```diff
@@ -406,6 +408,24 @@ subroutine set_vartype(objin , var)
  end select
 end subroutine set_vartype
 
+!> @brief Sets the send data time
+subroutine set_send_data_time (this, time)
```
thomas-robinson, line 412: Are these wrappers over the other set_send_data_time routines?  This seems convoluted.

thomas-robinson (changes requested):

diag_manager/fms_diag_field_object.F90
uramirez8707: This is the same thing that was done with the file and yaml object:
https://github.com/NOAA-GFDL/FMS/blob/dmUpdate/diag_manager/fms_diag_file_object.F90#L125-L140

It was done like this for convenience.

thomas-robinson (approved):

diag_manager/fms_diag_field_object.F90
thomas-robinson: Sure, but the description is wrong.  This doesn't set anything.  It calls `input_data_buffer%set_send_data_time` which does the setting.

PR #1473, opened 2024-03-04
dmUpdate: fix nvhpc compile error

rem1776:
**Description**
Fixes a nvidia compilation error coming from the diag_manager updates. I missed this one in PR #1463 because i was only using autotools. Interestingly this error seems to only happen with mkmf, even if i use the same target flags for autotools.

Just adds a nested select type for when assigning a class(*) field to a character variable.

**How Has This Been Tested?**
mkmf on gaea with nvhpc/22.3, autotools build on amd with nvhpc/23.11

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/exchange/xgrid.F90
+++ b/exchange/xgrid.F90
@@ -1542,7 +1542,8 @@ subroutine setup_xmap(xmap, grid_ids, grid_domains, grid_file, atm_grid, lnd_ug_
   integer                                      :: lnd_ug_id, l
   integer,                         allocatable :: grid_index(:)
   type(FmsNetcdfFile_t)                        :: gridfileobj, mosaicfileobj, fileobj
-  type(grid_type), allocatable, target         :: grids_tmp(:)
+  type(grid_type), allocatable, target         :: grids_tmp(:) !< added for nvhpc workaround, stores xmap's
+                                                               !! grid_type array so we can safely point to it
 
   call mpp_clock_begin(id_setup_xmap)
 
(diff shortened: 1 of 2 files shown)


exchange/xgrid.F90
```diff
@@ -1593,9 +1595,18 @@ subroutine setup_xmap(xmap, grid_ids, grid_domains, grid_file, atm_grid, lnd_ug_
   endif
 
   call mpp_clock_begin(id_load_xgrid)
-  do g=1,size(grid_ids(:))
-     grid => xmap%grids(g)
-     if (g==1) grid1 => xmap%grids(g)
+
+  ! nvhpc compiler workaround
+  ! saves passed in xmap as an allocatable
+  ! this lets us use a normal assignment instead of a pointer assignment in the loop to avoid a compiler error
```
thomas-robinson, line 1601: Do we need to report this to nvidia?

bensonr: Let's make sure we have tested using the very latest nvhpc 24.3 before filing a bug with NVidia.

rem1776: The xgrid compilation error was pretty hard to reproduce, Niki first saw this using 22.3 and I've only been able to get the error using his xml to compile. I don't seem to get the error when i compile with autotools, and didn't even get it when i tried compiling one of my xml's with nvidia.

thomas-robinson (approved):

PR #1478, opened 2024-03-12
Modern_diag_manager :: multiple_send_data calls option

uramirez8707:
**Description**
- Adds an optional variable (`multiple_send_data`) to register_diag_field, which if set to .True. it indicates that send_data is called multiple times for the same time
- If that option is set to .True. the data will be buffered
   - If the field's mask changes over time, the input buffer will be updated only when the mask is set to .True. This will be the case for some of the river diagnostics (https://gitlab.gfdl.noaa.gov/uriel.ramirez/land_lad2/-/blob/multiple_send_data_fix/river/river_physics.F90?ref_type=heads#L161-L166)
   - If the field's mask does not change over time , the input_buffer will be added and a counter will be added. Once the code hits diag_send_complete it will divide the field by the number of times send _data is called). This will be the case for some variable in vert_diff_driver (https://gitlab.gfdl.noaa.gov/fms/atmos_phys/-/commit/e1bf0539060a4b1e9b63f0e3fbfaf65f663c6c08)
- Modifies logic so that the time dimension will be written at least once. This is needed so that the combiner works for when doing regression testing.  

Fixes # (issue)

**How Has This Been Tested?**
CI
AM4 - the river diagnostics and the vert_diff_drivers were changing answers before this update

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_field_object.F90
+++ b/diag_manager/fms_diag_field_object.F90
@@ -95,6 +95,7 @@ type fmsDiagField_type
      procedure :: setID => set_diag_id
      procedure :: set_type => set_vartype
      procedure :: set_data_buffer => set_data_buffer
+     procedure :: prepare_data_buffer
      procedure :: set_data_buffer_is_allocated
      procedure :: set_send_data_time
      procedure :: get_send_data_time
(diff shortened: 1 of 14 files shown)


rem1776 (approved):

diag_manager/fms_diag_field_object.F90
```diff
@@ -426,10 +441,30 @@ function get_send_data_time(this) &
   rslt = this%input_data_buffer%get_send_data_time()
 end function get_send_data_time
 
+!> @brief Prepare the input_data_buffer to do the reduction method
+subroutine prepare_data_buffer(this)
+  class (fmsDiagField_type) , intent(inout):: this                !< The field object
+
+  if (.not. this%multiple_send_data) return
+  if (this%mask_variant) return
```
thomas-robinson, line 449: looks like these could be combined

thomas-robinson: Why not wrap this function with these IF statements?

thomas-robinson (commented):
Does this need to have a test added?

uramirez8707:
> Does this need to have a test added?

A test has been added

diag_manager/fms_diag_field_object.F90
uramirez8707: I think it is cleaner like this.

PR #1486, opened 2024-04-09
dmUpdate: fixes for gcc

rem1776:
**Description**
Fixes failures in a bunch of unit tests when compiling with gcc and runtime checks enabled.

The majority of these fixes are related to derived type assignments. There seems to be a compiler bug that breaks assignments with certain derived types (in this case its diagYamlFiles_type and diagYamlObject_type), so the workaround is to just point to the objects instead of doing a full copy.

**How Has This Been Tested?**
Tested on amd box with gcc 13 + `-g -O0 -fbacktrace -fcheck=all`
and also with oneapi

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_field_object.F90
+++ b/diag_manager/fms_diag_field_object.F90
@@ -1193,6 +1193,8 @@ subroutine write_field_metadata(this, fms2io_fileobj, file_id, yaml_id, diag_axi
   character(len=120)                       :: cell_methods!< Cell methods attribute to write
   integer                                  :: i           !< For do loops
   character (len=MAX_STR_LEN), allocatable :: yaml_field_attributes(:,:) !< Variable attributes defined in the yaml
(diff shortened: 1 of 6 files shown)


diag_manager/fms_diag_field_object.F90
```diff
@@ -1235,8 +1237,10 @@ subroutine write_field_metadata(this, fms2io_fileobj, file_id, yaml_id, diag_axi
   endif
 
   if (this%has_interp_method()) then
-    call register_variable_attribute(fms2io_fileobj, var_name, "interp_method", this%get_interp_method(), &
-      str_len=len_trim(this%get_interp_method()))
+    interp_method_tmp = this%interp_method
```
thomas-robinson, line 1240: Is this to deal with some compiler peculiarity?

rem1776: Yeah the previous code triggers an internal compiler error with gcc

thomas-robinson (approved):

PR #1493, opened 2024-04-23
fix hdf5 compiler check

### LINKED ISSUE #1492: configure fails on gaea
**Describe the bug**
Configure fails on gaea with the following message:
configure: error: The HDF5 version used to build netcdf is incompatible with the set FCFLAGS. NetCDF must be built with a HDF5 version other than 1.14.3 to support floating point exception traps.

Even though I don't have the floating point exception traps flag on

**To Reproduce**

```
module load intel-classic/2022.2.1
module load cray-hdf5/1.12.2.3
module load cray-netcdf/4.9.0.3

autoreconf 
mkdir build
cd build
../configure FC=ftn CC=cc FCFLAGS="-O0 -g" LDFLAGS="-O0 -g"
```

**Expected behavior**
Configure should work

**System Environment**
GAEA with the environment above

**Additional context**
I think the test program should have nc_create instead of nc_open:
https://github.com/NOAA-GFDL/FMS/blob/main/configure.ac#L268

rem1776:
**Description**
Changes the configure check used to determine if a bug in hdf5 is present. Before it was using nf90_open which wouldn't work on certain systems.

Fixes #1492 

**How Has This Been Tested?**
tested on gaea, amd box, and my laptop

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -265,7 +265,7 @@ AC_MSG_CHECKING([if HDF5 version causes floating point exceptions with set flags
 AC_RUN_IFELSE([AC_LANG_PROGRAM([], [[
       use netcdf
       integer i, j
-      j = nf90_open("test.nc", NC_WRITE, i)
+      j = nf90_create("test.nc", NC_WRITE, i)
 ]])], [hdf5_fpe_bug=no], [hdf5_fpe_bug=yes])
 AC_MSG_RESULT([$hdf5_fpe_bug])
 if test $hdf5_fpe_bug = yes; then
--- a/Makefile.am
+++ b/Makefile.am
@@ -125,10 +125,10 @@ check-code-coverage: check
 .PHONY: check-code-coverage
 
 clean-local:
-	-rm -rf .mods coverage-data coverage-report
+	-rm -rf .mods coverage-data coverage-report test.nc
 else
 clean-local:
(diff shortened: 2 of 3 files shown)


configure.ac
```diff
@@ -265,7 +265,7 @@ AC_MSG_CHECKING([if HDF5 version causes floating point exceptions with set flags
 AC_RUN_IFELSE([AC_LANG_PROGRAM([], [[
       use netcdf
       integer i, j
-      j = nf90_open("test.nc", NC_WRITE, i)
+      j = nf90_create("test.nc", NC_WRITE, i)
```
bensonr, line 268: Is it guaranteed that test.nc will not exist?  If so, we are fine.  If not, we should make it NC_OVERWRITE.

rem1776: I think autotools does the compile and runtime checks in some temp directory. Couldn't find anything specific in the docs but in my testing the check worked fine with that filename present in the build and source directories.

bensonr (dismissed):

rem1776:
This needed one other change for distcheck to pass. I also ran into a separate issue with the monin test script when i was testing this on different systems so I just included the fix here, just needed to force the `rm` so it wouldn't hang.

bensonr (approved):

PR #1514, opened 2024-05-08
Simplify argument lists of subroutines in `get_grid_version_mod`

### LINKED ISSUE #1371: clean up get_grid_version routines for data_override
get_grid_version_1 and 2 have an argument for compute domain indices, but then queries the domain for global and compute indices. (only uses global domain indices though)

The arguments should be cleaned up to only provide whats needed

indices argument can probably be removed

**Additional context**
came up with mixed precision

J-Lentz:
**Description**
Remove compute domain indices from the argument lists of `get_grid_version_1` and `get_grid_version_2`. Query compute domain indices from `mpp_domains_mod` and check the lower bounds of `lon(:,:)` and `lat(:,:)` at runtime.

Fixes #1371

**How Has This Been Tested?**
Builds on the AMD box with Intel compiler version 2024.1.0.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/data_override/include/get_grid_version.inc
+++ b/data_override/include/get_grid_version.inc
@@ -47,7 +47,6 @@ subroutine GET_GRID_VERSION_1_(grid_file, mod_name, domain, lon, lat, min_lon, m
   call mpp_get_global_domain(domain, isg, ieg, jsg, jeg)
   call mpp_get_compute_domain(domain, isc, iec, jsc, jec)
 
-#ifdef _DEBUG
   if (lbound(lon,1).ne.isc .or. lbound(lon,2).ne.jsc) then
     call mpp_error(FATAL, "get_grid_version_1: Incorrect lower-bound dimensions of lon(isc:,jsc:)")
   endif
@@ -55,7 +54,6 @@ subroutine GET_GRID_VERSION_1_(grid_file, mod_name, domain, lon, lat, min_lon, m
   if (lbound(lat,1).ne.isc .or. lbound(lat,2).ne.jsc) then
     call mpp_error(FATAL, "get_grid_version_1: Incorrect lower-bound dimensions of lat(isc:,jsc:)")
   endif
-#endif
 
   select case(mod_name)
   case('ocn', 'ice')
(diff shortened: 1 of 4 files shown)


rem1776 (approved):

data_override/include/get_grid_version.inc
```diff
@@ -45,8 +44,16 @@ subroutine GET_GRID_VERSION_1_(grid_file, mod_name, domain, isc, iec, jsc, jec,
...
   call mpp_get_global_domain(domain, isg, ieg, jsg, jeg)
+  call mpp_get_compute_domain(domain, isc, iec, jsc, jec)
+
+  if (lbound(lon,1).ne.isc .or. lbound(lon,2).ne.jsc) then
+    call mpp_error(FATAL, "get_grid_version_1: Incorrect lower-bound dimensions of lon(isc:,jsc:)")
+  endif
+
+  if (lbound(lat,1).ne.isc .or. lbound(lat,2).ne.jsc) then
+    call mpp_error(FATAL, "get_grid_version_1: Incorrect lower-bound dimensions of lat(isc:,jsc:)")
+  endif
```
rem1776, line 56: @J-Lentz I noticed this was failing the GNU CI and looked into it a bit since it was coming from a test I wrote in the coupler subdirectory.

I was confused at first why this check kept failing since the passed in array had the right bounds, but after a little digging I realized that when passed in the array bounds get reset from whatever it may be to `1:<size of the array>`.

I thought it was a bug but apparently that's how fortans supposed to work(?). I think we can either leave the lower bounds in the argument declaration so it can be checked, or we could just check the size here instead.

rem1776 (changes requested):

data_override/include/get_grid_version.inc
J-Lentz: Since `lon` and `lat` are `intent(out)`, I've worked around this by having the `get_grid_version` subroutines do the allocation themselves, which avoids the need for this check. It also simplifies the calling code a bit.

PR #1538, opened 2024-06-03
set is_allocated in all horiz_interp_type_new routines

### LINKED ISSUE #1533: Issue with horiz_interp
flag to show that weight arrays have been allocated may not be set. 
optional arguments to bilinear interp_new cannot be used if using horiz_interp_new.

I assume that with FMS2 the desired use of horiz_interp_new is as follows. 

call horiz_interp_new ( Interp, T_grid%x, T_grid%y, lon_out, lat_out, interp_method="bilinear")
followed by a
call horiz_interp_del( Interp )

 I'm running into an issue where the interpolation point seems to be right on the corner of the grid. When this happens I end up at
call mpp_error(FATAL, "horiz_interp_bilinear_mod: No solution found")

I'll assume this issue cropped up previously as there used to be a workaround for this with the new_search and no_crash_when_not_found arguments.
However I can't send these down through the horiz_interp_new interface.

I used to call 
 call horiz_interp_bilinear_new (Interp, T_grid%x*DEG_TO_RAD, T_grid%y*DEG_TO_RAD,&
               & lon_out, lat_out, new_search=.true., no_crash_when_not_found=.true.)

If I use this format and call
call horiz_interp_bilinear_del( Interp )
I then run into an issue when it loops around and tries to call horiz_interp_new again. It gives an allocation error.
The reason is that even though
include/horiz_interp_bilinear.inc HORIZ_INTERP_BILINEAR_NEW_2D
allocates
Interp%HI_KIND_TYPE_%wti 
etc.
it does not set the value of
Interp% HI_KIND_TYPE_ % is allocated.
That is done in include/horiz_interp.inc HORIZ_INTERP_NEW

But when I call horiz_interp_bilinear_del it checks the value of is_allocated and neither the Real4 or Real8 logical is true. So it doesn't deallocate the weights and when I call the next horiz_interp_new I find they are already allocated.

How to get around this?

1) Add the optional arguments to horiz_interp_new and pass them to horiz_interp_bilinear_new
(this would allow me to use horz_interp_new/del)
and
2) Set Interp%XXX%is_allocated to true in each _new routine (bilinear,spherical, conserve)
?

rem1776:
**Description**
Sets the `is_allocated` field as well as the `method_type` within horiz_interp_types when called from any of the method specific horiz_interp_new routines. This should prevent reallocation errors as described in the issue below.

Also adds some calls to the unit test to test this functionality and renames one of the CPP macros to make more sense.

Fixes #1533

**How Has This Been Tested?**
make distcheck with oneapi 2024.01 on amd box

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/horiz_interp/test_horiz_interp.F90
+++ b/test_fms/horiz_interp/test_horiz_interp.F90
@@ -1071,7 +1071,7 @@ implicit none
         call horiz_interp_del(Interp_new1)
         call horiz_interp_del(Interp_new2)
         call horiz_interp_del(Interp_cp)
-        ! test deletion after direct calls 
+        ! test deletion after direct calls
         call horiz_interp_conserve_new(Interp_new1, lon_in_1d, lat_in_1d, lon_out_1d, lat_out_1d)
         call horiz_interp_del(Interp_new1)
         call horiz_interp_conserve_new(Interp_new1, lon_in_1d, lat_in_1d, lon_out_2d, lat_out_2d)
@@ -1102,7 +1102,7 @@ implicit none
         call horiz_interp_del(Interp_new1)
         call horiz_interp_del(Interp_new2)
         call horiz_interp_del(Interp_cp)
-        ! test deletion after direct calls 
+        ! test deletion after direct calls
         ! this set up is usually done within horiz_interp_new
         nlon_in  = size(lon_in_1d(:))-1;  nlat_in  = size(lat_in_1d(:))-1
         nlon_out = size(lon_out_1d(:))-1; nlat_out = size(lat_out_1d(:))-1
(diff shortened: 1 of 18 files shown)


test_fms/horiz_interp/test_horiz_interp.F90
```diff
@@ -979,6 +984,10 @@ subroutine test_assignment()
         real(HI_TEST_KIND_) :: D2R = real(PI,HI_TEST_KIND_)/180._lkind
         real(HI_TEST_KIND_) :: R2D = 180._lkind/real(PI,HI_TEST_KIND_)
         real(HI_TEST_KIND_), parameter :: SMALL = 1.0e-10_lkind
+        !! for bicubic set up
+        real(HI_TEST_KIND_), allocatable :: lon_src_1d(:), lat_src_1d(:)
+        real(HI_TEST_KIND_), allocatable :: lon_dst_1d(:), lat_dst_1d(:)
```
thomas-robinson, line 989: These should have doxygen documentation instead of just a comment above

thomas-robinson (changes requested):

rem1776:
@thomas-robinson @bensonr This should be ready now, I updated the SPHERICA to SPHERICAL. I noticed that the interpolation method values were made public in the libFMS.F90 file so I also removed them from there as well since they should only be internal.

bensonr (approved):

thomas-robinson (approved):

PR #1540, opened 2024-06-03
Modern_diag_manger:: Use the correct date when prepending the date

### LINKED ISSUE #1539: Modern diag_manager:: Static files don't end up with the corrent prepend date
**Describe the bug**
I tried running a production run of AM5 with the new diag manager, and got an error in the PP at the second year. 

/home/Uriel.Ramirez/am5/am5f7b12r0_2024.01.01/c96L65_am5f7b12r0_pdclim2010F_modern/gfdl.ncrc5-intel23-classic-prod-openmp/stdout/postProcess/c96L65_am5f7b12r0_pdclim2010F_modern_aerosol_cmip_00020101.o37348926
```
foreach aff ( /xtmp/Uriel.Ramirez/job37348926/am5/am5f7b12r0_2024.01.01/c96L65_am5f7b12r0_pdclim2010F_modern/gfdl.ncrc5-intel23-classic-prod-openmp/c96L65_am5f7b12r0_pdclim2010F_modern_00020101/0002*/*00020101.grid_spec* )
foreach: No match.
```

The static files were not saved as `*00020101.grid_spec` they were actually `*00010101.grid_spec`. The non static files worked correctly. 

The code is using the base_date
https://github.com/NOAA-GFDL/FMS/blob/5757c7813f1170efd28f5a4206395534894095b4/diag_manager/fms_diag_file_object.F90#L262
https://github.com/NOAA-GFDL/FMS/blob/5757c7813f1170efd28f5a4206395534894095b4/diag_manager/fms_diag_file_object.F90#L1180-L1185

For non-static fields, diag_file%start_time is updated by what is called in the register calls. Since this doesn't happen for non-static fields the start_time is never updated so for multiple year runs it ends up being wrong.

**To Reproduce**
[This test reproduces the issue. ](https://github.com/uramirez8707/FMS/blob/Fix_prepend/test_fms/diag_manager/test_prepend_date.sh) The time passed into data_override is not the same as the base_date.
**Expected behavior**
The correct date should be prepended to the file name

**System Environment**
This happens in any system

**Additional context**
N/A

uramirez8707:
**Description**
Use the correct date when prepending the date

Fixes #1539 

**How Has This Been Tested?**
CI, including new tests

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_file_object.F90
+++ b/diag_manager/fms_diag_file_object.F90
@@ -264,8 +264,8 @@ logical function fms_diag_files_object_init (files_array)
      !! This will be the base_time if nothing was passed in
      !! This time is appended to the filename if the prepend_date namelist is .True.
      obj%start_time = diag_init_time
-     obj%last_output = get_base_time()
-     obj%model_time = get_base_time()
+     obj%last_output = diag_init_time
+     obj%model_time = diag_init_time
      obj%next_output = diag_time_inc(obj%start_time, obj%get_file_freq(), obj%get_file_frequnit())
      obj%next_next_output = diag_time_inc(obj%next_output, obj%get_file_freq(), obj%get_file_frequnit())
 
@@ -1009,18 +1009,18 @@ subroutine add_start_time(this, start_time)
   class(fmsDiagFile_type), intent(inout)       :: this           !< The file object
   TYPE(time_type),         intent(in)          :: start_time     !< Start time to add to the fileobj
 
-  !< If the start_time sent in is equal to the base_time return because
-  !! this%start_time was already set to the base_time
-  if (start_time .eq. get_base_time()) return
+  !< If the start_time sent in is equal to the diag_init_time return because
+  !! this%start_time was already set to the diag_init_time
+  if (start_time .eq. diag_init_time) return
 
-  if (this%start_time .ne. get_base_time()) then
-    !> If the this%start_time is not equal to the base_time from the diag_table
+  if (this%start_time .ne. diag_init_time) then
+    !> If the this%start_time is not equal to the diag_init_time from the diag_table
     !! this%start_time was already updated so make sure it is the same or error out
     if (this%start_time .ne. start_time)&
       call mpp_error(FATAL, "The variables associated with the file:"//this%get_file_fname()//" have"&
       &" different start_time")
   else
-    !> If the this%start_time is equal to the base_time,
+    !> If the this%start_time is equal to the diag_init_time,
     !! simply update it with the start_time and set up the *_output variables
     this%model_time = start_time
     this%start_time = start_time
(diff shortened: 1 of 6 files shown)


diag_manager/fms_diag_file_object.F90
```diff
@@ -1005,18 +1009,18 @@ subroutine add_start_time(this, start_time)
...
-  !! this%start_time was already set to the base_time
-  if (start_time .eq. get_base_time()) return
+  !< If the start_time sent in is equal to the diag_init_time return because
+  !! this%start_time was already set to the diag_init_time
+  if (start_time .eq. diag_init_time) return
 
-  if (this%start_time .ne. get_base_time()) then
-    !> If the this%start_time is not equal to the base_time from the diag_table
+  if (this%start_time .ne. diag_init_time) then
+    !> If the this%start_time is not equal to the diag_init_time from the diag_table
```
bensonr, line 1017: I only see a test where the date is prepended correctly.  Is there a way to force this scenario to ensure it is caught?

diag_manager/fms_diag_file_object.F90
```diff
@@ -1005,18 +1009,18 @@ subroutine add_start_time(this, start_time)
...
     !! this%start_time was already updated so make sure it is the same or error out
     if (this%start_time .ne. start_time)&
       call mpp_error(FATAL, "The variables associated with the file:"//this%get_file_fname()//" have"&
       &" different start_time")
   else
-    !> If the this%start_time is equal to the base_time,
+    !> If the this%start_time is equal to the diag_init_time,
```
bensonr, line 1023: Is the code below here needed?  Your comments at LL1012-1014 indicate it will never be executed because start_time=this%start_time and the early return at 1014 will prevail.

uramirez8707: Note that  the `start_time` is the time passed into the `register_diag_field` call. 

if the diag_time_init is not passed into diag_manager_init, `this%star_time` and `diag_init_time` will be set to the base date in the beginning of the run. 

In the second year of the simulation, `start_time` won't be equal to `diag_init_time`, so it will get passed L1012-1014.

 this%start_time will be equal to `diag_init_time`, so it will get to the else and this%start_time and the other variables will be set to start_time. This will only happen for the first variable in the file, after that `start_time` is expected to the same this%start_time or you will get to the FATAL.

diag_manager/fms_diag_file_object.F90
uramirez8707: The second test in [test_prepend_date.sh](https://github.com/NOAA-GFDL/FMS/pull/1540/files#diff-f98593e6e7ff4d938f937f53d8ed60454134e7fc8f3b04ff12845c3dedd5b664) should catch this.

PR #1566, opened 2024-08-02
Add `use_center_grid_points` namelist

### LINKED ISSUE #1537: Data_override:: Issue with reading the grid file in data_override when using 0:180:-180:0 grids
**Describe the bug**
Grid files that have longitude to ranges from 0 to 180 then -180 to 0 are not being read correctly by data_override when using "OCN" and "ICE" grids.

The code gets here: 
https://github.com/NOAA-GFDL/FMS/blob/460c5921249a2c872135bf3f37d57fe498bde93d/data_override/include/get_grid_version.inc#L218-L224

At the dateline, it will use values around -180 and 180 and average those which will cause bad longitudes.
You can see the issue in this file:
ncview /gpfs/f5/gfdl_f/scratch/Uriel.Ramirez/REGIONAL_MODEL/FMS/build/test_fms/data_override/LineOfWeirdness_Test /grid_read_in_ocn.nc

MOM6 seems to use the centroids when it reads its grid:
https://github.com/NOAA-GFDL/MOM6/blob/8ecb2603119b8c9d4dee9ee6b174e1864415d1cc/src/initialization/MOM_grid_initialize.F90#L220-L222

**To Reproduce**
This Test reproduces the error:
https://github.com/uramirez8707/FMS/blob/LineOfWeirdnessTest/test_fms/data_override/test_data_override_LOW.F90

image

The data files to run it can be found here:
/gpfs/f5/scratch/Uriel.Ramirez/gfdl_f/REGIONAL_MODEL/FMS/build/test_fms/data_override/LineOfWeirdness_Test

**Expected behavior**
The grid file should be read correctly by data override. We should either:
1. Read the centroids from the file directly
2. Add logic to the averaging to deal with this corner case
3. Model components can pass in their own lat/lon to use by data_override

**System Environment**
This happens in any system

**Additional context**
This was seen by @theresa-morrison and @kshedstrom

uramirez8707:
**Description**
Adds a new namelist, `use_center_grid_points`, which indicates whether or not to use the centroid values of the supergrid from the grid file as opposed to calculating it by taking the average of the four corner points. This is only relevant to OCN and ICE grids.

The namelist is needed because answers will not bit for bit reproduce, but overall the answers should not be very different. 

The namelist should be set to .True.

Fixes #1537 

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes


bensonr (approved):
lgtm

rem1776 (approved):

Issue #1564, opened 2024-08-01
remove path length parameter from field_manager


rem1776:
**Is your feature request related to a problem? Please describe.**
There is a a variable for file/path name length `fm_path_name_len` set in the field manager.  This should be removed when possible so we can use consistent lengths across the code instead of defining different lengths in different areas.

**Describe the solution you'd like**
Once a consistent path length is added to FMS, we should remove usage of the field_manager-specific variable (used in `land_lad2` and `ocean_BGC`), and then remove it from field_manager.

**Describe alternatives you've considered**
Keeping the variable as is, it will now be set to a parameter set in platform_mod.

**Additional context**
came up while working on #1518

rem1776:
going to re-open this until we get the components updated just so we have somewhere to track it

We need these two prs merged for the next tag as well:

https://gitlab.gfdl.noaa.gov/fms/land_lad2/-/merge_requests/46
https://github.com/NOAA-GFDL/ocean_BGC/pull/40

uramirez8707:
> going to re-open this until we get the components updated just so we have somewhere to track it
> 
> We need these two prs merged for the next tag as well:
> 
> https://gitlab.gfdl.noaa.gov/fms/land_lad2/-/merge_requests/46 [NOAA-GFDL/ocean_BGC#40](https://github.com/NOAA-GFDL/ocean_BGC/pull/40)

Those two tags have been created, so closing this issue.

PR #1567, opened 2024-08-02
add consistent character lengths for files and paths

### LINKED ISSUE #1518: Consistent maximum length of filenames?
There is code in fms that still uses string lengths of 128 characters for filenames. This is causing problems with models such as GEOS, which uses deep paths (see e.g. https://github.com/GEOS-ESM/swell/issues/340).

I looked around in the fms code and found that there is a "wide range" of character limits for filenames. I also found that several places are using
```
./mosaic2/mosaic2.F90:     MAX_FILE = 1024, & !> max length of the file names
./mosaic2/mosaic2.F90:    character(len=MAX_FILE) :: gridfile
```
which coincidentally fits with @mathomp4's suggestion to use 1024 characters (aka ESMF_MAXPATHLEN). See https://github.com/GEOS-ESM/swell/issues/340#issuecomment-2104793923.

It would be really nice if all file names in fms - unless an edge case requires shorter paths - were using `MAX_FILE`.

Manually bump up character limits for filenames in several places (not a good idea).

**Additional context**
n/a

rem1776:
**Description**
- Adds two parameter values to platform_mod, `FMS_PATH_LEN` and `FMS_FILE_LEN` that are set by the `FMS_MAX_PATH_LEN` and `FMS_MAX_FILE_LEN` cpp macros in the fms_platform.h file and default to 1024 and 255 respectively.
- Replaces/removes any lengths for file/path names with the exception of `fm_path_name_len`. Since it is used externally, I removed internal usage and set the value to FMS_PATH_LEN so we can remove it later along with updating the components, see issue #1564.
- Uses `len=*` instead of a set length for any filename related character arguments

It can be somewhat ambiguous what is intended as a filename vs a full path name, so let me know if you think a length should be one instead of the other or if anything was missed.

Fixes #1518 

**How Has This Been Tested?**
gnu 13 + mpich on the amd box

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -1348,7 +1348,8 @@ end function rarray_to_char
     integer, dimension(2) :: get_ascii_file_num_lines_and_length !< number of lines (1) and
                                                                  !! max line length (2)
     integer :: num_lines, max_length
-    character(len=FMS_FILE_LEN) :: str_tmp
+    integer, parameter :: LENGTH=1024
+    character(len=LENGTH) :: str_tmp
     character(len=5) :: text
     integer :: status, f_unit, from_pe
     logical :: file_exist
@@ -1388,10 +1389,10 @@ end function rarray_to_char
                    call mpp_error(FATAL, 'get_ascii_file_num_lines: Error reading line '//trim(text)// &
                         ' in file '//trim(FILENAME)//'.')
                 end if
-                if ( len_trim(str_tmp) == FMS_FILE_LEN ) then
-                   write(UNIT=text, FMT='(I5)') FMS_FILE_LEN
+                if ( len_trim(str_tmp) == LENGTH) then
+                   write(UNIT=text, FMT='(I5)') LENGTH 
                    call mpp_error(FATAL, 'get_ascii_file_num_lines: Length of output string ('//trim(text)//&
-                                       & ' is too small. Increase the FMS_MAX_FILE_LEN cpp value.')
+                                       & ' is too small. Increase the LENGTH value.')
                 end if
                 if (len_trim(str_tmp) > max_length) max_length = len_trim(str_tmp)
                 num_lines = num_lines + 1
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -1390,7 +1390,7 @@ end function rarray_to_char
                         ' in file '//trim(FILENAME)//'.')
                 end if
                 if ( len_trim(str_tmp) == LENGTH) then
-                   write(UNIT=text, FMT='(I5)') LENGTH 
+                   write(UNIT=text, FMT='(I5)') LENGTH
                    call mpp_error(FATAL, 'get_ascii_file_num_lines: Length of output string ('//trim(text)//&
                                        & ' is too small. Increase the LENGTH value.')
                 end if
--- a/mpp/include/mpp_util.inc
+++ b/mpp/include/mpp_util.inc
@@ -1348,7 +1348,6 @@ end function rarray_to_char
     integer, dimension(2) :: get_ascii_file_num_lines_and_length !< number of lines (1) and
                                                                  !! max line length (2)
     integer :: num_lines, max_length
-    integer, parameter :: LENGTH=FMS_FILE_LEN
     character(len=FMS_FILE_LEN) :: str_tmp
     character(len=5) :: text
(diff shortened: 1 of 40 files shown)


thomas-robinson (dismissed):
I looked at the platform files and that looked alright, and very briefly skimmed the rest.  Will defer to @bensonr because I am not going to be available for a more thorough review.

thomas-robinson:
@climbfuji @mathomp4 @Dooruk Does this work sufficiently fix your issue?

mpp/include/mpp_util.inc
```diff
@@ -1348,8 +1348,7 @@ end function rarray_to_char
     integer, dimension(2) :: get_ascii_file_num_lines_and_length !< number of lines (1) and
                                                                  !! max line length (2)
     integer :: num_lines, max_length
-    integer, parameter :: LENGTH=1024
-    character(len=LENGTH) :: str_tmp
+    character(len=FMS_FILE_LEN) :: str_tmp
```
bensonr, line 1351: This is not a filename, but the length of a string being read from an ascii file for conversion to a character variable.  Using FMS_FILE_LEN would now impose a limit of 255 from the original 1024.

mpp/include/mpp_util.inc
```diff
@@ -1389,10 +1388,10 @@ end function rarray_to_char
                    call mpp_error(FATAL, 'get_ascii_file_num_lines: Error reading line '//trim(text)// &
                         ' in file '//trim(FILENAME)//'.')
                 end if
-                if ( len_trim(str_tmp) == LENGTH ) then
-                   write(UNIT=text, FMT='(I5)') length
+                if ( len_trim(str_tmp) == FMS_FILE_LEN ) then
+                   write(UNIT=text, FMT='(I5)') FMS_FILE_LEN
```
bensonr, line 1392: I think we want to revert this to the original code here as the LENGTH here is not referring to a file length or pathname.

mpp/include/mpp_util.inc
```diff
@@ -1389,10 +1388,10 @@ end function rarray_to_char
...
                    call mpp_error(FATAL, 'get_ascii_file_num_lines: Length of output string ('//trim(text)//&
-                                       & ' is too small. Increase the LENGTH value.')
+                                       & ' is too small. Increase the FMS_MAX_FILE_LEN cpp value.')
```
bensonr, line 1394: I think we want to revert this as well.

time_interp/time_interp_external2.F90
```diff
@@ -126,7 +126,7 @@ module time_interp_external2_mod
     !> Holds filename and file object
     !> @ingroup time_interp_external2_mod
     type, private :: filetype
-        character(len=128) :: filename = ''
+        character(len=FMS_PATH_LEN) :: filename = ''
```
bensonr, line 129: This is a filename, shouldn't it be limited using FMS_FILE_LEN?  If so, need to change the platform module use statement as well.

bensonr (changes requested):
Are there any tests codes that would need changed as well?

mpp/include/mpp_util.inc
rem1776: Fixed with f9cac03

time_interp/time_interp_external2.F90
rem1776: Yeah I think this one should be a file length, fixed with f9cac03

bensonr:
> > Are there any tests codes that would need changed as well?
> 
> @bensonr I skipped the tests because most file/path names used in the tests are typically set once and not modified so there's not really a chance of one being too short. There's a couple cases where a filename can be set by a namelist, but then the values are just set in the test scripts so it would always be set to a value we control.
> 
> Do you think they should be updated as well?

@rem1776 - hopefully we'll find out during testing or via the automated ci/cd

mpp/include/mpp_util.inc
```diff
@@ -1228,7 +1228,7 @@ end function rarray_to_char
     integer, dimension(2) :: lines_and_length
     logical :: file_exist
     character(len=len(peset(current_peset_num)%name)) :: pelist_name
-    character(len=128) :: filename
+    character(len=FMS_PATH_LEN) :: filename
```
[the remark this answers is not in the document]
rem1776: The filename can be set by the `alt_input_path_name` optional variable so this should be a full path length.

bensonr (approved):

PR #1588, opened 2024-09-09
Make define_blocks warning a note

### LINKED ISSUE #1570: Make define_blocks warning a note
**Describe the bug**
Whenever you don't have even blocks, you will get this warning

```
"WARNING from PE     0: atmos_modeldefine_blocks: domain (  16  12) is not an even divisor with definition (   1   8) - blocks will not be uniform
```

Because this is warning, every PE outputs that warning, which causes the stdout to be very lengthy

**To Reproduce**
Use an uneven block size and you will get that warning

**Expected behavior**
This should be a NOTE, so that only the root pe outputs that warning

**System Environment**
This happens in any system

**Additional context**
N/A

uramirez8707:
**Description**
Update so that only the root pe outouts the non uniform block size warning instead of every PE, which will make the stdout very large.

Fixes #1570 

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/block_control/block_control.F90
+++ b/block_control/block_control.F90
@@ -23,8 +23,9 @@
 
 module block_control_mod
 
-use mpp_mod,         only: mpp_error, NOTE, WARNING, FATAL
+use mpp_mod,         only: mpp_error, NOTE, WARNING, FATAL, mpp_sum, mpp_npes
 use mpp_domains_mod, only: mpp_compute_extent
+use fms_string_utils_mod, only: string
 implicit none
 
 public block_control_type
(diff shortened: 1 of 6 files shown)


bensonr (dismissed):
Nice update.

rem1776:
It looks like this is hitting some kind of deadlock in the CI and is hanging during the openmp test for the new diag_manager. It only happens for the builds that have openmp turned off though.

https://github.com/NOAA-GFDL/FMS/actions/runs/10774360061/job/29934973086?pr=1588#step:7:6253

You should be able to reproduce it if you configure with `--disable-openmp`.

rem1776 (approved):

bensonr (approved):

PR #1589, opened 2024-09-10
Use $CMAKE_INSTALL_LIBDIR instead of hardcoding "lib"

harshula:
Since `GNUInstallDirs` (https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) is already included, we could use `$CMAKE_INSTALL_LIBDIR` instead of hardcoding `lib`.

During my testing `$CMAKE_INSTALL_LIBDIR` expanded to `lib64`, not `lib`:
```
$ ls /opt/release/linux-rocky8-x86_64/intel-2021.2.0/fms-git.use-cmake-install-libdir_0-git.3289-o47ywsb444ovhjfysqu2jqgzbe7szfrk/
include_r4  lib64
```

Is `lib` hardcoded to avoid the library being installed in `lib64`? If not, please consider this patch.


climbfuji (approved):
Yes please, thanks!

bensonr (approved):

rem1776 (approved):

PR #1597, opened 2024-10-09
Fix `mosaic2` tests with strict debug flags

J-Lentz:
**Description**
This fixes two issues that occur in the `mosaic2` tests when FMS is built using strict debug flags:

1) The `area` argument of `calc_mosaic_grid_area` has been changed from `intent(inout)` to `intent(out)`, to avoid an uninitialized value error.

2) `atan` has been changed to `atan2` in the unit tests, to avoid divide-by-zero errors.

**How Has This Been Tested?**
Fixes the `mosaic2` unit tests when FMS is built with ifort 2021.13.0 on the AMD box with the following debug flags:
`-check all -fpe0 -fp-stack-check -fstack-security-check -ftrapuv -init=arrays,minus_huge,snan`

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/mosaic2/include/mosaic2.inc
+++ b/mosaic2/include/mosaic2.inc
@@ -86,14 +86,12 @@
   !> <br>Example usage:
   !!            call calc_mosaic_grid_area(lon, lat, area)
   subroutine CALC_MOSAIC_GRID_AREA_(lon, lat, area)
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)    :: lon
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)    :: lat
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(inout) :: area
+     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)  :: lon
+     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)  :: lat
+     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(out) :: area
      integer :: nlon, nlat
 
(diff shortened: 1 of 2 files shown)


laurenchilutti:
@abrooks1085 Can you try this branch with ifx prod flags? See if it fixes the unit test crashes in mosaic2?

mosaic2/include/mosaic2.inc
```diff
@@ -86,15 +86,13 @@
   !> <br>Example usage:
   !!            call calc_mosaic_grid_area(lon, lat, area)
   subroutine CALC_MOSAIC_GRID_AREA_(lon, lat, area)
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)    :: lon
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(in)    :: lat
-     real(kind=FMS_MOS_KIND_), dimension(:,:), intent(inout) :: area
```
laurenchilutti, line 91: I am asking purely for my own learning and do not know what is best: Why would you decide to change the intent instead of just initializing the output variable in the code before this subroutine is called?

J-Lentz: `CALC_MOSAIC_GRID_AREA_` is a wrapper around `get_grid_area`, which sets the area but doesn't read the value that's passed in. So `intent(out)` more clearly communicates to the user what the code is doing.

J-Lentz: Or in other words, there's no reason it should need to be initialized, because it gets set by `get_grid_area`.

rem1776 (approved):

uramirez8707 (approved):

PR #1606, opened 2024-10-24
Fix test failures with ifx due to use of host associated arrays

abrooks1085:
**Description**
Adds save attribute to host associated arrays in test_topography and test_diag_integral so that these tests now pass with the ifx compiler.

**How Has This Been Tested?**
The Topography and Diag_integral unit tests successfully build and pass using the Intel oneAPI 2024.2 IFX compiler on an AMD box.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/test_fms/diag_integral/test_diag_integral.F90
+++ b/test_fms/diag_integral/test_diag_integral.F90
@@ -97,7 +97,7 @@ program test_diag_integral
 
 contains
   !-------------------------------------
-  !------------------------------------- 
+  !-------------------------------------
   subroutine initialize_arrays
 
     !> made up numbers
--- a/test_fms/diag_integral/test_diag_integral.F90
+++ b/test_fms/diag_integral/test_diag_integral.F90
@@ -51,14 +51,14 @@ program test_diag_integral
   character(9), parameter :: field_nameh='immadeuph' !> made up field name to test
   character(8), parameter :: std_digits   = 'e23.15e3' !> write out precision for r8_kind data
 
-  real(TEST_DI_KIND_), save :: immadeup2(nxy,nxy)     !> array to test sum_field_2d
-  real(TEST_DI_KIND_), save :: immadeup3(nxy,nxy,nxy) !> array to test sum_field_3d
-  real(TEST_DI_KIND_), save :: immadeupw(nxy,nxy,nxy) !> array to test sum_field_wght_3d
-  real(TEST_DI_KIND_), save :: weight(nxy,nxy,nxy)    !> weights required to test sum_field_wght_3d
-  real(TEST_DI_KIND_), save :: immadeuph(nxy,nxy)     !> array to test sum_field_2d_hemi
+  real(TEST_DI_KIND_) :: immadeup2(nxy,nxy)     !> array to test sum_field_2d
+  real(TEST_DI_KIND_) :: immadeup3(nxy,nxy,nxy) !> array to test sum_field_3d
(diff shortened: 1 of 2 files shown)


laurenchilutti (approved):

test_fms/diag_integral/test_diag_integral.F90
```diff
@@ -175,12 +175,13 @@ end subroutine test_sum_diag_integral_field
...
-    character(*), parameter :: di_file='diag_integral.out'
-    integer :: iunit
+    character(17), parameter :: di_file='diag_integral.out'
+    integer, parameter  :: iunit=100
+
     character(100) :: cline1, cline2, cline3, cline4, cline5, clin6
 
     !> read in computed values
-    open(newunit=iunit, file=di_file)
+    open(unit=iunit,file=trim(di_file))
```
rem1776, line 184: This should be changed back to using newunit rather than a hardcoded unit number. It's typically a better practice and I think this would also break the test with cray since it reserves 100 for stdin.

uramirez8707 (dismissed):

rem1776 (changes requested):
Just curious, what error is being fixed by adding the `save` attributes?

test_fms/diag_integral/test_diag_integral.F90
laurenchilutti: Andrew just needs to merge in the main branch - he made these changes before Jesse's newunit fix got merged in so now it shows as Andrew changing it back

abrooks1085: Yes, Lauren is correct. I'll fix it

bensonr:
Same question as @rem1776.

laurenchilutti:
Can you rewrite this test to pass the arrays into the subroutines rather than keep them as program level variables? From our group discussions, this seems to be agreed upon as a path forward with this unit test.

abrooks1085 (commented):
Updated my branch and removed the save attribute in favor of passing arrays to subroutine directly

uramirez8707 (approved):

test_fms/diag_integral/test_diag_integral.F90
```diff
@@ -96,6 +96,24 @@ program test_diag_integral
...
+    lon=1.0_lkind
+    lat=1.0_lkind
+    area=1.0_lkind
+    immadeup2=1.0_lkind
+    immadeup3=1.0_lkind
+    immadeupw=1.0_lkind
+    immadeuph=1.0_lkind
+    weight=1.0_lkind
+
+  end subroutine initialize_arrays
```
J-Lentz, line 116: Why not get rid of `initialize_arrays` and initialize these in the declaration statements at the top of the program instead? As a bonus, that would give them an implicit `save` attribute.

test_fms/diag_integral/test_diag_integral.F90
```diff
@@ -96,6 +96,24 @@ program test_diag_integral
   call test_sum_diag_integral_field !< compare read in values to the expected values.
 
 contains
+  !-------------------------------------
```
rem1776, line 99: Should this test be updated like the other one to pass in the variables to each routine?

The only real change I'm seeing in this test is moving the `intialize_arrays` subroutine up in the file, which shouldn't change anything technically, but maybe ifx thinks otherwise.

rem1776 (approved):

PR #1610, opened 2024-10-30
Set `fms_diag_accept_data`

### LINKED ISSUE #1591: Modern Diag Manager crashes in debug mode because FMS_DIAG_ACCEPT_DATA is never set
**Describe the bug**
When running with the modern diag manager, ESM is debug mode is crashing with
```
forrtl: severe (194): Run-Time Check Failure. The variable 'fms_diag_object_mod_mp_fms_diag_accept_data_$FMS_DIAG_ACCEPT_DATA' is being used in '/gpfs/f5/gfdl_f/scratch/Niki.Zadeh/fre/FMS2024.03_mom6_20231130/esm4.2_compile/src/FMS/diag_manager/fms_diag_object.F90(718,1)' without being defined
Image              PC                Routine            Line        Source                                              
fms_esm4.2_compil  000000000E5F3707  diag_manager_mod_        1792  diag_manager.F90                                    
fms_esm4.2_compil  000000000E5E1A5E  diag_manager_mod_        1578  diag_manager.F90  
```

In this case the code leaves [here](https://github.com/NOAA-GFDL/FMS/blob/44a211a7540576db0237a9e7d2edcedfbb2eb155/diag_manager/fms_diag_object.F90#L712), so `fms_diag_accept_data` is never set.

**To Reproduce**
Run in debug mode (without openmp)

**Expected behavior**
Debug mode should work

**System Environment**
This happens in any system. 

**Additional context**
N/A

uramirez8707:
**Description**
1. Ensure that `fms_diag_accept_data` is always set by removing return statements and setting it in the end. This is to prevent errors in debug mode in cases where `fms_diag_accept_data` is not set.  `fms_diag_accept_data` is always set to .true., if there any errors in the subroutine it will crash with a FATAL.
2. Fixes the calls to send_data for a static variable in the unit tests. 

Fixes #1591 

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_manager.F90
+++ b/diag_manager/diag_manager.F90
@@ -1789,7 +1789,7 @@ END FUNCTION register_static_field
       allocate(mask_remap(1:size(mask,1), 1:size(mask,2), 1:size(mask,3), 1))
       mask_remap(:,:,:,1) = mask
     endif
-    diag_send_data = fms_diag_object%fms_diag_accept_data(diag_field_id, field_remap, mask_remap, rmask_remap, &
+    call fms_diag_object%fms_diag_accept_data(diag_field_id, field_remap, mask_remap, rmask_remap, &
                                                           time, is_in, js_in, ks_in, ie_in, je_in, ke_in, weight, &
                                                           err_msg)
     deallocate (field_remap)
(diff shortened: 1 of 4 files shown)


rem1776 (commented):
If this will always return true, should we just make this a subroutine instead?  I think there's only like two calls to it.

rem1776 (approved):
thanks for making that change, looks good.

PR #1634, opened 2025-01-08
1631: Improve modern diag manager performance

### LINKED ISSUE #1631: Modern diag manager: Improve performance
**Describe the bug**
When comparing the performance between the modern diag manager and the legacy diag manager, you can see that the performance of the modern diag manager is consistently 4-7% slower than the legacy diag manager

**To Reproduce**
Compare performance between the modern and the legacy diag manager

One way to improve performance is to avoid x**power (where power is equal to 1)
https://github.com/uramirez8707/FMS/blob/bd9cbebd9efd1f94207a141cf43ee65cd0d041cd/diag_manager/include/fms_diag_reduction_methods.inc#L323-L325

Another way is to avoid having diag_send_complete loop through all the variable to check if it is time to do the buffer math/write output if send_data has not been called since the last time diag_send complete call.

Another way is to explore using do concurrent loops to parallelize the do loops when doing the math

**Expected behavior**
The performance should be similar or better

**System Environment**
This will happen in any system

**Additional context**
N/A

uramirez8707:
**Description**
1.  Improves modern diag manager performance by avoiding doing `x**power` (where power is equal to 1)
2. Adds a flag to prevents `diag_send_complete` from doing unnecessary worked when there isn't any new data to process

Fixes #1631 

**How Has This Been Tested?**
CI
AM5 c96L65 2010 climo run

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/fms_diag_object.F90
+++ b/diag_manager/fms_diag_object.F90
@@ -61,6 +61,8 @@ private
   type(fmsDiagField_type), allocatable :: FMS_diag_fields(:) !< Array of diag fields
   type(fmsDiagOutputBuffer_type), allocatable :: FMS_diag_output_buffers(:) !< array of output buffer objects
                                                                        !! one for each variable in the diag_table.yaml
+  logical, private :: data_was_send !< True if send_data has been successfully called for at least one variable
+                                    !< diag_send_complete does nothing if it is .false.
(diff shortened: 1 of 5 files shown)


rem1776 (approved):
LGTM!

PR #1647, opened 2025-02-19
Add a test for collective netcdf mpi io

uramirez8707:
**Description**
Add a test for collective netcdf mpi io

The test creates a 1 undistributed file io_layout of 1,1 using domain decomposed fms2_io then reads the file using collective netcdf (currently collective netcdf only works if the layout and the io_layout are the same when writing)

Fixes # (issue)

**How Has This Been Tested?**
CI

**Checklist:**
- [x] `make distcheck` passes


bensonr (approved):

rem1776 (changes requested):
Looks good, i think that `m4/check_parallel4.c` file can probably be removed though. It should be autogenerated during configure so might have been added accidently.

uramirez8707:
> Looks good, i think that `m4/check_parallel4.c` file can probably be removed though. It should be autogenerated during configure so might have been added accidently.

That is not auto generated. It is used for cmake builds.
https://github.com/NOAA-GFDL/FMS/blob/6fea71ec3bf5c62edcb4f423ada199bd8b00f4ea/CMakeLists.txt#L292-L294

I wasn't sure about the best way to do this, so let me know if you have a better idea

rem1776:
The cmake netcdf module should have a check for that already, I think we just weren't using it before:

https://github.com/NOAA-GFDL/FMS/blob/eef4a897e6d783f61e2eef378537fc6d675d2194/cmake/FindNetCDF.cmake#L12-L18

We should be able to use that in the conditional instead

PR #1733, opened 2025-07-14
fms2_io: Add collective writes

J-Lentz:
**Description**
Adds parallel writes to fms2_io for 2D, 3D, 4D, and 5D cases with a new unit test.

**How Has This Been Tested?**
Builds with ifx 2025.2.0 on the AMD box and the test passes.

**Checklist:**
- [x] `make distcheck` passes


fms2_io/include/domain_write.inc
```diff
@@ -1258,4 +1279,575 @@ subroutine domain_write_5d(fileobj, variable_name, vdata, unlim_dim_level, &
...
+end subroutine netcdf_mpi_write_2d
+
+subroutine netcdf_mpi_write_3d(fileobj, variable_name, vdata, unlim_dim_level, corner, edge_lengths)
+  type(FmsNetcdfDomainFile_t), intent(in) :: fileobj !< File object.
+  character(len=*), intent(in) :: variable_name !< Variable name.
+  class(*), dimension(:,:,:), intent(in) :: vdata !< Data that will be written out to the netcdf file.
+  integer, intent(in), optional :: unlim_dim_level !< Level for the unlimited dimension.
+  integer, dimension(3), intent(in), optional :: corner !< Array of starting indices describing where the data
+                                                        !! will be written to.
+  integer, dimension(3), intent(in), optional :: edge_lengths !< The number of elements that will be written
```
uramirez8707, line 1398: These should be `dimension(4)`, for x,y,z and the unlimited dimension.

Similar changes are needed for the other dimensions.

fms2_io/include/domain_write.inc
```diff
@@ -1258,4 +1279,575 @@ subroutine domain_write_5d(fileobj, variable_name, vdata, unlim_dim_level, &
...
+      endif
+      i1 = i0 + yc_size - 1
+      j1 = j0 + xc_size - 1
+  end select
+
+  varid = get_variable_id(fileobj%ncid, trim(variable_name), &
+    msg="file:"//trim(fileobj%path)//" and variable:"//trim(variable_name))
+
+  err= nf90_var_par_access(fileobj%ncid, varid, nf90_collective)
+  call check_netcdf_code(err, "Failed to set the variable '"//trim(variable_name)//"' to collective-write mode")
```
uramirez8707: Instead of doing this here, let's remove this and do this at the end of `netcdf_add_variable`
```   if (fileobj%use_netcdf_mpi) then
    err= nf90_var_par_access(fileobj%ncid, varid, nf90_collective)
    call check_netcdf_code(err, append_error_msg)
  endif 
```

uramirez8707:
As we discuss on Monday, if using collective mpi netcdf the code is going to put out a NOTE stating that the io_domain will be ignored and only write out 1 file.

These are some other changes that are needed:
https://github.com/uramirez8707/FMS/commit/6999342538a7ec67698db830620e88ba07340b51 This is so that the dimensions are registered with the correct size (the compute domain rather the io compute domain)

https://github.com/uramirez8707/FMS/commit/5908a7220aadc33d43f3f2659d05b0642f3b6c81 This is so that the code doesn't write the `domain decomposition` attribute since it is not needed since there is no combining

fms2_io/include/domain_write.inc
```diff
@@ -1258,4 +1279,575 @@ subroutine domain_write_5d(fileobj, variable_name, vdata, unlim_dim_level, &
...
+    return
+  endif
+
+  !! Get some more info about the variable
+  call domain_offsets(size(vdata, xdim_index), size(vdata, ydim_index), fileobj%domain, &
+                      xpos, ypos, isd, isc, xc_size, jsd, jsc, yc_size, &
+                      buffer_includes_halos, msg="file:"//trim(fileobj%path)//" and variable:"//trim(variable_name))
+
+  c = 1
+  e = 1
```
uramirez8707: Rather than initializing `e` to 1, do
```
  e(1:2) = shape(vdata)
```

And for the 3D case,
```
e(1:2) = shape(vdata)
e(3) = 1 ! unlimited dimension
```

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -319,11 +319,123 @@ function is_dimension_registered(fileobj, dimension_name) &
...
+  endif
+  call check_netcdf_code(err, "open_collective_netcdf_file:"//trim(fileobj%path))
+
+  allocate(fileobj%xdims(max_num_domain_decomposed_dims))
+  fileobj%nx = 0
+  allocate(fileobj%ydims(max_num_domain_decomposed_dims))
+  fileobj%ny = 0
+
+  allocate(fileobj%pelist(mpp_get_domain_npes(domain)))
+  call mpp_get_pelist(domain, fileobj%pelist)
```
uramirez8707: As we discussed on monday, let's do this instead
```
  ! Every rank is in its own pelist.
  ! This forces all ranks to hit any NetCDF calls,
  ! which are usually inside `if (fileobj%is_root)` blocks, leading to hangs when using collective netcdf
  allocate(fileobj%pelist(1))
  fileobj%pelist(1) = mpp_pe()
  fileobj%io_root = mpp_pe()
  fileobj%is_root = .True.
```

and remove the if `fileobj%use_netcdf_mpi` checks

fms2_io/include/domain_write.inc
bensonr: Is there a requirement for the last index to be the unlimited one?  Will this have any effects on future efforts to implement generalized indices?

fms2_io/include/domain_write.inc
J-Lentz: I'm assuming you mean `c` and `e` should be one dimension higher, not the optional `corner` and `edge_lengths` arguments - is that correct?

fms2_io/include/domain_write.inc
J-Lentz: Done

fms2_io/include/domain_write.inc
J-Lentz: I think we've been assuming so far in the generalized indices work (at least, I did in the test that I worked on) that the unlimited dimension would always be last.

It's not clear to me if this is a hard requirement. It seems like it might not be with NetCDF 4, but the NetCDF documentation also says (https://docs.unidata.ucar.edu/netcdf-c/4.9.3/unlimited_dims.html):

> For programmers, the unlimited dimension will correspond with the slowest-varying dimension. In C this is the first dimension of an array, in Fortran, the last.

fms2_io/fms_netcdf_domain_io.F90
J-Lentz: Done

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -319,11 +319,125 @@ function is_dimension_registered(fileobj, dimension_name) &
 
 end function is_dimension_registered
 
+function open_collective_netcdf_file(fileobj, path, mode, domain, is_restart, dont_add_res_to_filename) &
```
rem1776, line 324: Could you just throw a brief description above this function for the docs

test_fms/fms2_io/test_collective_io.sh
```diff
@@ -27,7 +27,7 @@
 . ../test-lib.sh
 
 if [ ! -z $parallel_skip ]; then
-  SKIP_TESTS="test_collective_io.1"
+  SKIP_TESTS="test_collective_io.1 test_collective_io.3"
```
rem1776: we should probably be skipping all the tests right?
```suggestion
  SKIP_TESTS="test_collective_io.[1-3]"
```

J-Lentz: I believe the second test runs with collective reads disabled, so I think it should work without parallel HDF5.

J-Lentz: I've pushed an update with improved descriptions of the tests.

fms2_io/fms_netcdf_domain_io.F90
J-Lentz: Done

test_fms/fms2_io/test_collective_io.sh
rem1776: Thanks! yeah i think thats helpful

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -458,7 +595,11 @@ subroutine register_domain_decomposed_dimension(fileobj, dim_name, xory, domain_
   if (mpp_domain_is_symmetry(fileobj%domain) .and. present(domain_position)) then
     dpos = domain_position
   endif
-  io_domain => mpp_get_io_domain(fileobj%domain)
+  if (fileobj%use_netcdf_mpi) then
+    io_domain => fileobj%domain
```
uramirez8707, line 602: Add a note here, something like

"If using mpi netcdf, the io domain is ignored so use the domain to determine the correct size of each domain decomposed dimension"

fms2_io/fms_netcdf_domain_io.F90
```diff
@@ -319,11 +319,127 @@ function is_dimension_registered(fileobj, dimension_name) &
...
+  character(len=FMS_PATH_LEN) :: full_path
+  logical :: is_res
+  logical :: dont_add_res
+  integer :: success
+  integer :: domain_size
+
+  success = .true.
+  call string_copy(fileobj%non_mangled_path, path)
+
+  !TODO Lots of duplicate code between this and netcdf_file_open
```
uramirez8707, line 352: Please create a issue for this so we don't forget about it

fms2_io/include/domain_write.inc
```diff
@@ -1258,4 +1279,563 @@ subroutine domain_write_5d(fileobj, variable_name, vdata, unlim_dim_level, &
     end select
   endif
 end subroutine domain_write_5d
+
+subroutine netcdf_mpi_write_2d(fileobj, variable_name, vdata, unlim_dim_level, corner, edge_lengths)
```
uramirez8707, line 1284: Please add doxygen documentation for these new routines.

J-Lentz: Done

fms2_io/fms_netcdf_domain_io.F90
J-Lentz: Done

fms2_io/fms_netcdf_domain_io.F90
J-Lentz: #1753

uramirez8707 (approved):

vithikashah001 (approved):

PR #1756, opened 2025-08-25
Collective IO in modern diag manager

### LINKED ISSUE #1716: MODERN DIAG MANAGER: Implement fms2io netcdf collective mpi writes to diag_manager
Because of the most IO writes happen in the diag manager, there should be a way to control whether or not the collective writes implemented in #1715 are used and because chunking makes a big difference in performance, there should be a way to control how the chunksizes are defined. 

Implement fms2io netcdf collective mpi writes to diag manager with a way to control whether or not it used

Implement a way to control the chunksizes when using netcdf collective mpi writes

N/A

**Additional context**
N/A

uramirez8707:
**Description**
Implements collective io in modern diag manager. This option is enabled by setting the key `use_collective_writes` for each file.  There is also an option to define the chunksizes at a file level for all the variables in a file, and a way to override it at the field level. 

Requires changes to the schema, but the schema changes are backwards compatible since the schema will still work for old files. 
Fixes #1716

**How Has This Been Tested?**
CI, including new tests. 
Ran a "successful" test with this option turned on using AM5.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_yaml_format.md
+++ b/diag_manager/diag_yaml_format.md
@@ -582,7 +582,7 @@ diag_files:
 ```
 
 **Things to consider:**
-1.  This option currently only works for files with variables that are domain-decomposed (i.e., each processor has its own section of the data). Support for land's unstructured grid and icebergs will be added in the future; at present, using this option with those grids will result in a FATAL error.
-2.  The domain must be evenly decomposed, meaning each processor has the same amount of data. For example, a C3072L65 grid with a layout of (30, 32) will not work and will result in a FATAL error, since 3072 is not evenly divisible by 30.
+1.  This option currently only works for files with variables that are domain-decomposed (i.e., multiple mpi-tasks each with their own section of data). Support for land's unstructured grid and icebergs will be added in the future; at present, using this option with those grids will result in a FATAL error.
+2.  The domain must be evenly decomposed, meaning each processor has the same amount of data. For example, a C3072L65 grid with a layout of (30, 32) will not work and will result in a FATAL error, since 3072 is not evenly divisible by 30. This limitation exists because NetCDF collective I/O requires each process to access the same pattern of chunks. With uneven decomposition, the chunk shapes per process vary, making it impossible to determine a consistent chunking strategy. Currently there are no plans to support this in the future.
 3. The chunksizes must be a comma seperated list of five integers 1 for each dimension.
 4. Chunking is supported even when not using collective writes as long as netcdf-4 file are being used (by setting fms2_io_nml::netcdf_default_format="netcdf4)


diag_manager/diag_yaml_format.md
```diff
@@ -524,4 +525,63 @@ diag_files:
...
+  unlimdim: time
+  module: ocn_mod
+  reduction: none
+  kind: r4
+  use_collective_writes: True
+  chunksizes: 96, 8, 1, 1, 1
+  varlist:
+  - var_name: var4
+```
+Chunking can also have a major impact on performance. You can set chunk sizes for all variables in a file using the `chunksizes` key at the file level. By default, the diag manager will choose chunk sizes equal to the compute domain for domain-decomposed dimensions (e.g., for C3072L65 and a layout of 32,64, the chunksizes would be 96,48). For non-domain-decomposed dimensions, the chunk size is set to the dimension length. For unlimited dimensions, the chunk size is always 1. For example:
```
bensonr: Just to be clear, there is no chunking for the K-dim?  If not, that should be specified in the documentation and the  example chunksize would be 96,48,64 (65) depending on layer (level) and assuming we are explicitly setting the k-dim chunksize.

diag_manager/fms_diag_field_object.F90
```diff
@@ -1218,25 +1219,27 @@ end subroutine get_dimnames
 
 !> @brief Wrapper for the register_field call. The select types are needed so that the code can go
 !! in the correct interface
-subroutine register_field_wrap(fms2io_fileobj, varname, vartype, dimensions)
+subroutine register_field_wrap(fms2io_fileobj, varname, vartype, dimensions, chunksizes)
   class(FmsNetcdfFile_t),            INTENT(INOUT) :: fms2io_fileobj!< Fms2_io fileobj to write to
   character(len=*),                  INTENT(IN)    :: varname       !< Name of the variable
   character(len=*),                  INTENT(IN)    :: vartype       !< The type of the variable
   character(len=*), optional,        INTENT(IN)    :: dimensions(:) !< The dimension names of the field
+  integer,   optional,               INTENT(IN)    :: chunksizes(:) !< Chunksize to use, only relevant is using
```
bensonr: 'is using' or 'when using'?

diag_manager/fms_diag_field_object.F90
```diff
@@ -1342,6 +1357,60 @@ subroutine write_field_metadata(this, fms2io_fileobj, file_id, yaml_id, diag_axi
   endif
 end subroutine write_field_metadata
 
+!> @brief Determine the appropriate chunksizes for a diagnostic field based on its axes.
+!! For "X" and "Y" axes, the function returns a chunksize equal to the axis size divided by the layout.
+!! If the dimension is not evenly divisible by the layout, the function raises an error.
+!! For the other axis it return a chunksize equal to the axis length
+!! For sub-z axes (e.g., layer bounds), a chunksize of 1 is returned for now, as this case is not yet implemented.
```
bensonr, line 1363: Are you explicitly setting the k dimension for the chunksize and to what value are you setting it?  If you are using 1, will that result in really small chunksizes? Don't leave any ambiguity specify it explicitly to be `nz` or whatever is appropriate.

diag_manager/fms_diag_field_object.F90
```diff
@@ -1257,12 +1262,22 @@ subroutine write_field_metadata(this, fms2io_fileobj, file_id, yaml_id, diag_axi
...
-    call register_field_wrap(fms2io_fileobj, var_name, this%get_var_skind(field_yaml), dimnames)
+
+    !! Collective writes are only used for 2D+ variables
+    if (use_collective_writes .and. size(this%axis_ids) >= 2) then
+      chunksizes = this%get_chunksizes(diag_axis, field_yaml)
+      call register_field_wrap(fms2io_fileobj, var_name, this%get_var_skind(field_yaml), dimnames, &
+        chunksizes = chunksizes)
+    else
+      call register_field_wrap(fms2io_fileobj, var_name, this%get_var_skind(field_yaml), dimnames)
+    endif
```
J-Lentz, line 1279: Why is `chunksizes` only passed to `register_field_wrap` when collective writes are enabled? Unless I'm misunderstanding, I thought custom chunk sizes would be supported regardless of whether collective writes are enabled.

diag_manager/diag_yaml_format.md
uramirez8707: The z dimension is chunked. That's what the original line was trying to explain with "For non-domain-decomposed dimensions, the chunk size is set to the dimension length." I’ll reword this part to make it clearer.

diag_manager/fms_diag_field_object.F90
uramirez8707: The chunksize is set to nz by default, but if they are outputting a section of the Z axis, like 
```yaml
   - var_name: var3
    output_name: var3_Z
    zbounds: 2. 3.
```

The default chunksize will be set to 1. Ideally, we will want to set to the size of the sub z axes. I have that as a todo to fix later as this is not very common.

diag_manager/fms_diag_field_object.F90
uramirez8707: Ah yes you are correct. The logic/documentation was updated.

diag_manager/diag_yaml_format.md
```diff
@@ -524,4 +525,64 @@ diag_files:
...
+  use_collective_writes: True
+  chunksizes: 96, 8, 1, 1, 1
+  varlist:
+  - var_name: var3
+    chunksizes: 96, 4, 1, 1, 1
+  - var_name: var4
+```
 
+**Things to consider:**
+1.  This option currently only works for files with variables that are domain-decomposed (i.e., each processor has its own section of the data). Support for land's unstructured grid and icebergs will be added in the future; at present, using this option with those grids will result in a FATAL error.
```
bensonr: suggest languate of "multiple mpi-tasks each with their own section of data"

diag_manager/diag_yaml_format.md
```diff
@@ -524,4 +525,64 @@ diag_files:
...
+2.  The domain must be evenly decomposed, meaning each processor has the same amount of data. For example, a C3072L65 grid with a layout of (30, 32) will not work and will result in a FATAL error, since 3072 is not evenly divisible by 30.
```
bensonr: Let's explain the limitation and state this will never work.

PR #1782, opened 2025-10-10
Implement OpenMP offload for do group update

### LINKED ISSUE #1771: Enable OpenMP GPU-to-GPU MPI blocking transfers

We've started porting MOM6 to GPUs using OpenMP target offload. Currently, when performing halo updates, we ship data back to CPU to communicate data via MPI. We want to update FMS (mainly the mpp parts) to be able to perform the transfers directly between GPUs so we can avoid shipping data back and forth between the CPU and GPU, which is currently significantly negatively impacting our performance.

For now, porting blocking communications should be sufficient. This requires a few changes:
* an optional flag in `mpp_do_group_update` to indicate whether CPU or GPU transfer should happen
  * this flag would need to be passed to called subroutines like `mpp_send/recv` and consequently `mpp_transmit`. 
  * The flag can be optional with a default so existing code that uses mpp_do_group_update doesn't need to be updated.
* In the packing/unpacking step, OpenMP target directives need to be added to allocate `buffer` on GPU and to pack that variable with GPU data, which would optionally execute based on value of the above flag.
  * The packing/unpacking loops probably need to be modified to enable GPU parallelization.
* The MPI calls would need to be wrapped in openmp directives to ensure the GPU version of the variable is being used (conditionally based on above flag).

So far, we've been shipping data from the GPU back to the CPU, which is terrible for performance.

**Additional context**

I've made a first attempt at these changes in [my fork of FMS](https://github.com/edoyango/FMS/compare/bd32c3b...2025.03-ompoffload) which hopefully illustrates the changes needed. In my changes, `mpp_do_group_update` has an `omp_offload` flag which, if set to `.true.` does the packing and communication on the GPU (assuming the MPI is GPU-aware). These changes have been tested with openmpi 5.0.8 built with cuda 12.9 and using nvfortran 24.9 using flags `-mp=gpu -gpu=mem:separate -O1 -Mnofma`.

The set of changes in my fork is just enough to get our test case working which only uses the real versions of the calls. More changes are needed to cover more cases - especially in the packing/unpacking step.

I'll be opening a pull request once I've ported the remaining packing loops.

edoyango:
**Description**
This adds the necessary code to make mpp_do_group_update work with arrays that are managed by NVIDIA's OpenMP offload runtime. This attempts to be minimally disruptive in that non-nvidia compilers will see the same behaviour as previously by adding macros around the relevant openmp directives.

Fixes #1771

**How Has This Been Tested?**
The OpenMP offload capability is currently tested on the "double gyre" case in MOM6-examples using the `nvfortran` compiler and a cuda-aware openmpi. We have [some notes](https://github.com/marshallward/mom6-gpu-report/blob/main/gpu_report.rst) on how to run the gpu-enabled MOM6, but is outdated.

**Checklist:**
- [ ] `make distcheck` passes

### DIFF
--- a/mpp/include/group_update_pack.inc
+++ b/mpp/include/group_update_pack.inc
@@ -35,7 +35,9 @@ if( group%k_loop_inside ) then
               ptr_field = group%addrs_s(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: field(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: field(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do j = js, je
@@ -52,7 +54,9 @@ if( group%k_loop_inside ) then
               ptr_field = group%addrs_s(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: field(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: field(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do i = is, ie
@@ -69,7 +73,9 @@ if( group%k_loop_inside ) then
               ptr_field = group%addrs_s(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: field(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: field(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do i = ie, is, -1
@@ -86,7 +92,9 @@ if( group%k_loop_inside ) then
               ptr_field = group%addrs_s(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: field(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: field(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do j = je, js, -1
@@ -106,7 +114,9 @@ if( group%k_loop_inside ) then
               ptr_fieldx = group%addrs_x(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do j = js, je
@@ -124,7 +134,9 @@ if( group%k_loop_inside ) then
                  ptr_fieldy = group%addrs_y(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
                  !$omp target teams distribute parallel do collapse(3) private(idx) &
-                 !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+                 !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) &
+                 !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+                 !$omp   if (use_device_ptr)
 #endif
                  do k = 1, ksize
                     do i = is, ie
@@ -141,7 +153,9 @@ if( group%k_loop_inside ) then
                  ptr_fieldy = group%addrs_y(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
                  !$omp target teams distribute parallel do collapse(3) private(idx) &
-                 !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+                 !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) &
+                 !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+                 !$omp   if (use_device_ptr)
 #endif
                  do k = 1, ksize
                     do i = is, ie
@@ -159,7 +173,9 @@ if( group%k_loop_inside ) then
               ptr_fieldy = group%addrs_y(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
               !$omp target teams distribute parallel do collapse(3) private(idx) &
-              !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+              !$omp   map(to: fieldy(is:ie,js:je,1:ksize)) &
+              !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+              !$omp   if (use_device_ptr)
 #endif
               do k = 1, ksize
                  do i = ie, is, -1
@@ -179,7 +195,9 @@ if( group%k_loop_inside ) then
                  ptr_fieldx = group%addrs_x(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
                  !$omp target teams distribute parallel do collapse(3) private(idx) &
-                 !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+                 !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) &
+                 !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+                 !$omp   if(use_device_ptr)
 #endif
                  do k = 1, ksize
                     do j = je, js, -1
@@ -196,7 +214,9 @@ if( group%k_loop_inside ) then
                  ptr_fieldx = group%addrs_x(l)
 #ifdef __NVCOMPILER_OPENMP_GPU
                  !$omp target teams distribute parallel do collapse(3) private(idx) &
-                 !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) map(from: buffer(pos+1:pos+ksize*nj*ni)) if(use_device_ptr)
+                 !$omp   map(to: fieldx(is:ie,js:je,1:ksize)) &
+                 !$omp   map(from: buffer(pos+1:pos+ksize*nj*ni)) &
+                 !$omp   if (use_device_ptr)
 #endif
                  do k = 1, ksize
                     do j = je, js, -1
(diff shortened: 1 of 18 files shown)


mpp/include/group_update_unpack.inc
```diff
@@ -18,92 +18,129 @@
...
+           !$omp target teams distribute parallel do collapse(3) if(use_device_ptr) private(idx) &
+           !$omp   map(to: buffer(pos+1:pos+ksize*nj*ni)) &
+           !$omp   map(from: field(is:ie,js:je,1:ksize))
+#endif
            do k = 1, ksize
               do j = js, je
                  do i = is, ie
-                    pos = pos + 1
-                    field(i,j,k) = buffer(pos)
+                    idx = pos + (k-1)*nj*ni + (j-js)*ni + (i-is) + 1
```
nikizadehgfdl, line 48: How are these two implementations equivalent? Is new idx = old pos always?

edoyango: Yes, they're equivalent. For any iteration, `idx = pos + (k-1)*nj*ni + (j-js)*ni + (i-is) + 1` produces the same value that `pos` would have had at that point. The formula accounts for all the iterations that would have occurred in the nested loops up to that (i,j,k) position.

The reason for the change is that each nested iteration is now independent and can be performed in parallel.

marshallward:
I've finally managed to put together a timing test for MOM6, and the timings for domain decomposition are promising.  There is strong scaling for the larger domains as we saturate the GPU.

gpu_mpi_comm

Scaling is shown for up to 8 GPUs over 4 nodes on Ursa, so this method should work on our NOAA resources.

Ursa does not have Nvlink, so its comm time shouldn't be taken as reflective of production runs.  But I was surprised that it was not too large, even after going through the PCIe and Infiniband layers.

I will look into including `mpp` tests which exercise this new code.

mpp/include/group_update_pack.inc
```diff
@@ -1,158 +1,233 @@
...
+!* License along with FMS.  If not, see <http://www.gnu.org/licenses/>.
 !***********************************************************************
 
 if( group%k_loop_inside ) then
+#ifndef __NVCOMPILER_OPENMP_GPU
 !$OMP parallel do default(none) shared(npack,group,ptr,nvector,ksize,buffer_start_pos) &
 !$OMP                          private(buffer_pos,pos,m,is,ie,js,je,rotation, &
-!$OMP                                  ptr_field, ptr_fieldx, ptr_fieldy,n,k)
+!$OMP                                  ptr_field, ptr_fieldx, ptr_fieldy,n,k,ni,nj,idx)
+#endif
```
bensonr, line 29: In the case where __NVCOMPILER_OPENMP_GPU is defined, the compiled library may offload the packing to the device (assuming `use_device_ptr`).  But if `use_device_ptr` is false, then we get zero parallelism.  It would be good to have both options available at the same time for maximum flexibility.

bensonr (commented):
For some of the target omp statements, the `if (use_device_ptr)` is prior to the target data mapping and other times after.  Does it matter where in the !$omp specification the if-test occurs?

The preference for !$omp is `default(none)` with full specification of `shared` and `private`.

mpp/include/group_update_pack.inc
marshallward: I think the plan is to rename this to something that is set at compile-time (i.e. cmake, autoconf).

bensonr: @marshallward - Not sure I understand your comment and I'm sure mine wasn't entirely clear.  I think we need both a CPU and GPU version of routines like this to allow offload and traditional execution within the same programmatic unit.  It isn't entirely clear we get this with such fine-grained GPU offloading and a mix of macro if-tests.

marshallward: Sorry for the brief reply above.  As you say, the current form would only allow either CPU parallelism or GPU parallelism.  I don't think we imagined using both at this stage of development, and assumed one would make this decision at compile-time.

Replacing `__NVCOMPILER_OPENMP_GPU` with a generic macro would decouple that decision from the choice of compiler, but I agree it doesn't permit the use of both in the same model.

We can create separate versions of `group_update_[un]pack.inc` rather than rely on preprocessing, if you are OK with the code duplication.

There are also bigger picture ideas here that could consolidate the two versions (e.g. team-parallelism of groups with thread-parallelism of buffer transfer), but AFAIK there are technical issues preventing this, such as the presence of Cray pointers.

edoyango: > @marshallward - Not sure I understand your comment and I'm sure mine wasn't entirely clear. I think we need both a CPU and GPU version of routines like this to allow offload and traditional execution within the same programmatic unit. It isn't entirely clear we get this with such fine-grained GPU offloading and a mix of macro if-tests.

Is there a preference for how this is implemented? would something like below work? To avoid code duplication

```fortran
! below include contains openmp offloaded directives but not cpu openmp directives
#include <group_update_pack.inc>
! undefine macro to enable compilation of cpu openmp parallel loop
#ifdef GPU_MACRO
#undef GPU_MACRO
if (.not.omp_offload) then
! contains cpu openmp directives without openmp offload directives
#include <group_update_pack.inc>
endif
#define GPU_MACRO
#endif
```

bensonr: @edoyango - macro-driven, templated Fortran is preferable to maintaining different, yet identical versions that differ by compile-time comment lines only.  Please implement this - if you haven't already.

edoyango: hi @bensonr I've attempted to implement the suggested macro solution, but nvfortran fails to run. 

the fail happens when there's an openmp offload loop and a openmp cpu loop - either with `default(none)` - referencing the same cray pointer in the same procedure. For the latest compilers, whichever loop has `default(none)` either hangs or segfaults at runtime. 

the workaround is to use `default(shared)` in both directives, or to put them into isolated subroutines.

reproducer:
[test_cray_ptr_omp.txt](https://github.com/user-attachments/files/26764707/test_cray_ptr_omp.txt)
compile with `nvfortran -mp=gpu` and run with `./a.out 0/1/2/3`. 0, 1 cases are gpu and cpu (respectively) with `default(none)`, and 2, 3 are with `default(shared)`.

Could a concession be made to use `default(shared)`? otherwise we can't make progress until cray pointers are removed.

mpp/include/mpp_transmit.inc
```diff
@@ -170,7 +171,7 @@
       call mpp_transmit( put_data, put_len, to_pe, dummy, 1, NULL_PE, tag=tag, send_request=request )
     end subroutine MPP_SEND_
 
-    subroutine MPP_RECV_SCALAR_( get_data, from_pe, glen, block, tag, request )
+    subroutine MPP_RECV_SCALAR_( get_data, from_pe, glen, block, tag, request, omp_offload )
```
bensonr, line 174: @edoyango - I see you only added the offload option for the scalar version of the send/recv pairs.  Is there a reason this is not being done for all of the send/recv routines?

mpp/include/group_update_pack.inc
bensonr: @edoyango - Thanks for explaining the issue with default(none) and why you'd need default(shared) for now.  It would be helpful to leave the default(none) logic in place along with extra comments for the people in the process of removing cray pointers.

edoyango: @bensonr I've:

* added required shared/private to the openmp offloaded loops
* swapped from `default(none)` to `default(shared)` in the packing/unpacking loops
* added comments to the top of `group_update_un/pack.inc` files explaining the restrictions imposed by nvfortran with cray pointers
* through macro manipulation, ensured openmp cpu parallelism is activated when `ompoffload==.false.`

I hope that's sufficient.

mpp/include/mpp_transmit.inc
edoyango: I've tried to keep the scope of this PR to only cover what's required to get `do_group_update` working and `do_group_update` only uses `MPP_SEND/RECV_SCALAR_`. I can add the `omp_offload` flag to the other implementations, but it would be untested.

bensonr: Thanks for the update.  I'll discuss with the rest of the team how we want to handle the other routines.

Now that you've mentioned testing, will you be adding or adapting any of the existing unit tests to add GPU-specific tests?

edoyango: |> Now that you've mentioned testing, will you be adding or adapting any of the existing unit tests to add GPU-specific tests?

Sorry for the late reply! I just pushed some tests to this branch. It co-opts the existing mpp domain tests by adding a test_group_offload switch (thanks @rem1776 for mocking one together). The test passes for both nvhpc 26.3. gcc passes as well, although it doesn't actually offload to GPU. 

Note that the domain tests include non-blocking tests, but the openmp offload version of those haven't been implemented, so those tests just run on CPU.

rem1776:
@edoyango Could you update this with main when you get a chance? Looks like its showing some conflicts that we should probably clear up before we get this reviewed.

edoyango:
> @edoyango Could you update this with main when you get a chance? Looks like its showing some conflicts that we should probably clear up before we get this reviewed.

Hi @rem1776 i've tried rebasing on top of main locally, but nvfortran still can't build main because of the use of `select rank` in `mpp/include/mpp_pack.fh`

rem1776:
> > @edoyango Could you update this with main when you get a chance? Looks like its showing some conflicts that we should probably clear up before we get this reviewed.
> 
> Hi @rem1776 i've tried rebasing on top of main locally, but nvfortran still can't build main because of the use of `select rank` in `mpp/include/mpp_pack.fh`

Sorry about that! Thought the fix (#1873) was merged but looks like its still waiting for a review. It should get merged sometime this week, I'll let you know.

edoyango:
@rem1776 ok i've rebased on top of main as well as your select rank workaround and kept my tests.

I did notice that main + your workaround produces garbled terminal output with nvhpc though...

e.g. this is MOM6 standalone output:

```
P-�p@Xpr��������-�������@����+��:��;�pkhprL|�'�mgpr�:gp�gprP�\�Ɉhpr�gprpkhpr�\��\�gpr��\�Ɉhpr�
%���\���\��\�
             �ݞ@�gprL|�'�gpr(�\T�\�ohpr�gprL|�'p�\�P�\���hprpkhpr��\���hprohpr��\�pr
pkhpr�\��\�gpr��\�Ɉhpr�
%���\���\��\�
             �ݞ@�gprL|�'�gpr(�\T�\�ohpr�gprL|�'p�\�P�\���hprpkhpr��\���hprohpr��\�prNOTE: MPP_DOMAINS_SET_STACK_SIZE: stack size set to   955296.
S_SET_STACK_SIZE: stack size set to   955296.
```

edoyango:
> @rem1776 ok i've rebased on top of main as well as your select rank workaround and kept my tests.
> 
> I did notice that main + your workaround produces garbled terminal output with nvhpc though...
> 
> e.g. this is MOM6 standalone output:
> 
> ```
> P-�p@Xpr��������-�������@����+��:��;�pkhprL|�'�mgpr�:gp�gprP�\�Ɉhpr�gprpkhpr�\��\�gpr��\�Ɉhpr�
> %���\���\��\�
>              �ݞ@�gprL|�'�gpr(�\T�\�ohpr�gprL|�'p�\�P�\���hprpkhpr��\���hprohpr��\�pr
> pkhpr�\��\�gpr��\�Ɉhpr�
> %���\���\��\�
>              �ݞ@�gprL|�'�gpr(�\T�\�ohpr�gprL|�'p�\�P�\���hprpkhpr��\���hprohpr��\�prNOTE: MPP_DOMAINS_SET_STACK_SIZE: stack size set to   955296.
> S_SET_STACK_SIZE: stack size set to   955296.
> ```

this turned out to be an issue with an uninitialized msg variable being printed when `pe == 1`. Creating an `else` branch to handle `pe == 1` case solves it. It's small, so I've just added it to this PR.

rem1776 (dismissed):

rem1776:
@edoyango Thanks for putting that mpp_error fix in! i think i was seeing something similar with gcc but i like your fix better. I'm going it mark this PR as ready for review so we can try to include it in our next tag.

vithikashah001 (dismissed):

uramirez8707 (dismissed):
Thank you @edoyango 

Looks good to me.

edoyango:
Apologies, I was missing the latest commit from https://github.com/NOAA-GFDL/FMS/pull/1873 so I rebased which has unfortunately dismissed the reviews for this PR.

@rem1776 @vithikashah001 @bensonr @uramirez8707

uramirez8707 (approved):

rem1776 (approved):

bensonr (commented):
@edoyango - As I look through, the work seems soild.  What I could not easily discern is if there is a capability to have both offload capable and cpu-versions of the same routine when compiling with NVHPC.  Can you  confirm that is supported within this PR?  Once you do, I will finalize my approval.

edoyango:
> @edoyango - As I look through, the work seems soild. What I could not easily discern is if there is a capability to have both offload capable and cpu-versions of the same routine when compiling with NVHPC. Can you confirm that is supported within this PR? Once you do, I will finalize my approval.

yes gpu offload and cpu parallel execution are supported in the same routines. Passing `ompoffload = .true.` triggers the offloaded branch, and `ompoffload = .false.` triggers the omp CPU parallel branch for the relevant packing/unpacking code. This is implemented with macros ([example](https://github.com/NOAA-GFDL/FMS/pull/1782/changes#diff-8ffb4a20e513f20a23401548dd071b4858f29fa101ed50c5fa3ec6dca89e17e2R511))

bensonr:
>yes gpu offload and cpu parallel execution are supported in the same routines. Passing ompoffload = .true. triggers the offloaded branch, and ompoffload = .false. triggers the omp CPU parallel branch for the relevant packing/unpacking code. This is implemented with macros ([example](https://github.com/NOAA-GFDL/FMS/pull/1782/changes#diff-8ffb4a20e513f20a23401548dd071b4858f29fa101ed50c5fa3ec6dca89e17e2R511))

Thanks for reminding me of the means you used to implement the capabilities.  I was focusing on the __NV_COMPILER ifdefs and not the offload flag.

bensonr (approved):

Issue #1786, opened 2025-10-22
unit test failures with cmake


rem1776:
**Describe the bug**
These two tests run into issues when run via ctest:

- [ ] test_mpp_nesting; fails with gcc 13 and later, any flags. Not sure why this one is failing, passes with autotools so likely some type of flag/precision issue. Errors are from slightly off checksums.
- [ ] test_sat_vapor_pres; only fails on ARM with release build type flags.

**To Reproduce**
mpp nesting test - build and run `make test` anywhere with cmake using gcc 13 or later
sat_vapor_pres - build and run on a arm system with release flags

**Expected behavior**
These tests should pass, especially the nested domain test.
 
**System Environment**
amd box, gcc 14

**Additional context**
these tests will be skipped in the CI for now

PR #1787, opened 2025-10-22
fixes for cmake 32/64bit builds and enable CI testing via ctest

### LINKED ISSUE #1783: Are 32BIT and 64BIT CMake options still supported?
Per the docs and CMake, I see there are still:
```
-D32BIT       "Build 32-bit (r4) FMS library"        DEFAULT: OFF
-D64BIT       "Build 64-bit (r8) FMS library"        DEFAULT: OFF
```
as options. However, I'm not sure you can build FMS (2025.04) any more with both on. I tried recently building FMS with those both on (since that is the way we've built it with GEOS since time immemorial) but when you try, you get a hojillion CMake errors like:
```
CMake Error at CMakeLists.txt:784 (target_link_libraries):
  Target "check_time_sum" links to:

    FMS::fms

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

I think that's because when you build with `-D32BIT=ON -D64BIT=ON` the targets created are `FMS::FMS_r4` and `FMS::fms_r8` but the CMake is assuming the FMS target will be `FMS::fms`.

For now, I'm going to try and see if I can get the multi-precision variant (aka `FMS::fms`) working with GEOS because, well, that's what we actually want to use, but I wanted to ask if this was expected.

Perhaps the CMake being triggered should be avoided if not building as multi-precision?

rem1776:
**Description**
Fixes errors that come up when building using the 32 or 64 bit build options. Now it will link in whatever's built so it won't fail when it gets to the unit test executables.

Also includes a few small changes to get the unit tests from cmake running in the CI. Skips two tests (#1786) until we can get them working since they have cmake/arm specific bugs.

Fixes #1783 

**How Has This Been Tested?**
tested on amd box with gcc 13/14 + mpich

**Checklist:**
- [x] `make distcheck` passes


CMakeLists.txt
```diff
@@ -343,7 +343,7 @@ foreach(kind ${kinds})
                                                  amip_interp/include)
   target_compile_definitions(${libTgt}_f PRIVATE "${fms_defs}")
   target_compile_definitions(${libTgt}_f PRIVATE "${${kind}_defs}")
-  set_target_properties(${libTgt}_f PROPERTIES COMPILE_FLAGS ${${kind}_flags})
+  set_target_properties(${libTgt}_f PROPERTIES COMPILE_FLAGS "${${kind}_flags}")
```
mlee03, line 346: comments not for this line, but just looking at line 76 - 79, the messages seem misleading.  Could they be changed to something like "building select modules and procedures in 32bit precision"?

mlee03: Also, line 278 - 278, I don't think the OVERLOAD macros  are there anymore

CMakeLists.txt
```diff
@@ -588,6 +588,14 @@ include(CTest)
 set(MPI_LAUNCHER "mpirun")
 # used in the test-lib.sh.in to make it behave differently when parsed by cmake
 set(USING_CMAKE "true")
+# set the fms library to link tests with based on whats built
+if(NOT kinds)
+  set(fmsLibraryName FMS::fms)
```
mlee03, line 595: line 439, isn't kind always defined?

CMakeLists.txt
```diff
@@ -588,6 +588,14 @@ include(CTest)
...
+elseif(64BIT)
+  set(fmsLibraryName FMS::fms_r8)
+else()
```
mlee03: maybe this should  explicitly be elseif(32BIT)

test_fms/time_interp/test_time_interp_conservative_hi.F90
```diff
@@ -33,6 +33,7 @@ program test_time_interp_with_interp
   use time_interp_external2_mod, only: init_external_field, time_interp_external_init, time_interp_external
 
   use time_manager_mod, only: JULIAN, time_type, set_date, set_calendar_type, time_manager_init
+  use platform_mod, only: r4_kind, r8_kind
```
mlee03, line 36: ah this probably worked because DTEST_FMS_KIND_ is set to 4 or 8 in Makefile.am
Is this use statement needed?

.github/workflows/github_cmake_gnu.yml
```diff
@@ -12,42 +12,46 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        omp-flags: [ -DOPENMP=on, -DOPENMP=off ]
```
mlee03, line 15: just curious, why was this removed?

CMakeLists.txt
```diff
@@ -879,8 +890,16 @@ foreach (testScript ${TEST_SCRIPTS})
         ENVIRONMENT "parser_skip=skip"
     )
   endif()
+
 endforeach()
 
+# skip parallel netcdf tests if its not enabled in given install
+IF(NOT NetCDF_PARALLEL)
```
mlee03, line 1000: Leaving what @J-Lentz pointed out, line 290 should say "parallel netcdf io" instead of "collective netcdf io"

mlee03 (changes requested):

test_fms/time_interp/test_time_interp_conservative_hi.F90
mlee03: unless cmake substitutes them as r4_kind and r8_kind somewhere?

CMakeLists.txt
rem1776: line 439 uses `kind` as opposed to `kinds` which is only set if one of the 32/64bit options is enabled.

test_fms/time_interp/test_time_interp_conservative_hi.F90
rem1776: > unless cmake substitutes them as r4_kind and r8_kind somewhere?

Yeah cmake uses r4/r8_kind so that its uniform across all the different unit tests.

.github/workflows/github_cmake_gnu.yml
rem1776: That was mainly to make the amount of jobs smaller. It'll just compile with openmp everytime.

I was thinking as we move away from autotools we can add more options to this one, just didn't want to make the CI too bloated.

CMakeLists.txt
rem1776: > Also, line 278 - 278, I don't think the OVERLOAD macros are there anymore

nice catch! should be removed now

rem1776:
@mlee03 any comments i didn't respond to should be fixed now with https://github.com/NOAA-GFDL/FMS/pull/1787/commits/985d6e76728f59041eadaa406bc94fcf71496bbf

mlee03 (approved):
All comments addressed :)

CMakeLists.txt
```diff
@@ -69,14 +69,14 @@ option(USE_DEPRECATED_IO     "THIS OPTION HAS NO EFFECT AND WILL BE REMOVED IN A
 
 if(32BIT)
   list(APPEND kinds "r4")
+  message(STATUS "Building library with 4-byte real defaults (with mixed precision real support for most modules).")
```
mlee03, line 71: one more request, line 68 needs to be removed

uramirez8707 (dismissed):

uramirez8707 (approved):

Issue #1753, opened 2025-08-19
Refactor `open_collective_netcdf_file` and `netcdf_file_open`


J-Lentz:
`open_collective_netcdf_file` and `netcdf_file_open` should be refactored at some point to reduce the amount of duplicated code.

PR #1789, opened 2025-11-03
fms2_io: Refactor MPI-IO support

### LINKED ISSUE #1767: Consolidate fms2io collective writes/reads implementation
The collective netcdf reads/writes were implemented in similar but different ways.
For example, to turn on collective reads, you will need to do
```F90
  fileobj%use_collective = .true.
  fileobj%tile_comm = mpp_get_domain_tile_commid(Domain_read)
  if (open_file(fileobj, "test_collective_io.nc", "read", Domain_read, nc_format=nc_format)) then
      !! ...
  endif

```
but for collective writes, you will need to do
```F90
if (open_file(fileobj, filename, "overwrite", domain, use_netcdf_mpi=.true.)) then
endif
```

(` fileobj%tile_comm = mpp_get_domain_tile_commid(Domain_read)` is handled inside the open_file)

Collective reads should be implemented in the same way as collective writes.
Warning should be added if using the old format.

N/A

**Additional context**
N/A

J-Lentz:
**Description**
MPI-IO support in fms2_io has been refactored. `open_collective_netcdf_file` has been removed and merged into `netcdf_file_open`, to reduce the amount of duplicated code. The API for using collective reads has been changed to be consistent with parallel writes. Collective read unit tests have been updated to use the new API.

Fixes #1767 and #1753

**How Has This Been Tested?**
Builds with ifx on the AMD box and unit tests pass.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/fms2_io/readme.md
+++ b/fms2_io/readme.md
@@ -14,6 +14,7 @@ Before introducing the FMS2_io module, subroutines and functions in fms_io and m
 - [G. Ascii_io](readme.md#g-ascii_io)
 - [H. FMS2_io namelist](readme.md#h-fms2_io-namelist)
 - [I. Chunking](readme.md#i-chunking)
+- [J. MPI-IO](readme.md#j-mpi-io)
 
 ### A. FMS2_io Fileobjs
 FMS2_io provides three new derived types, which target the different I/O paradigms used in GFDL models.
@@ -630,3 +631,15 @@ dim_names, chunsizes=chunksizes)
 - **NOTE: This argument is only valid with "NETCDF4" file formats, otherwise it will be ignored. You can set the netcdf file format for all files using the 'netcdf_default_format' namelist or in a per file basis by using the 'nc_format' argument in the 'open_file' call**
 
 - See the NETCDF user guide for more information: (https://cluster.earlham.edu/bccd-ng/testing/mobeen/GALAXSEEHPC/netcdf-4.1.3/man4/netcdf.html#Chunking)
+
+### J. MPI-IO
+When opening a NetCDF-4 file, the HDF5 library's MPI-IO driver can be used by calling `open_file` with the optional
+`use_netcdf_mpi=.true.` argument. When a file is opened in this manner, the I/O domain is ignored and every PE directly
+performs reads or writes in parallel. In the case of parallel reads, the MPI-IO driver handles coordination and metadata
+sharing among the PEs, which can improve performance compared to the standard POSIX I/O driver. Additionally, the
+`use_collective=.true.` argument can be passed to `open_file` to enable collective reads/writes: this can improve I/O
+performance significantly, but it requires each read/write call to run synchronously across all participating PEs (i.e.,
+it introduces an MPI barrier at the start of each read/write call).
+
+MPI-IO support is only available for NetCDF-4 files. `use_netcdf_mpi` and `use_collective` are both disabled by default.
+`use_collective` has no effect unless `use_netcdf_mpi` is enabled.


fms2_io/netcdf_io.F90
```diff
@@ -567,59 +567,83 @@ function netcdf_file_open(fileobj, path, mode, nc_format, pelist, is_restart, do
                                               !! is a restart file.  Defaults
                                               !! to false.
   logical, intent(in), optional :: dont_add_res_to_filename !< Flag indicating not to add
-                                              !! ".res" to the filename
+                                                            !! ".res" to the filename
+  integer, intent(in), optional :: tile_comm      !< MPI communicator used for parallel I/O. Passing this
+                                                  !! argument enables parallel I/O.
```
uramirez8707, line 572: I don't think we need this option.

Fms2io can get the tile communication from the domain.

fms2_io/netcdf_io.F90
```diff
@@ -567,59 +567,83 @@ function netcdf_file_open(fileobj, path, mode, nc_format, pelist, is_restart, do
...
+
+  if (fileobj%use_collective) then
+    fileobj%use_netcdf_mpi = .true.
+    call mpp_error(NOTE, "Setting fileobj%use_collective to enable collective reads is deprecated. &
+                          Please use open_file(..., use_collective=.true.) instead.")
+  else
+    fileobj%use_collective = .false.
+    if (present(use_collective)) fileobj%use_collective = use_collective
   endif
+
```
uramirez8707, line 614: I think we don't need to the else statement here.

Before this update to use collective io, they will do:
```
fileobj%use_collective = .True.
fileobj%tile_comm = ...
open_file(fileobj, ...)
```

Now they will just need to do
```
open_file(fileobj, ..., use_netcdf_mpi=.True.
```
use_netcdf_mpi and use_collective are equivalent.

So to not break, their current code we should just do:
```
  if (fileobj%use_collective) then
    fileobj%use_netcdf_mpi = .true.
    call mpp_error(NOTE, "Setting fileobj%use_collective to enable collective reads is deprecated. &
                          Please use open_file(..., use_collective=.true.) instead.")
endif
```

and remove any use of `use_collective` in the code.

fms2_io/include/netcdf_read_data.inc
```diff
@@ -150,6 +150,8 @@ subroutine netcdf_read_data_0d(fileobj, variable_name, buf, unlim_dim_level, &
   else
     bcast = .true.
   endif
+  if (fileobj%use_netcdf_mpi) bcast = .false.
```
uramirez8707, line 153: :) smart! I see why you did this!

test_fms/fms2_io/test_collective_io.sh
```diff
@@ -43,16 +43,4 @@ test_expect_success "Test NetCDF-4 collective reads" '
...
 
-rm -rf *.nc*
-# The code should still run if not using netcdf4 files, it just won't use collective io
-cat <<_EOF > input.nml
-&test_collective_io_nml
-  nc_format = "64bit"
-/
-_EOF
-
-test_expect_success "Test fallback to non-collective reads" '
```
uramirez8707, line 54: Let's add this test back and expect a failure.

uramirez8707 (changes requested):
Some minor changes but very nice job. 

Can you also add a section for this in the "documentation"?
https://github.com/NOAA-GFDL/FMS/blob/main/fms2_io/readme.md

fms2_io/netcdf_io.F90
J-Lentz: The problem is that `mpp_get_domain_tile_commid()` requires a `domain` argument. In `netcdf_file_open`, `fileobj` is `class(FmsNetcdfFile_t)`, which doesn't provide access to the `domain` attribute. Therefore, it makes more sense to call `mpp_get_domain_tile_commid()` within `open_domain_file`, where we do have access to the domain.

fms2_io/netcdf_io.F90
J-Lentz: This refactor allows the MPI-IO driver (enabled with the `use_netcdf_mpi` argument to `open_file`) and collective read/write operations (enabled with `use_collective`) to be controlled independently. `use_netcdf_mpi` is required for the `use_collective` option to have any effect, but I'm not sure if there's a compelling use case for non-collective MPI-IO reads/writes (which would correspond to `use_netcdf_mpi=.true., use_collective=.false.`). It might be worth doing some performance testing for this case if we haven't already.

If there's no compelling reason to support MPI-IO in non-collective mode, then I agree with your suggestions.

PR #1863, opened 2026-05-04
Allocate dummy arrays to pass to `mpp_gather` calls

### LINKED ISSUE #1860: MPI errors during domain writes when compiled with GCC
**Describe the bug**
Most of the modern diag_manager unit tests and some of the fms2_io tests fail when compiled with GNU 15/14 and mpich. The usual error is usually something like this during a write_data call:

```shell
CFI_is_contiguous: Base address of C descriptor is already NULL
Abort(205168898) on node 1: Fatal error in internal_Type_create_hvector: Invalid count, error stack:
```

Since this issue goes away when the mpi_f08 module changes are reverted, this might be a issue with MPICH's handling of the MPI new types.

**To Reproduce**
Build with gcc 15 and mpich, along with --with-yaml/-DWITH_YAML.

Most the of the test_time_<reduction>.sh tests will fail, along with some tests in fms2_io.

**System Environment**
amd box, gaea

**Additional context**
Tests seem to pass with openmpi on the GNU 15 github CI image.

J-Lentz:
**Description**
Many unit tests currently fail when built with GCC, due to un-allocated arrays which are passed to `mpi_gather`. While the un-allocated arrays are on non-root PEs where they shouldn't really matter, they're being checked for contiguity in the GCC builds, which causes a crash with the following message:
```
CFI_is_contiguous: Base address of C descriptor is already NULL
Abort(205168898) on node 1: Fatal error in internal_Type_create_hvector: Invalid count, error stack:
```
This PR allocates dummy arrays of size 1 wherever the respective arrays would otherwise not be allocated.

Fixes #1860.

**How Has This Been Tested?**
GCC 14.2.0 on C5. Fixes all the `fms2_io` tests and most of the `diag_manager` tests.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/diag_manager/diag_output.F90
+++ b/diag_manager/diag_output.F90
@@ -48,7 +48,6 @@ use,intrinsic :: iso_c_binding, only: c_double,c_float,c_int64_t, &
   use mpp_domains_mod, only: mpp_get_UG_io_domain
   use mpp_domains_mod, only: mpp_get_UG_domain_npes
   use mpp_domains_mod, only: mpp_get_UG_domain_pelist
-  use mpp_mod,         only: mpp_gather
   use mpp_mod,         only: uppercase,lowercase
   use fms2_io_mod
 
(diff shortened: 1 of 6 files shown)


uramirez8707 (approved):
Thanks for fixing this @J-Lentz 
Looks good to me.

rem1776 (approved):

PR #1873, opened 2026-05-14
add deferred-rank-less versions of mpp_pack and mpp_global_field

rem1776:
**Description**
adds a workaround for nvhpc's lack of support for deferred-rank arrays. Copies over the nice new routines that use deferred ranks and makes them worse by using explicit rank sizes, then adds them to the corresponding interfaces to support all required dimensions the classic fortran way.

Open to any advice for how to name/manage the macro. not sure if i should just leave it as __NVCOMPILER, but it would be more convenient.

Fixes #1864

**How Has This Been Tested?**
nvhpc 26.1ng (beta release) + mpich was able to compile and pass most of the mpp tests. I got 3 failures, but I think they are all unrelated to these updates.

25.1 was able to compile too, but many of the mpp tests fail.

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included
- [x] `make distcheck` passes

PR #1875, opened 2026-05-21
add workflow for nvhpc testing

rem1776:
**Description**
adds a workflow for testing with nvhpc compilers. Instead of using a container image like the other workflows, this will use spack to build the compiler, then save it to a spack OCI build cache for subsequent runs.

The package versions will be determined by the latest software available in the `packages_ref` option, which corresponds to the latest tag in the spack-packages repo. This prevents frequent rebuilds due to changing hashes by using a stable package list, but will need to updated.

needs #1873 merged first since that fixes the compilation error.

**How Has This Been Tested?**
tested on my fork with my package registry:
https://github.com/rem1776/FMS/actions/runs/26187563194/job/77047118521

**Checklist:**
- [x] `make distcheck` passes


bensonr (commented):
Please take a look at the loop ordering and access to the *`buf`* and *`field_d`* arrays to make sure we are using a unit stride where possible.

PR #1876, opened 2026-05-27
Fix Intel CI workflow and migrate off Intel container images 

### LINKED ISSUE #1872: fix broken intel classic workflow and migrate from intel images
**Describe the bug**
The current workflow for running intel classic tests weekly is failing from apt commands, which looks like an issue with the package list intel is using. I think we should move away from using their images, since they also tend to be large and hit storage quotas on the runners.

**To Reproduce**
n/a, CI issue

**Expected behavior**
workflow should pass, or at least run the tests

**System Environment**
CI

**Additional context**
I think this is a good opportunity to update both our intel workflows to use this action @AlexanderRichert-NOAA kindly made available:

https://github.com/NOAA-EMC/ci-install-intel-toolkit

vithikashah001:
**Description**
Fix Intel CI workflow by replacing broken Intel container image usage and updating environment variables for compatibility.

Fixes #1872 

**How Has This Been Tested?**

**Checklist:**
- [ ] `make distcheck` passes


.github/workflows/github_autotools_intel_classic.yml
```diff
@@ -12,51 +12,56 @@ concurrency:
...
       id: cache
       uses: actions/cache@v4.2.0
       with:
         path: /libs
         key: intel-libs
+    - name: Install Intel compilers & MPI
+      uses: NOAA-EMC/ci-install-intel-toolkit@develop
+      with:
+        install-classic: true
+        install-oneapi: true
```
rem1776: I would set this one to false instead, we'll only need the classic compiler for this workflow.

rem1776 (commented):
It looks like this is failing from not having the intel compilers in the PATH during the hdf5 build, even though its able to run configure sucessfully. The install action you added should be making those changes for us automatically, so not sure why its not there.

I would try adding a couple of debugging commands like `echo $PATH` and `which mpiicx` as a step after the compilers are installed so we can see where things are on the filesystem and get a better idea of whats going on.

.github/workflows/github_autotools_intel_classic.yml
```diff
@@ -12,53 +12,72 @@ concurrency:
...
+        ./configure --enable-remote-fortran-bootstrap --prefix=/libs
+        sudo make -j install
+        sudo make -j -k build-netcdf-fortran
+        sudo make -j install-netcdf-fortran
         wget https://github.com/yaml/libyaml/releases/download/0.2.5/yaml-0.2.5.tar.gz
         tar xf yaml-0.2.5.tar.gz && cd yaml-0.2.5
         ./configure --prefix=/libs
-        make -j install && cd
+        sudo make -j install
+        cd ../ && sudo rm -rf *
```
rem1776:
```suggestion
        cd /home/runner/work/FMS/FMS && sudo rm -rf *
```
lets try cd-ing into the exact directory its going to clone to

PR #1878, opened 2026-05-28
Generalized indices: interpolator_mod

J-Lentz:
**Description**
Add generalized indices support to `interpolator_mod` via a new `dim_order` argument, which has been added to the 2D, 3D, and 4D variants of the `interpolator` subroutine.

**How Has This Been Tested?**
Builds on the AMD box and tests pass using ifx 2025.2.0 and GCC 15.2.0.

**Checklist:**
- [x] `make distcheck` passes

### DIFF
--- a/interpolator/include/interpolator.inc
+++ b/interpolator/include/interpolator.inc
@@ -1379,6 +1379,7 @@ integer, parameter :: lkind=FMS_INTP_KIND_
 !! @param [in] <js> OPTIONAL: Index for the physics window
 !! @param [out] <interp_data> The model fields with the interpolated climatology data
 !! @param [out] <clim_units> OPTIONAL: The units of field_name
+!! @param [in] <dim_order> OPTIONAL: Order of the x (1), y (2), z (3), and field (4) dimensions
 !!
 !! @throw FATAL "interpolator_4D : You must call interpolator_init
 !!                    before calling interpolator"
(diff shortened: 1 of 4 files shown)


interpolator/include/interpolator.inc
```diff
@@ -1813,28 +1831,28 @@ do j = 1, size(phalf,2)
...
 
+!--lwh
+     do i= 1, nfields
+       select case(clim_type%mr(i))
+         case(KG_M2)
+           interp_buf(:,:,:,i) = interp_buf(:,:,:,i)*phalf_diff
+       end select
      end do
 
+     interp_data = reshape(interp_buf, shape(interp_data), order=dim_map)
```
uramirez8707: Can we change this so that it only happens if the dim_map is not the default [1, 2, 3, 4]? Not sure if there is a performance penalty for doing this.

uramirez8707 (commented):
Just a minor comment, but it looks good otherwise.

interpolator/include/interpolator.inc
J-Lentz: 2b309ec9e838a1e47ed799e16859d384dd16900e avoids the temporary buffer unless it's actually needed, and prints a `NOTE` when it's used. Only applicable to the 3D and 4D cases, since the `reshape` operation just replaces a copy operation in the 2D case, and therefore doesn't add an additional unnecessary copy.

uramirez8707 (approved):

PR #1885, opened 2026-06-05
Generalized indices: Refactor group_update for improved performance

J-Lentz:
**Description**
Refactor the packing and unpacking routines of mpp_group_update for improved performance with generalized indices. The older versions of the packing and unpacking loops have been kept, and the new generalized versions are only called when generalized indices are required.

**How Has This Been Tested?**
Builds on C5 with ifx. `mpp` unit tests all pass.

**Checklist:**
- [x] `make distcheck` passes


uramirez8707:
I tested this branch with AM5 to compare performance, but I am getting a seg fault error

forrtl: severe (174): SIGSEGV, segmentation fault occurred
/gpfs/f5/gfdl_f/scratch/Uriel.Ramirez/volatile/am5/am5f12e3r0_2026.02-alpha3-candidate/c96L65_am5f12e3r0_amip/ncrc5.intel25-prod-openmp/stdout/run/c96L65_am5f12e3r0_amip_1x0m8d_576x2a.o135778381

The error appears to be consistent. 
1 thread jobs fail as well:
/gpfs/f5/gfdl_f/scratch/Uriel.Ramirez/volatile/am5/am5f12e3r0_2026.02-alpha3-candidate/c96L65_am5f12e3r0_amip/ncrc5.intel25-prod-openmp/stdout/run/c96L65_am5f12e3r0_amip_1x0m8d_576x1a.o135778386

c384 failed as well:
/gpfs/f5/gfdl_f/scratch/Uriel.Ramirez/volatile/am5/am5f12e3r0_2026.02-alpha3-candidate/c384L65_am5f12e3r0_amip/ncrc5.intel25-prod-openmp/stdout/run/c384L65_am5f12e3r0_amip_1x0m8d_1728x4a.o135778376

uramirez8707:
@J-Lentz Can you merge main and fix the merge conflicts?

J-Lentz:
> @J-Lentz Can you merge main and fix the merge conflicts?

Main has been merged. I suspect that the two remaining CI failures are unrelated to this PR.

@rem1776 - What are your thoughts on the CI failures?

rem1776:
@J-Lentz Looks like it just needed to be retriggered, its passing now. I think that one test tends to hit a race condition with writing it's input files.

J-Lentz:
> @J-Lentz Looks like it just needed to be retriggered, its passing now. I think that one test tends to hit a race condition with writing it's input files.

Thanks Ryan!

mpp/include/group_update_pack.inc
```diff
@@ -16,619 +16,940 @@
...
+                   pos = pos + nj*ni
+                end do
+             end if
+          end select ! select case( rotation(n) )
+       endif
+    enddo
+  endif
+end subroutine GROUP_UPDATE_PACK_
+
+subroutine GROUP_UPDATE_PACK_GENERAL_ (group, buffer, buffer_start_pos)
```
abrooks1085, line 665: I might be missing something here, but how is omp_offload=.true. expected to behave for generalized indices?

GROUP_UPDATE_PACK_ receives use_device_ptr, but GROUP_UPDATE_PACK_GENERAL_ is called without it. The generalized loops also use !$omp parallel do rather than !$omp target ..., while mpp_send/mpp_recv still receive omp_offload.

Would this mean the generalized path packs into the host copy of buffer while the communication uses the device copy? The unpack path may have the same host/device mismatch in the opposite direction.

J-Lentz: Do we need to support generalized indices in conjunction with GPU offloading? If not, the easiest solution for now might be to issue a fatal error if the user tries to use a non-default index order while `use_device_ptr` is true.

J-Lentz: 225ee81de38b5fa68f0b9dc23c3da5daae3d4292 adds a fatal error if a non-default index order is used when `use_device_ptr` is true. If this use case does indeed need to be supported, we can incorporate new OpenMP pragmas into `GROUP_UPDATE_PACK_` and `GROUP_UPDATE_UNPACK_` based on the pragmas from Edward Yang's PR.

uramirez8707: > https://github.com/NOAA-GFDL/FMS/commit/225ee81de38b5fa68f0b9dc23c3da5daae3d4292 adds a fatal error if a non-default index order is used when use_device_ptr is true.

There isn't a use case for using generalized indices + gpu offloading, so I think this is the proper way to go.

uramirez8707 (approved):

abrooks1085 (approved):

PR #1886, opened 2026-06-05
Generalized Indices: MPP_REDISTRIBUTE

abrooks1085:
**Description**
This PR adds generalized storage-layout support to MPP_REDISTRIBUTE for 2D–5D fields. Callers may pass an optional axis_to_storage mapping, defined as:

axis_to_storage(logical_axis) = storage_dimension

For example, axis_to_storage = (/3,1,2/) means logical x, y, and z are stored in dimensions 3, 1, and 2, respectively.

The redistribute wrappers now pass this mapping through to the low-level pack/unpack logic, which maps logical domain bounds onto the correct storage dimensions before accessing field data. For 4D and 5D fields, extra dimensions are handled as repeated contiguous 3D chunks instead of being collapsed onto the third dimension, avoiding the previous assumption that storage dimension 3 is always logical z.

**How Has This Been Tested?**
amdbox + intel

**Checklist:**
- [x] `make distcheck` passes


mpp/include/mpp_do_redistribute.fh
```diff
@@ -17,97 +17,126 @@
...
+       shape3(map(2)) = domain_in%y(1)%domain_data%end - domain_in%y(1)%domain_data%begin + 1
+       shape3(map(3)) = ke
+       addr = int(f_in(l), kind=c_intptr_t)
+       p_in = transfer(addr, c_null_ptr)
+       call c_f_pointer(p_in, field_in, shape3)
+       do k = 1,ke
+          do j = js,je
+             jo = j - js + 1
+             do i = is,ie
+                io = i - is + 1
```
uramirez8707: These `io` and `jo` variables appear never be used?

mpp/include/mpp_update_domains2D.fh
```diff
@@ -217,39 +217,46 @@
...
     subroutine MPP_REDISTRIBUTE_3D_( domain_in, field_in, domain_out, field_out, complete, free, list_size, &
-                                   &  dc_handle, position )
+                                   &  dc_handle, position, axis_to_dim_in )
       type(domain2D), intent(in) :: domain_in, domain_out
       MPP_TYPE_, intent(in)  :: field_in (:,:,:)
       MPP_TYPE_, intent(out) :: field_out(:,:,:)
       logical, intent(in), optional :: complete, free
       integer, intent(in), optional :: list_size
       integer, intent(in), optional :: position
+      integer, intent(in), optional :: axis_to_dim_in(3)
```
uramirez8707: Please document this new arguments. 

Can we make it consistent with the other subroutines? I think we used `dim_order`.

mpp/include/mpp_update_domains2D.fh
```diff
@@ -217,39 +217,46 @@
...
     subroutine MPP_REDISTRIBUTE_2D_( domain_in, field_in, domain_out, field_out, complete, free, list_size, &
-                                   &  dc_handle, position )
+                                   &  dc_handle, position, axis_to_dim_in )
       type(domain2D), intent(in) :: domain_in, domain_out
       MPP_TYPE_, intent(in)  :: field_in (:,:)
       MPP_TYPE_, intent(out) :: field_out(:,:)
       logical, intent(in), optional :: complete, free
       integer, intent(in), optional :: list_size
       integer, intent(in), optional :: position
+      integer, intent(in), optional :: axis_to_dim_in(3)
```
uramirez8707: should this be 2? `axis_to_dim_in(2)`

uramirez8707 (changes requested):
Just minor comments regarding the new argument name. Looks good to me otherwise.

mpp/include/mpp_update_domains2D.fh
```diff
@@ -339,38 +365,44 @@
...
+                                   &  dc_handle, position, axis_to_dim_in )
       type(domain2D), intent(in) :: domain_in, domain_out
       MPP_TYPE_, intent(in)  :: field_in (:,:,:,:)
       MPP_TYPE_, intent(out) :: field_out(:,:,:,:)
       logical, intent(in), optional :: complete, free
       integer, intent(in), optional :: list_size
       integer, intent(in), optional :: position
+      integer, intent(in), optional :: axis_to_dim_in(3)
       MPP_TYPE_ :: field3D_in (size(field_in, 1),size(field_in, 2),size(field_in ,3)*size(field_in ,4))
       MPP_TYPE_ :: field3D_out(size(field_out,1),size(field_out,2),size(field_out,3)*size(field_out,4))
```
J-Lentz: Unless I'm misunderstanding something, it seems to me that this approach of fusing the third and fourth dimensions should only work with (i,j,\*,\*) or (j,i,\*,\*) dimension orders. Are these the only cases that we need to support?

mpp/include/mpp_update_domains2D.fh
J-Lentz: Same as above. Unless I'm misunderstanding, it seems to me that this should only work with (i,j,\*,\*,\*) or (j,i,\*,\*,\*) dimension orders.

mpp/include/mpp_update_domains2D.fh
abrooks1085: Thanks, I think you are right to question the logical equivalence here.

I was initially focused on preserving the storage layout: fusing storage dimensions 3 and 4 into a rank-3 view preserves Fortran column-major ordering and avoids a repack or transpose. However, looking more carefully at MPP_DO_REDISTRIBUTE_3D_, the routine redistributes logical x/y rectangles and carries the full logical z extent along with each communicated region. The part I missed is the fourth dimension may be fused with an intended decomposed horizontal axis.

I'm going to instead treat the fourth dimension as a list of independent 3D views, one for each field(:,:,:,l) slice. That is, I'll pass a 1D list of base addresses/pointers, where each entry points to the start of one contiguous 3D slice of the original 4D field. Each 3D slice can then be passed through the existing 3D redistribution path using axis_to_dim without copying.

This will preserves the intended logical behavior: x/y determine the communicated domain region, while the full logical z extent and each fourth-dimension slice are carried along.

I can also limit the performance impact, by making packing/unpacking loops oriented towards storage layout instead of logical layout like I have here. Using axis_to_dim, compute the logical bounds for x/y/z, map them onto storage dimensions, and traverse each 3D view in column-major storage order. That keeps the innermost loop unit-stride where possible while still preserving the logical decomposition.

Thanks again for pointing this out.

mpp/include/mpp_do_redistribute.fh
abrooks1085: removed unused variables

mpp/include/mpp_update_domains2D.fh
abrooks1085: Good point. I’ll add documentation for the new argument.

I also agree that this should be consistent across the MPP interfaces. While reviewing this, I found that dim_order is ambiguous because it has been used in different senses. In mpp_redistribute, the mapping is from logical axis to storage dimension:

axis_to_storage(logical_axis) = storage_dimension

For example, axis_to_storage = (/3,1,2/) means logical x, y, and z are stored in dimensions 3, 1, and 2.

The current pelist gather/scatter implementation uses dim_order in the opposite direction internally, mapping storage dimension to logical axis. To avoid exposing two conventions, I’m renaming the public argument to axis_to_storage for redistribute and will update pelist gather/scatter to use the same public convention, converting internally to the reverse map where needed. This should also make the interface consistent with @J-Lentz’s group update PR.

rem1776:
@abrooks1085 I don't think the current oneapi CI failure is from this PR, seems to be happening with every branch.

rem1776:
> @abrooks1085 I don't think the current oneapi CI failure is from this PR, seems to be happening with every branch.

Might've spoken too soon on this actually, I would try pulling main i think the last merge might've fixed it.

abrooks1085:
> > I don't think the current oneapi CI failure is from this PR, seems to be happening with every branch.
> 
> Might've spoken too soon on this actually, I would try pulling main i think the last merge might've fixed it.

@rem1776  Should already but up to date with main

uramirez8707 (approved):

rem1776 (approved):
LGTM. I'll try bumping up our intel CI version after this is merged and see if that fixes the error. Doesn't seem to happen anywhere else but the CI environment.

PR #1897, opened 2026-07-07
Implement workarounds for modern diag manager with NVHPC

### LINKED ISSUE #1895: `fms_array_to_pointer` returns garbage pointers with NVHPC
**Describe the bug**

Discovered when trying the modern diag manager with MOM6 + NVHPC (but no GPU flags).

Segfault occurs in the file/variable name sorting in `diag_yaml_object_init` during the sorting step. The segfault is due to getting the C pointer addresses from the list if file and variable names.

```fortran
file_list%file_pointer = fms_array_to_pointer(file_list%file_name)
variable_list%var_pointer = fms_array_to_pointer(variable_list%var_name)
```
The pointers returned do not match the actual addresses of the elements in `variable_list%var_name`. Inlining `fms_array_to_pointer` manually fixes the error. 

**To Reproduce**
Compile FMS with NVHPC and `-Duse_yaml` and compile/link below program against libFMS

```fortran
program repro1_dangling
  use fms_string_utils_mod, only: fms_array_to_pointer, fms_sort_this
  use iso_c_binding,        only: c_ptr, c_loc, c_null_char, c_intptr_t
  implicit none
  character(len=255), allocatable, target :: names(:)
  type(c_ptr),        allocatable         :: ptrs(:)
  integer,            allocatable         :: ids(:)
  integer(c_intptr_t) :: got, expect
  integer :: i
  logical :: dangling

  allocate(names(3), ids(3)); ids = [1, 2, 3]
  names(1) = "alpha"//c_null_char
  names(2) = "beta"//c_null_char
  names(3) = "gamma"//c_null_char

  ptrs = fms_array_to_pointer(names)

  dangling = .false.
  do i = 1, size(names)
    got    = transfer(ptrs(i),        0_c_intptr_t)
    expect = transfer(c_loc(names(i)), 0_c_intptr_t)
    print *, "i =", i, "  returned ptr =", got, "  address of names(i) =", expect
    if (got /= expect) dangling = .true.
  end do
  if (dangling) print *, "BUG: fms_array_to_pointer pointers do not alias names(:) -> dangling"
  flush(6)

  ! Deref of the dangling pointers -> SIGSEGV, independent of string contents.
  call fms_sort_this(ptrs, 3, ids)
  print *, "ids =", ids, " (did not crash)"
end program repro1_dangling
```
with `mpifort repro1_dangling.f90 {-I,-L}${FMS_HOME} -lFMS && ./a.out` I get
```
 i =            1   returned ptr =                311039184
   address of names(i) =                311039184
 i =            2   returned ptr =                311071952
   address of names(i) =                311039439
 i =            3   returned ptr =                311104720
   address of names(i) =                311039694
 BUG: fms_array_to_pointer pointers do not alias names(:) -> dangling
Segmentation fault (core dumped)
```
Note only the first address matches, but the other 2 don't.

In a real run with MOM6, this manifests as a segfault when using the modern diag manager.

**Expected behavior**
The address of all 3 `names(i)` elements match the addresses that are returned from `fms_array_to_pointer`.

**System Environment**
- OS: Linux Mint 22.3 and ubuntu 22.04
- Compiler(s): NVHPC 26.3 and 26.5
- MPI type, and version: OpenMPI 5.0.10
- netCDF Version: C: 4.9.3, Fortran: 4.6.2
- Configure options: `FCFLAGS=-O1 -Mnofma -Mnovect -g`

**Additional context**

This is likely a bug on the nvhpc side, but probably requires a workaround in the meantime.

Started a topic on the nvidia forums: https://forums.developer.nvidia.com/t/problem-with-getting-c-loc-pointers-from-character-arrays/375393

After hotfixing this, using the new diag manager in MOM6 still leads to a segfault, but a separate issue will be added for that.

### LINKED ISSUE #1896: `fms_diag_object%fms_register_static_field` crashes with NVHPC
**Describe the bug**
Segfault happens when compiling FMS with NVHPC and using `fmsDiagObject_type%fms_register_static_field` and not passing in `range`. This is apparently an issue with the underlying `this%register` where NVHPC seems to force a copy into the `register` `varrange` argument - even if `range` isn't passed into `fms_register_static_field`. But for some reason making `varrange` assumed shape instead of `(2)` fixes it?

**To Reproduce**
compile and link the below program against NVHPC-compiled libFMS.
```fortran
program repro2_fms
  use fms_diag_object_mod, only: fms_diag_object
  implicit none
  integer :: id

  print *, "repro2_fms: calling fms_register_static_field with NO range (absent optional)"; flush(6)
  id = fms_diag_object%fms_register_static_field("ocean_model", "some_field", [1, 2])
  print *, "repro2_fms: returned id=", id, " -- reached the forward without crashing"; flush(6)
end program repro2_fms
```
This also manifests as a segfault when using the modern diag manager in MOM6 with NVHPC.

**Expected behavior**
The above program should complete.

**System Environment**
* OS: Linux Mint 22.3 and ubuntu 22.04
* Compiler(s): NVHPC 26.3 and 26.5
* MPI type, and version: OpenMPI 5.0.10
* netCDF Version: C: 4.9.3, Fortran: 4.6.2
* Configure options: FCFLAGS=-O1 -Mnofma -Mnovect -g

**Additional context**
Is definitely a compiler a bug.
started a topic on the nvidia developer forums: https://forums.developer.nvidia.com/t/absent-optional-explicit-shape-class-bug-with-nvfortran/375427

edoyango:
**Description**

~This PR is built on top of https://github.com/NOAA-GFDL/FMS/pull/1873~

There are a couple compiler bugs in nvfortran that cause segfaults when using the modern diag manager. One is when sorting filenames (see https://github.com/NOAA-GFDL/FMS/issues/1895), and the other is when registering diagnostics (see https://github.com/NOAA-GFDL/FMS/issues/1896). This PR addresses them both with workarounds.

The filename sorting issue uses the workaround suggested by nvidia, which was to specify the input character length explicitly in `fms_array_to_pointer`, rather than using deferred shape arrays. This worked because the call sites passes arrays of character length 255. However, the test needed to be updated as the test passed in characters of length 10. This does reduce the flexibility of `fms_array_to_pointer`. An alternative workaround is to inline the routines manually (I think this would only be needed in two places).

The registration issue is addressed by changing explicit shape to deferred shape (same workaround as suggested in the issue).  

Fixes https://github.com/NOAA-GFDL/FMS/issues/1895
Fixes https://github.com/NOAA-GFDL/FMS/issues/1896

**How Has This Been Tested?**
* Running MOM6 double gyre with modern diag manager
* Running diag manager tests with NVHPC 26.3 and GCC 15.2 and confirming no changes in pass/fails.

**Checklist:**
- [x] `make distcheck` passes
  * pass with GCC
  * fail with NVHPC
    * previously `diag_yaml` checks would hang, but now they pass (but fail). Next hang at reduction checks.

### DIFF
--- a/diag_manager/diag_data.F90
+++ b/diag_manager/diag_data.F90
@@ -335,9 +335,7 @@ use platform_mod
     class(*), allocatable         :: att_value(:) !< Value of the attribute
     character(len=:), allocatable :: att_name     !< Name of the attribute
     contains
-#ifndef __NVCOMPILER
       procedure :: add => fms_add_attribute
-#endif
       procedure :: write_metadata
   end type fmsDiagAttribute_type
 ! Include variable "version" to be written to log file.
@@ -562,7 +560,6 @@ CONTAINS
     res = base_second
   end function get_base_second
 
-#ifndef __NVCOMPILER
   !> @brief Adds an attribute to the attribute type
   subroutine fms_add_attribute(this, att_name, att_value)
     class(fmsDiagAttribute_type), intent(inout) :: this         !< Diag attribute type
@@ -571,6 +568,7 @@ CONTAINS
 
     integer :: natt !< the size of att_value
 
+#ifndef __NVCOMPILER
     natt = size(att_value)
     this%att_name = att_name
     select type (att_value)
@@ -593,8 +591,8 @@ CONTAINS
           aval = att_value
       end select
     end select
-  end subroutine fms_add_attribute
 #endif
+  end subroutine fms_add_attribute
 
   !> @brief gets the type of a variable
   !> @return the type of the variable (r4,r8,i4,i8,string)
--- a/diag_manager/diag_data.F90
+++ b/diag_manager/diag_data.F90
@@ -335,9 +335,9 @@ use platform_mod
     class(*), allocatable         :: att_value(:) !< Value of the attribute
     character(len=:), allocatable :: att_name     !< Name of the attribute
     contains
-  #ifndef __NVCOMPILER
+#ifndef __NVCOMPILER
       procedure :: add => fms_add_attribute
-  #endif
+#endif
       procedure :: write_metadata
   end type fmsDiagAttribute_type
 ! Include variable "version" to be written to log file.
--- a/diag_manager/diag_data.F90
+++ b/diag_manager/diag_data.F90
@@ -335,7 +335,9 @@ use platform_mod
     class(*), allocatable         :: att_value(:) !< Value of the attribute
     character(len=:), allocatable :: att_name     !< Name of the attribute
     contains
+  #ifndef __NVCOMPILER
       procedure :: add => fms_add_attribute
+  #endif
       procedure :: write_metadata
   end type fmsDiagAttribute_type
 ! Include variable "version" to be written to log file.
@@ -560,6 +562,7 @@ CONTAINS
     res = base_second
   end function get_base_second
 
+#ifndef __NVCOMPILER
   !> @brief Adds an attribute to the attribute type
   subroutine fms_add_attribute(this, att_name, att_value)
     class(fmsDiagAttribute_type), intent(inout) :: this         !< Diag attribute type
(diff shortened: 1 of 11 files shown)


edoyango:
rebased on top of main since 1873 has been merged in already

uramirez8707 (approved):

